PR #9, opened 2013-09-01
Lang 4 x

lingeng1986:
add a ToStringStyle named ShortPrefixWithoutNullFieldToStringStyle , which has no difference with ShortPrefixToStringStyle except don't print null fields.


PascalSchumacher:
I tried but failed to extract the changes.

britter:
@PascalSchumacher I think @lingeng1986 should take care of clearing the PR up so that it can be easily reviewed and applied.

PR #29, opened 2014-07-21
Annotation @ToStringExclude

FelipeAdorno:
Use this annotation to builds a String excluding the annotated field.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/annotation/ToEqualsExclude.java
+++ b/src/main/java/org/apache/commons/lang3/annotation/ToEqualsExclude.java
@@ -1,0 +1,15 @@
+package org.apache.commons.lang3.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Use this annotation to builds a equals excluding the annotated field.
+ */
+@Retention(RUNTIME)
+@Target(ElementType.FIELD)
+public @interface ToEqualsExclude {
+
+}
--- a/src/test/java/org/apache/commons/lang3/builder/EqualsBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/EqualsBuilderTest.java
@@ -23,6 +23,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.Arrays;
 
+import org.apache.commons.lang3.annotation.ToEqualsExclude;
 import org.junit.Test;
 
 /**
@@ -1148,6 +1149,38 @@ public class EqualsBuilderTest {
         assertTrue(!EqualsBuilder.reflectionEquals(d1, d2));
         assertTrue(EqualsBuilder.reflectionEquals(d1, d1));
         assertTrue(EqualsBuilder.reflectionEquals(d1, d3));
+    }
+    
+	 static class TestObjectEqualsExclude {
+		 @ToEqualsExclude
+		 private int a;
+		 private int b;
+		 
+		 public TestObjectEqualsExclude(int a, int b) {
+			 this.a = a;
+			 this.b = b;
+		 }
+		 
+		 public int getA() {
+			 return a;
+		 }
+		 
+		 public int getB() {
+			 return b;
+		 }
+	 }
+    
+	@Test 
+    public void testToEqualsExclude() {
+    	TestObjectEqualsExclude one = new TestObjectEqualsExclude(1, 2);
+    	TestObjectEqualsExclude two = new TestObjectEqualsExclude(1, 3);
+    	
+    	assertTrue(!EqualsBuilder.reflectionEquals(one, two));
+    	
+    	one = new TestObjectEqualsExclude(1, 2);
+    	two = new TestObjectEqualsExclude(2, 2);
+    	
+    	assertTrue(EqualsBuilder.reflectionEquals(one, two));
     }
 
 }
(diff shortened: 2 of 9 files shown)


FelipeAdorno:
Hi @britter do you think this pull request is able to up in a new version?

src/test/java/org/apache/commons/lang3/builder/EqualsBuilderTest.java
```diff
@@ -1149,6 +1150,38 @@ public void testReflectionArrays() throws Exception {
...
+			 return b;
+		 }
+	 }
+    
+	@Test 
+    public void testToEqualsExclude() {
+    	TestObjectEqualsExclude one = new TestObjectEqualsExclude(1, 2);
+    	TestObjectEqualsExclude two = new TestObjectEqualsExclude(1, 3);
+    	
+    	assertTrue(!EqualsBuilder.reflectionEquals(one, two));
```
britter: Should be ```assertFalse(EqualsBuilder.reflectionEquals(one, two))```

src/test/java/org/apache/commons/lang3/builder/HashCodeBuilderTest.java
```diff
@@ -564,5 +566,52 @@ public void testToHashCodeEqualsHashCode() {
...
+		 
+		 public int getB() {
+			 return b;
+		 }
+	 }
+    
+	@Test 
+    public void testToEqualsExclude() {
+		TestObjectHashCodeExclude one = new TestObjectHashCodeExclude(1, 2);
+		TestObjectHashCodeExclude2 two = new TestObjectHashCodeExclude2(1, 2);
```
britter: No tabs please.

src/main/java/org/apache/commons/lang3/annotation/ToEqualsExclude.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Use this annotation to builds a equals excluding the annotated field.
+ */
+@Retention(RUNTIME)
+@Target(ElementType.FIELD)
+public @interface ToEqualsExclude {
```
britter: Should this really be a public class? It doesn't really make sense outside EqualsBuilder. What do you think about implementing it as inner type of EqualsBuilder?

FelipeAdorno: @britter I will go change the code for adapt to your comments.
About the ToEqualsExclude I think make sense stay outside EqualsBuilder because when you use the apache commons you need note the classes with  @ToEqualsExclude if this is inside EqualsBuilder you need note with @EqualsBuild.ToEqualsExclude. What do you think?

britter: @FelipeAdorno I don't understand our comment. Do you see a use case for using any of the new annotations without using the corresponding builder?

FelipeAdorno: @britter sorry but I can't understand I'll try explaining for you the use of this annotation.
Suppose you have a class and you need use EqualsBuilder or ToStringBuilder or HashCodeBuilder, but you have a field and this field not is necessary in equals and/or tostring and/or hashcode you can note this with @ToStringExclude  and/or @ToEqualsExlucde and/or @ToHashCodeExclude. If one of this annotations is a inner class of EqualsBuilder, ToStringBuilder or HashCodeBuilder when use the apache commons need note using @EqualsBuilder.ToEqualsExclude, @ToStringBuilder.ToStringExclude and  @HashCodeBuilder.ToHashCodeExclude and not only note @ToEqualsExclude, @ToStringExclude and  @ToHashCodeExclude.
Can you understand?

britter: @FelipeAdorno okay, I get it now. Makes sense to me.

src/main/java/org/apache/commons/lang3/annotation/ToEqualsExclude.java
```diff
@@ -0,0 +1,15 @@
+package org.apache.commons.lang3.annotation;
```
britter: Just saw this, when I was about to apply the patch. Please put the annotations in the builder package. I know it is a common pattern to create a specific package for annotation, but it simply makes no sense. It's better to package stuff by their domains and bot by their nature. For example you would never create a package called "interface" and put all your interfaces their. Why should we do this for annotations? They are no special in any way :-)

britter:
@FelipeAdorno we're almost there. There is just al little clean up to do, before I can apply the patch:
- remove tab characters. Only use spaces for indentation
- Some assertions can be simplyfied: use ```assertFalse(bool)``` instead of ```assertTrue(!bool)```
- Move annotations to the builder package
- Rename ```ToEqualsExclude``` to ```EqualsExclude``` and ```ToHashCodeExclude``` to ```HashCodeExclude```

let me elaborate on the last point some more: I think the names of the annotations should match the methods from ```Object``` they are supposed to be used for. We use the ```ToStringExclude``` annotations when building the result for ```toString``` method. When talking about the ```equals``` method, the corresponding annotations should be called ```EqualsExclude```. The same is true for ```hashcode```.

Makes sense for you?

britter:
@FelipeAdorno Commons Lang has been migrated from svn to git. Because of this the main branch has been renamed to 'master'. Can you please rebase your work against the master branch? Thank you!

FelipeAdorno:
Ok, I'll rebase today and I commit the modifications.
- remove tab characters. Only use spaces for indentation
- Some assertions can be simplyfied: use assertFalse(bool) instead of assertTrue(!bool)
- Move annotations to the builder package
- Rename ToEqualsExclude to EqualsExclude and ToHashCodeExclude to HashCodeExclude

britter:
@FelipeAdorno very nice, thank you!

britter:
Any news here? I'd like to merge this :)

FelipeAdorno:
sorry I made a mistake and I commit in a new pull request.
https://github.com/apache/commons-lang/pull/77

PR #34, opened 2014-10-17
Multiline recursive to string style

janmaterne:
Works with ToStringBuilder to create a "deep" toString(). 
But instead a single line like the RecursiveToStringStyle this creates a multiline String similar to the ToStringStyle.MultiLineToStringStyle.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/builder/MultilineRecursiveToStringStyleTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/MultilineRecursiveToStringStyleTest.java
@@ -27,233 +27,235 @@ import org.junit.Test;
 
 public class MultilineRecursiveToStringStyleTest {
 
-	private final String BR = SystemUtils.LINE_SEPARATOR;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/builder/MultilineRecursiveToStringStyle.java
```diff
@@ -0,0 +1,218 @@
...
+ * &nbsp; job=Job@43cd2[ <br>
+ * &nbsp; &nbsp; title=Manager <br>
+ * &nbsp;  ] <br>
+ * ]
+ * </code>
+ * </p>
+ * 
+ * @since 3.4
+ */
+class MultilineRecursiveToStringStyle extends RecursiveToStringStyle {
```
mcint: Should this have been public? I've been trying to use this class, but can neither import it nor find a factory method that might give me an instance of the class.

britter:
It has already been made public. The change will be released in 3.5, probably before the end of the year.

janmaterne:
Very good to hear!

PR #35, opened 2014-10-18
Avoid memory allocation when using date formating to StringBuffer.

mbracher:


britter:
Placeholder is https://issues.apache.org/jira/browse/LANG-1100

britter:
Hello @mbracher,

can you please explain this some more?

mbracher:
Hi Benedikt,

I'm calling the date formatter more then a million times a second. To avoid unnecesarry garbage collection i'm using a shared string buffer, but the formatter is still allocating intermediate strings.The patch avoids the creation of these intermediate strings and no garbage needs to be collected.

PR #36, opened 2014-10-24
Replace string

otaviojava:
During the Java compilation performance the JDK does improvements with String, this way, these may be replaced with java.lang.String expressions which results in simpler and possibly more efficient code.


src/test/java/org/apache/commons/lang3/BooleanUtilsTest.java
```diff
@@ -321,7 +321,7 @@ public void test_toBoolean_String() {
         assertFalse(BooleanUtils.toBoolean("false"));
         assertFalse(BooleanUtils.toBoolean("a"));
         assertTrue(BooleanUtils.toBoolean("true")); // interned handled differently
-        assertTrue(BooleanUtils.toBoolean(new StringBuffer("tr").append("ue").toString()));
+        assertTrue(BooleanUtils.toBoolean("tr" + "ue"));
```
britter: Doesn't make sense at all

otaviojava: It is the same test, but without the ´new `StringBuilder`

britter: Yes I understand. But the old one didn't make sense as well :-) I think we can drop this assertion completely.

britter:
@otaviojava Commons Lang has been migrated from svn to git. Because of this the main branch has been renamed to 'master'. Can you please rebase your work against the master branch? Thank you!

chonton:
Most changes were committed in https://github.com/apache/commons-lang/commit/cc1aed9bdf196403c673a886bbf723101171a9bf

otaviojava:
Thank you guys.
So this PR is not more necessary.

PR #37, opened 2014-11-11
add RandomEnumUtils to operate with random instances of enum types

gabrielruiu:
Added a utility class that returns a random instance of an Enum of a certain provided type. This, I consider, is useful when creating test data instances with random values for its fields, among which are most likely some Enum types as well.


britter:
@gabrielruiu as it stands now, there are two vetos in jira against including this. You can try convince the others if you like. But right know it looks like this is out of scope for lang.

gabrielruiu:
Well if it is not considered a useful feature then I see no reason to push for its integration.

PR #41, opened 2015-01-18
Lang 1081 - DiffBuilder left.equals(right)

j--baker:
Please see: https://issues.apache.org/jira/browse/LANG-1081

Two parts:
1) Added two (T) casts to get my old JDK to work during the Maven tests.
2) Added fix to DiffBuilder and updated the JUnit test.  A left.equals(right) (or vice versa) test is now done.

Thank you for your time.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/SerializationUtils.java
+++ b/src/main/java/org/apache/commons/lang3/SerializationUtils.java
@@ -122,7 +122,7 @@ public class SerializationUtils {
      * @since 3.3
      */
     public static <T extends Serializable> T roundtrip(final T msg) {
-        return (T) SerializationUtils.deserialize(SerializationUtils.serialize(msg));
+        return SerializationUtils.deserialize(SerializationUtils.serialize(msg));
     }
 
     // Serialize
--- a/src/main/java/org/apache/commons/lang3/SerializationUtils.java
+++ b/src/main/java/org/apache/commons/lang3/SerializationUtils.java
@@ -122,7 +122,7 @@ public class SerializationUtils {
      * @since 3.3
      */
     public static <T extends Serializable> T roundtrip(final T msg) {
-        return SerializationUtils.deserialize(SerializationUtils.serialize(msg));
+        return (T) SerializationUtils.deserialize(SerializationUtils.serialize(msg));
     }
 
     // Serialize
--- a/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
@@ -842,7 +842,7 @@ public class DiffBuilder implements Builder<DiffResult> {
         }
 
         // Not array type
-        if (left != null ? left.equals(right) : right.equals(left)) {
+        if (lhs != null ? lhs.equals(rhs) : rhs.equals(lhs)) {
             return this;
         }
 
--- a/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
@@ -842,6 +842,10 @@ public class DiffBuilder implements Builder<DiffResult> {
         }
 
         // Not array type
+        if (left != null ? left.equals(right) : right.equals(left)) {
+            return this;
+        }
+
         diffs.add(new Diff<Object>(fieldName) {
             private static final long serialVersionUID = 1L;
 
--- a/src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
@@ -321,10 +321,9 @@ public class DiffBuilderTest {
         assertEquals(class2.objectField, diff.getRight());                
     }
 
+    /** Test that "left" and "right" are the same instance and are equal. */
     @Test
     public void testObjectsSameAndEqual() throws Exception {
-        // left.objectField == right.objectField
-        // left.objectField.equals( right.objectField )
         final Integer sameObject = 1;
         final TypeTestClass left = new TypeTestClass();
         left.objectField = sameObject;
@@ -337,10 +336,9 @@ public class DiffBuilderTest {
         assertEquals(0, list.getNumberOfDiffs());
     }
 
+    /** Test that "left" and "right" are the same instance but are equal. */
     @Test
     public void testObjectsNotSameButEqual() throws Exception {
-        // left.objectField != right.objectField
-        // left.objectField.equals( right.objectField )
         final TypeTestClass left = new TypeTestClass();
         left.objectField = new Integer(1);
         final TypeTestClass right = new TypeTestClass();
@@ -352,10 +350,9 @@ public class DiffBuilderTest {
         assertEquals(0, list.getNumberOfDiffs());
     }
 
+    /** Test that "left" and "right" are not the same instance and are not equal. */
     @Test
     public void testObjectsNotSameNorEqual() throws Exception {
-        // left.objectField != right.objectField
-        // !left.objectField.equals( right.objectField )
         final TypeTestClass left = new TypeTestClass();
         left.objectField = 4;
         final TypeTestClass right = new TypeTestClass();
--- a/src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
@@ -32,6 +32,31 @@ public class DiffBuilderTest {
     
     private static final ToStringStyle SHORT_STYLE = ToStringStyle.SHORT_PREFIX_STYLE;
 
+    private static class IdentifiableTest {
+        private Integer id;
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == null) { return false; }
+            if (obj == this) { return true; }
+            if (obj.getClass() != getClass()) {
+                return false;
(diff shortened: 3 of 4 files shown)


src/main/java/org/apache/commons/lang3/SerializationUtils.java
```diff
@@ -122,7 +122,7 @@ public SerializationUtils() {
      * @since 3.3
      */
     public static <T extends Serializable> T roundtrip(final T msg) {
-        return SerializationUtils.deserialize(SerializationUtils.serialize(msg));
+        return (T) SerializationUtils.deserialize(SerializationUtils.serialize(msg));
```
britter: Which JDK are you using? What is the output of ```mvn -v```on your system? It should build with maven 3 and JDK 1.6. If it doesn't I'd like to fix this problem in a separate PR, since this change is unrelated to the problem described in LANG-1081.

britter: Building on my machine with JDK 1.6 works:

```
$ mvn -v
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T18:29:23+01:00)
Maven home: /usr/local/Cellar/maven/3.2.5/libexec
Java version: 1.6.0_65, vendor: Apple Inc.
Java home: /System/Library/Java/JavaVirtualMachines/1.6.0.jdk/Contents/Home
Default locale: de_DE, platform encoding: MacRoman
OS name: "mac os x", version: "10.10.1", arch: "x86_64", family: "mac"
```

j--baker:
"mvn -v" output from the old computer:

`Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T12:29:23-05:00) `
`Maven home: /opt/maven`
`Java version: 1.6.0_18, vendor: Sun Microsystems Inc.`
`Java home: /opt/jdk1.6.0_18/jre`
`Default locale: en_US, platform encoding: UTF-8`
`OS name: "linux", version: "2.6.30.10-105.2.23.fc11.x86_64", arch: "amd64", family: "unix"`


Issues during "mvn test" that went away with (T) cast:

- ... commons-lang/src/main/java/org/apache/commons/lang3/SerializationUtils.java:[125,46] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds T,java.lang.Object

- ... commons-lang/src/test/java/org/apache/commons/lang3/exception/AbstractExceptionContextTest.java:[181,55] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds T,java.lang.Object

britter:
Okay, nevertheless we need to split this up because it is unrelated to the issue we're talking about.

j--baker:
The (T) casts were removed from this PR yesterday.

They're in their own [branch](https://github.com/j--baker/commons-lang/commit/ecd0c0bc58145ebe579796199e61e6dbc9009596) now.

New JIRA ticket: [LANG-1083](https://issues.apache.org/jira/browse/LANG-1083)
Will do another PR from new branch against new ticket.

Thanks for your patience!

src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
```diff
@@ -842,6 +842,10 @@ public DiffBuilder append(final String fieldName, final Object lhs,
         }
 
         // Not array type
+        if (left != null ? left.equals(right) : right.equals(left)) {
```
britter: ```left != null``` will always be true, since the constructor prevents the field from getting initialized with null. If that wouldn't be the case, the proposed code could still throw NPE if left and right are both null.

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
```diff
@@ -32,6 +32,31 @@
     
     private static final ToStringStyle SHORT_STYLE = ToStringStyle.SHORT_PREFIX_STYLE;
 
+    private static class IdentifiableTest {
+        private Integer id;
```
britter: Why not ```int```?

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
```diff
@@ -32,6 +32,31 @@
     
     private static final ToStringStyle SHORT_STYLE = ToStringStyle.SHORT_PREFIX_STYLE;
 
+    private static class IdentifiableTest {
```
britter: "Test" isn't such a got suffix for a class used in testing. How about "ObjectId" or something like that?

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
```diff
@@ -321,15 +346,50 @@ public void testObject() throws Exception {
     
     @Test
     public void testObjectsEqual() throws Exception {        
-        final TypeTestClass class1 = new TypeTestClass();
-        final TypeTestClass class2 = new TypeTestClass();
-        class1.objectField = "Some string";
-        class2.objectField = "Some string";
-        final DiffResult list = class1.diff(class2);
-        assertEquals(0, list.getNumberOfDiffs());
+        // Same (left == right).
```
britter: Would be better to create separate test methods for the four blocks you're proposing here.

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
britter: Thinking about this some more, I don't think we need this class at all. We can simply use Integer instances fur der ```objectField``` of ```TypeTestClass```

src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
j--baker: :flushed:  Just added a [patch](https://github.com/j--baker/commons-lang/commit/f1a3e786617e50680589684218dc29ad7650c685) to fix this.  This was supposed to be comparing the fields (lhs, rhs), not the containers (left, right).  The containers are checked in the constructor, but this method is concerned with the fields (after getting past the objectsTriviallyEqual test).  Sorry!

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
j--baker: It's so obvious now that you mention it.  [Patch](https://github.com/j--baker/commons-lang/commit/728ed99c1f21b39f6bae9074112e0ffd31ba7c2b).

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
j--baker: [Done](https://github.com/j--baker/commons-lang/commit/728ed99c1f21b39f6bae9074112e0ffd31ba7c2b) in same patch as IdentifiableTest->Integer changes.

j--baker:
Sorry for screwing up the PR comments.  Just rebased on top of lastest from trunk.

src/main/java/org/apache/commons/lang3/SerializationUtils.java
```diff
@@ -122,7 +122,7 @@ public SerializationUtils() {
      * @since 3.3
      */
     public static <T extends Serializable> T roundtrip(final T msg) {
-        return (T) SerializationUtils.deserialize(SerializationUtils.serialize(msg));
+        return SerializationUtils.deserialize(SerializationUtils.serialize(msg));
```
britter: whoops, why are we removing this again? :-)

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
```diff
@@ -318,18 +320,53 @@ public void testObject() throws Exception {
...
     @Test
-    public void testObjectsEqual() throws Exception {        
-        final TypeTestClass class1 = new TypeTestClass();
-        final TypeTestClass class2 = new TypeTestClass();
-        class1.objectField = "Some string";
-        class2.objectField = "Some string";
-        final DiffResult list = class1.diff(class2);
+    public void testObjectsSameAndEqual() throws Exception {
+        // left.objectField == right.objectField
+        // left.objectField.equals( right.objectField )
```
britter: no commented out code, please

src/test/java/org/apache/commons/lang3/exception/AbstractExceptionContextTest.java
```diff
@@ -178,7 +178,7 @@ public void testGetContextEntries() {
     public void testJavaSerialization() {
         exceptionContext.setContextValue("test Poorly written obj", "serializable replacement");
         
-        final T clone = (T) SerializationUtils.deserialize(SerializationUtils.serialize(exceptionContext));
+        final T clone = SerializationUtils.deserialize(SerializationUtils.serialize(exceptionContext));
```
britter: see above

src/main/java/org/apache/commons/lang3/SerializationUtils.java
j--baker: In the future I'll be more careful when rebasing after splitting changes into separate branches.  The ancient JDK on my other computer thanks you.  :)

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
j--baker: The methods now have prose comments instead.

PR #43, opened 2015-01-21
LANG-1055: StrLookup.systemPropertiesLookup() not singleton.

j--baker:
- Assuming StrLookup.systemPropertiesLookup() should not be a singleton.
- Assuming each lookup should have its own snap-shot of the system properties.

Please review the proposed changes.
Thank you!


britter:
Very nice, thank you!

PR #45, opened 2015-02-19
Replace the busy wait from AtomicSafeInitializer with an alternate method that eliminates wasted CPU cycles

artnaseef:
This patch eliminates the busy-wait in AtomicSafeInitializer that consumes significant CPU cycles (confirmed via test program) for each thread waiting for the initialized result while the initializer runs.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
@@ -56,8 +56,7 @@ import java.util.concurrent.atomic.AtomicReference;
 public abstract class AtomicSafeInitializer<T> implements
         ConcurrentInitializer<T> {
     /** A guard which ensures that initialize() is called only once. */
-    private final AtomicReference<AtomicSafeInitializer<T>> factory =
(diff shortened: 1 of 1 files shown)


artnaseef:
I created a test program that verifies the original code is performing a busy wait; I'll gladly share it.  It may be possible to write a unit test that checks for that as well.

artnaseef:
I'm going to post an alternative that's less of a change and avoids the use of synchronized blocks and notifications.

src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
```diff
@@ -73,12 +72,31 @@
     public final T get() throws ConcurrentException {
         T result;
 
-        while ((result = reference.get()) == null) {
-            if (factory.compareAndSet(null, this)) {
-                reference.set(initialize());
+        synchronized ( reference ) {
```
ecki: I dont think this is a good idea to make a lock free code use such a large synchronized block now.

ecki:
Ah sorry, missed your comment that you re-implement it. :)

PR #47, opened 2015-03-04
swap and shift for arrays

beradrian:
I implemented swap and shift operations for arrays.
The operations works for the entire array or for just regions of them.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -1855,6 +1855,15 @@ public class ArrayUtils {
      * <p>There is no special handling for multi-dimensional arrays.</p>
      *
      * <p>This method does nothing for a {@code null} input array.</p>
+     * 
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap(["1", "2", "3"], 0, 2) -> ["3", "2", "1"]</li>
+     *         <li>ArrayUtils.swap(["1", "2", "3"], 0, 0) -> ["1", "2", "3"]</li>
+     *         <li>ArrayUtils.swap(["1", "2", "3"], 1, 0) -> ["2", "1", "3"]</li>
+     *         <li>ArrayUtils.swap(["1", "2", "3"], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
      *
      * @param array the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
@@ -1873,6 +1882,15 @@ public class ArrayUtils {
      *
      * <p>This method does nothing for a {@code null} input array.</p>
      *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([true, false, true], 0, 2) -> [true, false, true]</li>
+     *         <li>ArrayUtils.swap([true, false, true], 0, 0) -> [true, false, true]</li>
+     *         <li>ArrayUtils.swap([true, false, true], 1, 0) -> [false, true, true]</li>
+     *         <li>ArrayUtils.swap([true, false, true], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
+     * 
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
      * @param offset2 the index of the second element to swap
@@ -1889,6 +1907,15 @@ public class ArrayUtils {
      * <p>Swaps two elements in the given array.</p>
      *
      * <p>This method does nothing for a {@code null} input array.</p>
+     *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
      *
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
@@ -1907,6 +1934,15 @@ public class ArrayUtils {
      *
      * <p>This method does nothing for a {@code null} input array.</p>
      *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
+     * 
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
      * @param offset2 the index of the second element to swap
@@ -1923,6 +1959,15 @@ public class ArrayUtils {
      * <p>Swaps two elements in the given array.</p>
      *
      * <p>This method does nothing for a {@code null} input array.</p>
+     * 
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
      *
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
@@ -1941,6 +1986,15 @@ public class ArrayUtils {
      *
      * <p>This method does nothing for a {@code null} input array.</p>
      *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
+     * 
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
      * @param offset2 the index of the second element to swap
@@ -1958,6 +2012,15 @@ public class ArrayUtils {
      *
      * <p>This method does nothing for a {@code null} input array.</p>
      *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
+     * 
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
      * @param offset2 the index of the second element to swap
@@ -1975,6 +2038,15 @@ public class ArrayUtils {
      *
      * <p>This method does nothing for a {@code null} input array.</p>
      *
+     * <p>Examples:
+     *     <ul>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 2) -> [3, 2, 1]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 0) -> [1, 2, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 1, 0) -> [2, 1, 3]</li>
+     *         <li>ArrayUtils.swap([1, 2, 3], 0, 5) -> ArrayOutOfBoundsException</li>
+     *     </ul>
+     * </p>
+     * 
      * @param array  the array to swap, may be {@code null}
      * @param offset1 the index of the first element to swap
      * @param offset2 the index of the second element to swap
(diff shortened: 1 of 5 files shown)


britter:
Placeholder ticket is https://issues.apache.org/jira/browse/LANG-1099

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+     * </p>
+     * 
+     * @param array
+     *            the array to shift, may be {@code null}
+     * @param startIndexInclusive
+     *            the starting index. Undervalue (&lt;0) is promoted to 0, overvalue (&gt;array.length) results in no
+     *            change.
+     * @param endIndexExclusive
+     *            elements up to endIndex-1 are shiftd in the array. Undervalue (&lt; start index) results in no
+     *            change. Overvalue (&gt;array.length) is demoted to array length.
```
britter: Why does shift behave this way, when indices are out of bounds, but swap simply throws an exception?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+        }        
+        int n = endIndexExclusive - startIndexInclusive;
+        if (n <= 1) {
+        	return;
+        }
+    	offset %= n;
+    	if (offset < 0) {
+    		offset += n;
+    	}
+		while (n > 1 && offset > 0) {
```
britter: no tabs please

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
```diff
@@ -2087,7 +2087,730 @@ public void testReverseObjectRange() {
         assertEquals(null, array);
     }
     
+    //-----------------------------------------------------------------------
```
britter: possible simplification for all of your tests is to create a second array ```expected``` and then use ```assertArrayEquals```. IMHO this would be easier to read, because one could see the expected result of the operation directly instead of reconstructing it from the indices.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
         }
     }
 
+    // Swap
+    //-----------------------------------------------------------------------
+    /**
+     * <p>Swaps two elements in the given array.</p>
```
britter: It's good style to have some examples like so:

```java
ArrayUtils.swap(["1", "2", "3"], 0, 2) = ["3", "2", "1"]
ArrayUtils.swap(["1", "2", "3"], 0, 0) = ["1", "2", "3"]
ArrayUtils.swap(["1", "2", "3"], 1, 0) = ["2", "1", "3"]
ArrayUtils.swap(["1", "2", "3"], 0, 5) = ArrayOutOfBoundsException
```

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
beradrian: I did the tests this way to be in sync with all the other tests in ArrayUtilsTest.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: Which one do you think it will be best? I saw that usually ArrayUtils silently promotes/demotes the indices.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: I replaced all tabs by spaces.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: I added these tests to testSwapChar.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+     *
+     * <p>There is no special handling for multi-dimensional arrays.</p>
+     *
+     * <p>This method does nothing for a {@code null} input array.</p>
+     *
+     * @param array the array to swap, may be {@code null}
+     * @param offset1 the index of the first element to swap
+     * @param offset1 the index of the second element to swap
```
rfalke: Should be offset2.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+     * 
+     * <p>This method does nothing for a {@code null} input array.</p>
+     * 
+     * @param array the array to swap, may be {@code null}
+     * @param offset1 the index of the first element in the series to swap
+     * @param offset1 the index of the second element in the series to swap
+     * @param len the number of elements to swap starting with the given indices
+     * @throws ArrayIndexOutOfBoundsException if one of the indices (plus the length of the series to swap) is out of range
+     */
+    public static void swap(final boolean[] array,  int offset1, int offset2, int len) {
```
rfalke: Here an in general I miss a statement to overlapping array ranges. Suppose I call
swap([1,2,3], 0,1,2)
if the swap with the lowest index is executed first the result would be
[2,3,1]
if the swapping starts with the highest offset the result would be
[3,1,2]

Should two calls with exchanged offsets should yield the same result?
swap([1,2,3], 0,1,2)  == swap([1,2,3],1, 0, 2)

IMHO yes. This implementation does not guarantee this.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+        }        
+        int n = endIndexExclusive - startIndexInclusive;
+        if (n <= 1) {
+            return;
+        }
+        offset %= n;
+        if (offset < 0) {
+            offset += n;
+        }
+        while (n > 1 && offset > 0) {
```
rfalke: It is not easy to see how many times this loop will loop. Together with the loop in the swap method this may give
O(n^2). So this may be expressed in a simpler form.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1868,6 +1868,1013 @@ public static void reverse(final short[] array, final int startIndexInclusive, f
...
+            int n_offset = n - offset;
+            
+            if (offset > n_offset) {
+                swap(array, startIndexInclusive, startIndexInclusive + n - n_offset,  n_offset);
```
rfalke: An alternative would be to use a temporary array instead of doing it in place. Then two arrayCopy-calls (into the temp array) would be needed and one arrayCopy-call to copy it back.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: Fixed.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
britter: Well my feeling is, that invalid indicies should result in no op.

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
britter: No, we should take care of this in a separate issue. Leave it this way for now.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
britter: Sorry, I've been unclear. I was talking about adding some examples to the JavaDoc.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
britter: What advantage do you see in that approach?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: I added these in Javadoc too.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: Actually, in this implementation, swap([1,2,3], 0,1,2) == swap([1,2,3],1, 0, 2) = [2, 3, 1].
It is indeed always with the lowest index first, because the loop in which the swap is made has an increasing index.
I added the test `testSwapIntExchangedOffsets`

src/main/java/org/apache/commons/lang3/ArrayUtils.java
britter: @rfalke like how?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
rfalke: I would suggest the following variant:
```
    public static void shift(final Object[] array, int startIndexInclusive, int endIndexExclusive, int offset) {
        if (array == null) {
            return;
        }
        if (startIndexInclusive >= array.length - 1 || endIndexExclusive <= 0) {
            return;
        }
        if (startIndexInclusive < 0) {
            startIndexInclusive = 0;
        } 
        if (endIndexExclusive >= array.length) {
            endIndexExclusive = array.length;
        }        
        int n = endIndexExclusive - startIndexInclusive;
        if (n <= 1) {
            return;
        }
        offset %= n;
        if (offset < 0) {
            offset += n;
        }
        Object[] copy = new Object[n];
        System.arraycopy(array, startIndexInclusive, copy, 0, n);
        System.arraycopy(copy, 0, array, startIndexInclusive + offset, n - offset);
        System.arraycopy(copy, n - offset, array, startIndexInclusive, offset);
    }
```

I failed to find out how often the while loop in the previous code loops by looking at the 
code. I had to test it out by running it with various input values. This indicated for me 
that the code is hard to understand and therefore hard to reason about (for example 
the runtime characteristics). BTW it turns out that the number of loops depend on the 
offset but n is the upper bound.

There is only one drawback of the variant with the arraycopy calls: it creates an object 
an so add memory pressure on the GC. Still I would prefer the code which is easier 
to understand.

beradrian: Sorry for the delay in answering, but I was busy and I wanted to give a full explanation.

Let me calculate the complexity of the shift algorithm. Let Sh(n, k) be the problem of shifting k positions in an array of size n and Sw(k) be the problem of swapping k elements in an array. 

It is obvious that O(Sw(k)) = O(k).

It is also obvious that O(Sh(1, k)) = O(1), with k < 1. Also O(Sh(x, 0)) = O(1) - we'll use that later.

Now let's assume that O(Sh(n, k)) = O(n), whatever n, with k < n. I'll try to prove that O(Sh(n + 1, k')) = O(n), with k' < n + 1.

Analyzing the algorithm we have
O (Sh(n + 1, k')) = max(
- O(Sw(n + 1 - k')) + O(Sh(k', 2k' - n - 1)), if 2k' > n + 1
- O(Sw(k')) + O(Sh(n + 1 - k', k')), if 2k' < n + 1
- O(Sw(k')), if 2k' == n + 1

First case 2k' > n + 1
 O(Sh(n + 1, k')) = O(Sw(n + 1 - k')) + O(Sh(k', 2k' - n - 1)) = O(n) + O(Sh(k', 2k' - n - 1))
Because k' < n +1 => k' <= n and 2k' <= 2n => 2k' - n - 1 <= n - 1 => 2k' - n - 1 < n, then O (Sh(n + 1, k')) = O(n) + O(n) = O(n).

Second case 2k' < n + 1
O(Sh(n + 1, k')) = O(Sw(k')) + O(Sh(n + 1 - k', k')) = O(k') + O(Sh(n + 1 - k', k')) = O(n) + O(Sh(n + 1 - k', k')) 
Because 0 < k'   => n + 1 - k' < n + 1 => n + 1 - k' <= n and k' < (n + 1)/2 => k' <= n/2 => k' < n, then O(Sh(n + 1, k')) = O(n) + O(n) = O(n).

Third case 2k' == n + 1
O(Sh(n + 1, k')) = O(Sw(k')) = O(n)

So O(Sh(n + 1, k')) = O(n)  (qed).

Am I missing something?

rfalke: Wow what a detailed analysis! I haven't seen such an analysis a long time.

So the code with the loop has for shifting an array of size n the complexity of O(n). The loop less code I propose has the same complexity.

When the code variants are equal in this regard I would prefer the solution which is shorter and easier to understand: the one with System.arraycopy().

What advantages has your variant in your eyes that I may have missed?

beradrian: Indeed, the time complexity for both solutions is O(n). But space complexity for the one with System.arraycopy() is O(n), as opposed with the one with swap which is O(1).

rfalke: Ok so we are speaking about a different space complexity versus a different source 
code complexity. I would prefer the simpler source code. If some user at some point 
has a problem with this they can implement their own optimized version or can ask 
commons-lang. But till then I would prefer the easier code in commons-lang.

beradrian: So, you would prefer a code that seems slightly easier to understand instead of an optimized one? I don't know how to argue this. First time that I hear it. Especially on a library that supposedly should be used in many places and users do not care what's inside. If the API will become cumbersome - maybe, but not the implementation,
Moreover, the code that I wrote is not that hard to understand. And it can be documented to be easier to understand.
But not favoring the most optimized code without any significant drawbacks? Was this decision took on other implementations of commons-lang?

kinow: Hi @rfalke, @beradrian. What an interesting discussion! I even tweeted about it a couple of days ago. Thanks a lot for such interesting discussion.

Regarding this kind of discussion, on whether it is better to have readable or optimized code, the best place to have this kind of discussion is the dev mailing list, as @britter suggested earlier.

I think there's an old thread on this issue from some time ago. If not, any of you guys could drop a message to the mailing list (using the [lang] prefix), mention the JIRA issue ID and this pull request, and explain the current scenario.

Others will chime in, and we may have a consensus on this and complete the work on this issue. What do you think? :smiley:

beradrian: I summarized our discussion and asked others opinion on the dev mailing list. Thanks!

kinow: Excellent @beradrian ! @rfalke , please feel free to chime in and cast your opinion too :)

britter:
@beradrian Commons Lang has been migrated from svn to git. Because of this the main branch has been renamed to 'master'. Can you please rebase your work against the master branch? Thank you!

beradrian:
I rebased my branch against master. Should I create a new pull request?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
beradrian: What do you think it should happen with swap([1, 2, 3, 4, 5], 0, 3, 3)? It should result in no op or it should result in [4, 5, 3, 1, 2], meaning that it will swap as many as possible? I would incline to go for the latter.

britter:
@beradrian I've merged your changes manually but it looks like Github does not recognize this PR as merged. Can you close this PR please? Thank you!

beradrian:
I can close this pull request, but what about the swap and ArrayIndexOutOfBoundsException? I was asking your opinion above before doing any further changes.

britter:
Sorry I must have forgotten about this. What was the problem/question again? :)

beradrian:
You asked why does shift promotes/denotes indices that are out of bounds, but swap simply throws an exception? Your preference was towards no op instead of ArrayIndexOutOfBoundsException. 
But what do you think it should happen with swap([1, 2, 3, 4, 5], 0, 3, 3)? It should result in no op or it should result in [4, 5, 3, 1, 2], meaning that it will swap as many as possible? I would incline to go for the latter.

beradrian:
After we decide how to do it, I can implement it. I think I should close this PR and create a new one, right?

britter:
Hello @beradrian,

sorry for being so oblivious. Yes it would be best to create a separate PR for this issue. I've already created a new jira issue for tracking this: https://issues.apache.org/jira/browse/LANG-1122

PR #58, opened 2015-03-23
LANG-1091 fixes shutdown of ExecutorPools used in Tests.

CodingFabian:


britter:
Hello Fabian,

thank you for the PR. After applying the PR I get:

```
Failed tests:
  FastDateFormatTest.testParseSync:278->measureTime:317 did not complete tasks
```

CodingFabian:
Interesting, seems a to be a race somewhere. will have a look. From the output I assume this was a maven run?

CodingFabian:
fixed, there is a timeout with undocumented purpose. I added a comment to it. Looks like that my machine is to fast and completes the test in a second, while yours doesnt :-)

PR #59, opened 2015-03-27
LANG-821: Added all Mac OS X versions to date

hzpz:
I added constants for all Mac OS X versions to date to SystemUtils and also added corresponding tests. This is my first ever contribution to an ASF project. Let me know if anything is missing and please be gentle ;-)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/SystemUtils.java
+++ b/src/main/java/org/apache/commons/lang3/SystemUtils.java
@@ -1053,8 +1053,8 @@ public class SystemUtils {
      *
      * @since 3.4
      */
+    public static final boolean IS_OS_MAC_OSX_CHEETAH = getOSMatches("Mac OS X", "10.0");
 
-    public static final boolean IS_OS_MAC_OSX_CHEETAH = getOSMajorMinorVersionMatches("Mac OS X", "10.0");
     /**
      * <p>
      * Is {@code true} if this is Mac OS X Puma.
@@ -1065,7 +1065,7 @@ public class SystemUtils {
      *
      * @since 3.4
      */
-    public static final boolean IS_OS_MAC_OSX_PUMA = getOSMajorMinorVersionMatches("Mac OS X", "10.1");
+    public static final boolean IS_OS_MAC_OSX_PUMA = getOSMatches("Mac OS X", "10.1");
 
     /**
      * <p>
@@ -1077,7 +1077,7 @@ public class SystemUtils {
      *
      * @since 3.4
      */
-    public static final boolean IS_OS_MAC_OSX_JAGUAR = getOSMajorMinorVersionMatches("Mac OS X", "10.2");
+    public static final boolean IS_OS_MAC_OSX_JAGUAR = getOSMatches("Mac OS X", "10.2");
 
     /**
      * <p>
@@ -1089,7 +1089,7 @@ public class SystemUtils {
      *
      * @since 3.4
      */
-    public static final boolean IS_OS_MAC_OSX_PANTHER = getOSMajorMinorVersionMatches("Mac OS X", "10.3");
+    public static final boolean IS_OS_MAC_OSX_PANTHER = getOSMatches("Mac OS X", "10.3");
 
     /**
      * <p>
@@ -1101,7 +1101,7 @@ public class SystemUtils {
      *
      * @since 3.4
      */
-    public static final boolean IS_OS_MAC_OSX_TIGER = getOSMajorMinorVersionMatches("Mac OS X", "10.4");
+    public static final boolean IS_OS_MAC_OSX_TIGER = getOSMatches("Mac OS X", "10.4");
 
     /**
      * <p>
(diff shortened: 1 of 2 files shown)


britter:
Hello @hzpz,

thank you for this PR.

> please be gentle ;-)

sounds like you've made bad experiences in the past. I'll try my best ;-) Please have a look at my comments.

src/main/java/org/apache/commons/lang3/SystemUtils.java
```diff
@@ -1033,6 +1033,138 @@
...
      * <p>
+     * Is {@code true} if this is Mac OS X Cheetah.
+     * </p>
+     * <p>
+     * The field will return {@code false} if {@code OS_NAME} is {@code null}.
+     * </p>
+     *
+     * @since 3.4
+     */
+
```
britter: This empty line should probably be below the field declaration.

src/main/java/org/apache/commons/lang3/SystemUtils.java
```diff
@@ -1433,6 +1569,26 @@ static boolean isJavaVersionMatch(final String version, final String versionPref
...
+     * @param osVersion the actual OS version
+     * @param osNamePrefix the prefix for the expected OS name
+     * @param osVersionPrefix the expected major and minor OS version, e.g. "10.10"
+     * @return true if matches, or false if not or can't determine
+     */
+    static boolean isOSMajorMinorVersionMatch(final String osName, final String osVersion, final String osNamePrefix, final String osVersionPrefix) {
+        if (osName == null || osVersion == null) {
+            return false;
+        }
+        return osName.startsWith(osNamePrefix) && osVersion.matches(osVersionPrefix + "(\\.\\d+)?");
```
britter: How is this different from ```isOSMatch(final String osName, final String osVersion, final String osNamePrefix, final String osVersionPrefix)``` ? Looks rather complicated to use are regex instead of simply ```String.startsWith(String)```

src/test/java/org/apache/commons/lang3/SystemUtilsTest.java
```diff
@@ -373,6 +373,33 @@ public void testOSMatchesNameAndVersion() {
...
+        assertTrue(SystemUtils.isOSMajorMinorVersionMatch(osName, osVersion, "Mac OS X", "10.1"));
+        osVersion = "10.10.1";
+        assertTrue(SystemUtils.isOSMajorMinorVersionMatch(osName, osVersion, "Mac OS X", "10.10"));
+        osVersion = "10.10.10";
+        assertTrue(SystemUtils.isOSMajorMinorVersionMatch(osName, osVersion, "Mac OS X", "10.10"));
+    }
+
+
+    @Test
+    public void testNotConfuseMacOSXPumaAndYosemite() {
```
britter: I don't understand why we need this test case. It won't do anything on most machines.

britter: Ah, thinking about this some more I now understand why you couldn't use ```String.startsWith(String)```. Because 10.1 and 10.10 have the same prefix. Well that is a tricky one. I'm not sure I like the proposed solution. Maybe we can make ```isOSMatch```smarter?

britter: Have a look at https://issues.apache.org/jira/browse/LANG-1102 and http://svn.apache.org/r1669750, I've implemented a better algorithm for matching OS version. This way it should be easy for you to implement the Mac OS X versions.

hzpz:
Thanks @britter for the quick review. It didn't even hurt ;-)

I like the new isOSVersionMatch approach. But you may have broken Windows 98 and Windows ME detection (and possibly more). According to [this list](http://lopica.sourceforge.net/os.html) they identify themselfs as versions 4.10 and 4.90 ([Wikipedia](https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions) suggests the same). This would have worked with the startsWith approach, but it will fail using the new method. However, since there is [LANG-794](https://issues.apache.org/jira/browse/LANG-794) anyway (which I plan on working on, too) that won't be a problem for long.

The isOSVersionMatch method will work perfectly fine for the Mac OS X versions though. What is the preferred approach to submit my updated code? Should I merge the changes you made to my fork and update this pull request or should I close this pull request and create a new one?

hzpz:
As you can see, I went ahead and updated the pull request. If you would rather have me create a new one, let me know.

src/main/java/org/apache/commons/lang3/SystemUtils.java
```diff
@@ -1479,7 +1611,7 @@ static boolean isOSNameMatch(final String osName, final String osNamePrefix) {
         }
         return osName.startsWith(osNamePrefix);
     }
-
```
hzpz: Sorry about that.

src/main/java/org/apache/commons/lang3/SystemUtils.java
```diff
@@ -1479,7 +1611,7 @@ static boolean isOSNameMatch(final String osName, final String osNamePrefix) {
         }
         return osName.startsWith(osNamePrefix);
     }
-    
+
```
hzpz: Sorry about that.

PR #60, opened 2015-03-28
LANG-794: Use os.name only when determining Windows versions

hzpz:
As suggested in previous JIRA comments and according to some other sources (see http://lopica.sourceforge.net/os.html and http://hg.openjdk.java.net/jdk7u/jdk7u/jdk/file/tip/src/windows/native/java/lang/java_props_md.c for example) the Windows version can be determined by os.name alone. I also added a constant for Windows Server 2012.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/SystemUtils.java
+++ b/src/main/java/org/apache/commons/lang3/SystemUtils.java
@@ -1150,7 +1150,7 @@ public class SystemUtils {
      *
      * @since 2.0
      */
-    public static final boolean IS_OS_WINDOWS_2000 = getOSMatches(OS_NAME_WINDOWS_PREFIX, "5.0");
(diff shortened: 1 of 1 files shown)


britter:
Looks good to me! I'll integrate this as soon as I have the time. In the mean time: do you want to be listed in the contributors section of pom.xml? If so, what is the name you want to be listed with?

hzpz:
Well, I haven't really done much (yet) but sure, I'd like that! My full name is Timo Kockert.

britter:
@hzpz oh you have done a lot. For example you prevented Commons Lang 3.4 to be released with the regression I introduced during my refactoring of SystemUtils (http://markmail.org/message/bbtjpzekwr4jlahm).

I'll have some time this evening to include your changes. Keep up the good work!

PR #66, opened 2015-04-20
fix JsonToStringStyle

jacktan1991:
except number types and boolean type , all other types in java (char, String, Date etc.) should be `appendAsString`. 
+ add method `appendDetail(StringBuffer buffer, String fieldName, char value)` to `JsonToStringStyle`
+ modify `appendDetail(StringBuffer buffer, String fieldName, Object value)`

### DIFF
--- a/src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.fail;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Date;
 
 import org.apache.commons.lang3.SystemUtils;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
```diff
@@ -92,6 +93,39 @@ public void testAppendSuper() {
...
     @Test
+    public void testChar() {
+    
+        try {
+            new ToStringBuilder(base).append('A').toString();
+            fail("Should have generated UnsupportedOperationException");
+        } catch (UnsupportedOperationException e) {
+        }
+        
+        assertEquals("{\"a\":\"A\"}", new ToStringBuilder(base).append("a", 'A')
```
jacktan1991: In Json, `"a":A` is illeagle, `"a":"A"` is the right way

PR #67, opened 2015-04-20
fix JsonToStringStyle

jacktan1991:
except number types and boolean type , all other types in java (char, String, Date etc.) should be `appendAsString`.

add method `appendDetail(StringBuffer buffer, String fieldName, char value)` to `JsonToStringStyle`
modify `appendDetail(StringBuffer buffer, String fieldName, Object value)`


jacktan1991:
I just modified two java files, I don't konw why **Failed**.
https://github.com/apache/commons-lang/pull/67/files

britter:
@jacktan1991 Commons Lang has been migrated from svn to git. Because of this the main branch has been renamed to 'master'. Can you please rebase your work against the master branch? Thank you!

PR #68, opened 2015-04-23
LANG-1118 Fix StringUtils.repeat(char, int)

rikles:
Now doing what is said in JavaDoc comment :
when passing a negative repeat value to the StringUtils.repeat(char, int) function, it returns an empty String instead of throwing a NegativeArraySizeException.


britter:
@rikles Commons Lang has been migrated from svn to git. Because of this the main branch has been renamed to 'master'. Can you please rebase your work against the master branch? Thank you!

rikles:
Close due to rebase on master

PR #73, opened 2015-04-28
#LAN-1114 fixes bug in TypeUtils.equals(WildcardType, Type)

datalorax:
fixes bug in TypeUtils.equals(WildcardType, Type) where it was incorrectly returning true when the second argument was not a Wildcard type.


datalorax:
Who's of thought so many issues with such a tiny fix!  Thanks for your patience @britter :D

PR #74, opened 2015-04-29
fix JsonToStringStyle

jacktan1991:
except number, boolean, Array and nested object structures, other types like char, String, Date etc. in java should be `appendAsString`.

add method `appendDetail(StringBuffer buffer, String fieldName, char value)` to `JsonToStringStyle`
modify `appendDetail(StringBuffer buffer, String fieldName, Object value)`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
@@ -2573,13 +2573,28 @@ public abstract class ToStringStyle implements Serializable {
                 return;
             }
 
-            if (value instanceof Number || value.getClass() == Boolean.class) {
+            if (value instanceof String || value instanceof Character) {
 
-                buffer.append(value);
+            	appendValueAsString(buffer, value.toString());
                 return;
             }
-
-            appendValueAsString(buffer, value.toString());
+            
+            if (value instanceof Number || value instanceof Boolean) {
+            
+            	buffer.append(value);
+            	return;
+            }
+            
+            final String try_value_str = value.toString();
+            if (try_value_str.startsWith(getContentStart()) && try_value_str.endsWith(getContentEnd())
+                    || try_value_str.startsWith(getArrayStart()) && try_value_str.startsWith(getArrayEnd())
+                ) {
+                
+            	buffer.append(value);
+            	return;
+            }
+            
+            appendDetail(buffer, fieldName, try_value_str);
         }
 
         /**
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
```diff
@@ -90,6 +90,39 @@ public void testAppendSuper() {
...
+        assertEquals("{\"a\":\"A\"}", new ToStringBuilder(base).append("a", 'A')
+                .toString());
+        assertEquals("{\"a\":\"A\",\"b\":\"B\"}", new ToStringBuilder(base).append("a", 'A').append("b", 'B')
+                .toString());
+    }
+    
+    @Test
+    public void testDate() {
+        
+        final Date now = new Date();
```
britter: This fails on Travis because this import is missing.

jacktan1991: Thanks!! Have fixed it.

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -2568,13 +2573,13 @@ protected void appendDetail(StringBuffer buffer, String fieldName, Object value)
...
-            if (value.getClass() == String.class) {
+            if (value instanceof Number || value.getClass() == Boolean.class) {
 
-                appendValueAsString(buffer, (String)value);
+                buffer.append(value);
                 return;
             }
 
-            buffer.append(value);
+            appendValueAsString(buffer, value.toString());
```
britter: I don't think this is correct for nested object structures. Suppose we have the following class structure:

```java
public class Person {
   private String name;
   private Address address;
}

public class Address {
   private String city;
   private String zip;
}
```

Translating this to JavaScript/JSON should look like this if both, Person and Address use the JsonToStringStyle for their equals method:

```JavaScript
{
   "name": "Peter",
   "address": {
      "city": "New York",
      "zip": "10026"
   }
}
```

If we apply this fix, the result will look like this:

```JavaScript
{
   "name": "Peter",
   "address": "{
      "city": "New York",
      "zip": "10026"
   }"
}
```

Note the additional double quotes around the address object.

jacktan1991: Yes. But if `Address` like you suppose here does not *Override* `Object.toString()` (maybe someone just wants Address instance represented by its hash code), then its json style looks like `"address":"Address@1ab2f62"`.

britter:
Hello @jacktan1991,

thank you for this contribution. I had to do some clean ups, see 07ec8ad2dd862c8669b2557b995ec5180663d73f, 28656559d2fc67193a84d5cbc48d378b2016f74a and e8db7237a9c347d2225fce9b49ff92bb741adffa. Please have a look at [CONTRIBUTING.md](CONTRIBUTING.md) if you intend to contribute on a regular basis (which would be awesome! ;-) )

PR #76, opened 2015-05-03
LANG-1127: Use JUnit rules to set and reset the default Locale and TimeZone

britter:
Alternative proposal to fix [LANG-1127](https://issues.apache.org/jira/browse/LANG-1127) using JUnit TestRules.


chonton:
Overall, I like this approach better than explicit classes.  Two items I would like to see:
1. Allow tests to run in parallel.  This will probably requires some sort of synchronization to allow only one tests at a time to change default TimeZone or Locale.
2. Apply the rule to individual test methods instead of to all methods in a test class.  Perhaps this could be done by creating annotation(s) which enable the Rule.  I am thinking of something like SwitchDefaults(locale="locale_name", timezone="timezone_name").

britter:
Hello @chonton,

very good feedback, thank you!

> 1. Allow tests to run in parallel. This will probably requires some sort of synchronization to allow only one tests at a time to change default TimeZone or Locale.

Good point. Should be possible by wrapping the code in the apply method with some synchronization logic.

> 2. Apply the rule to individual test methods instead of to all methods in a test class. Perhaps this could be done by creating annotation(s) which enable the Rule. I am thinking of something like SwitchDefaults(locale="locale_name", timezone="timezone_name").

This is already possible by calling explicitly the setLocale method on the TestLocale instance. The idea is to set up a default for all tests and only switch it for tests when needed. However this doesn't make much sense in the light of parallel test execution...

britter:
@chonton I've worked some more on this. Renamed the Rule to SystemDefaultsSwitch and wrote a unit test. What do you think about integrating this now and create a new PR for the synchronization?

chonton:
Iterate fast and often. Go forward with it. 

Chas

PR #78, opened 2015-05-03
Add ThreadUtils (2nd PR)

salyh:
Add org.apache.commons.lang3.ThreadUtils which provide helper methods related to java.lang.Thread (2nd PR)

This PR replaces https://github.com/apache/commons-lang/pull/61

ThreadUtils class exposes the following static methods:
findThreadById(long)
findThreadById(long, String)
findThreadById(long, ThreadGroup)
findThreadGroups(ThreadGroup, boolean, ThreadGroupPredicate)
findThreadGroups(ThreadGroupPredicate)
findThreadGroupsByName(String)
findThreads(ThreadGroup, boolean, ThreadPredicate)
findThreads(ThreadPredicate)
findThreadsByName(String)
findThreadsByName(String, String)
findThreadsByName(String, ThreadGroup)
getAllThreadGroups()
getAllThreads()
getSystemThreadGroup()

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ThreadUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ThreadUtils.java
@@ -126,10 +126,16 @@ public class ThreadUtils {
             throw new IllegalArgumentException("The thread group name must not be null");
         }
 
+        final Collection<ThreadGroup> threadGroups = findThreadGroups(new NamePredicate(threadGroupName));
+
+        if(threadGroups.isEmpty()) {
+            return Collections.EMPTY_LIST;
+        }
+
         final Collection<Thread> result = new ArrayList<Thread>();
-        NamePredicate threadNamePredicate = null;
-        for(final ThreadGroup group : findThreadGroups(new NamePredicate(threadGroupName))) {
-            result.addAll(findThreads(group, false, threadNamePredicate==null?(threadNamePredicate=new NamePredicate(threadName)):threadNamePredicate));
+        final NamePredicate threadNamePredicate = new NamePredicate(threadName);
+        for(final ThreadGroup group : threadGroups) {
+            result.addAll(findThreads(group, false, threadNamePredicate));
         }
         return Collections.unmodifiableCollection(result);
     }
@@ -264,7 +270,7 @@ public class ThreadUtils {
      */
     //if java minimal version for lang becomes 1.8 extend this interface from java.util.function.Predicate
     public static interface ThreadGroupPredicate /*extends java.util.function.Predicate<ThreadGroup>*/{
-        
+
         /**
          * Evaluates this predicate on the given threadgroup.
          * @param threadGroup the threadgroup
@@ -295,7 +301,6 @@ public class ThreadUtils {
         public boolean test(@SuppressWarnings("unused") final Thread thread) {
             return true;
         }
-
     }
 
     /**
@@ -307,7 +312,7 @@ public class ThreadUtils {
 
         /**
          * Predicate constructor
-         * 
+         *
          * @param name thread or threadgroup name
          * @throws IllegalArgumentException if the name is {@code null}
          */
@@ -339,7 +344,7 @@ public class ThreadUtils {
 
         /**
          * Predicate constructor
-         * 
+         *
          * @param threadId the threadId to match
          * @throws IllegalArgumentException if the threadId is zero or negative
          */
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/ThreadUtils.java
```diff
@@ -0,0 +1,459 @@
...
+        if (threadName == null) {
+            throw new IllegalArgumentException("The thread name must not be null");
+        }
+        if (threadGroupName == null) {
+            throw new IllegalArgumentException("The thread group name must not be null");
+        }
+
+        final Collection<Thread> result = new ArrayList<Thread>();
+        for(final ThreadGroup group : findThreadGroups(new NamePredicate(threadGroupName))) {
+            result.addAll(findThreads(group, false, new NamePredicate(threadName)));
```
peteryhwong: The NamePredicate object is the same in every iteration. Hence only one instance is required outside of the for loop.

src/main/java/org/apache/commons/lang3/ThreadUtils.java
```diff
@@ -0,0 +1,459 @@
...
+     * @throws  SecurityException
+     *          if the current thread cannot access the system thread group
+     *
+     * @throws  SecurityException  if the current thread cannot modify
+     *          thread groups from this thread's thread group up to the system thread group
+     */
+    public static Thread findThreadById(final long threadId) {
+        final Collection<Thread> result = findThreads(new ThreadIdPredicate(threadId));
+
+        if(!result.iterator().hasNext()) {
```
peteryhwong: result.isEmpty() can be used instead to avoid creating an iterator instance.

src/main/java/org/apache/commons/lang3/ThreadUtils.java
```diff
@@ -0,0 +1,459 @@
...
+     */
+    public static Collection<Thread> findThreads(final ThreadGroup group, final boolean recurse, final ThreadPredicate predicate) {
+        if (group == null) {
+            throw new IllegalArgumentException("The group must not be null");
+        }
+        if (predicate == null) {
+            throw new IllegalArgumentException("The predicate must not be null");
+        }
+
+        final List<Thread> result = new ArrayList<Thread>();
```
peteryhwong: If result is defined after the do while loop, it can be initialised with count as the initial capacity, thereby avoid the potential additional array allocation/copy when invoking add().

src/main/java/org/apache/commons/lang3/ThreadUtils.java
```diff
@@ -0,0 +1,461 @@
...
+            throw new IllegalArgumentException("The thread name must not be null");
+        }
+        if (threadGroupName == null) {
+            throw new IllegalArgumentException("The thread group name must not be null");
+        }
+
+        final Collection<Thread> result = new ArrayList<Thread>();
+        NamePredicate threadNamePredicate = null;
+        for(final ThreadGroup group : findThreadGroups(new NamePredicate(threadGroupName))) {
+            result.addAll(findThreads(group, false, threadNamePredicate==null?(threadNamePredicate=new NamePredicate(threadName)):threadNamePredicate));
```
CodingFabian: why not move the init of the threadName predicate to 130? is it lazy in case there are no results?
I personally find it hard to read, maybe assigning findThreadGroups() to a local variable and checking it is more readable and still efficient?

src/main/java/org/apache/commons/lang3/ThreadUtils.java
```diff
@@ -0,0 +1,461 @@
...
+            throw new IllegalArgumentException("The group must not be null");
+        }
+        if (predicate == null) {
+            throw new IllegalArgumentException("The predicate must not be null");
+        }
+        
+        int count = group.activeGroupCount();
+        ThreadGroup[] threadGroups;
+        do {
+            threadGroups = new ThreadGroup[count + (count>>1) + 1];
```
CodingFabian: why the magic right shift? is this based on experience?

CodingFabian:
Thank you @salyh for your work on this. Overall I like the API and the implementation. I added two comments.
Also it looks like your formatter settings are not used. some spaces are missing (like i=0 instead of i = 0)

britter:
@CodingFabian @salyh agreed. Very good work! Let's resolve the last two comments from Fabian and then we can integrate it.

src/main/java/org/apache/commons/lang3/ThreadUtils.java
salyh: this was a suggestion from @chonton. IMHO its better (than using *2) because array size does not grow so fast

CodingFabian: ok, maybe add a comment, so that a maintainer in 3 years will not say WTF? :)

salyh: replaced the magic with something more readable (count/2), should to the same

PR #81, opened 2015-05-05
LANG-1135 - Add method containsAllWords to WordUtils

eduardomartins-luizalabs:
Checks if the String contains all words.
```java
StringUtils.containsAllWords("lorem ipsum dolor sit amet", "ipsum", "lorem", "dolor") == true
StringUtils.containsAllWords("lorem ipsum dolor sit amet", "ipsum", null, "lorem", "dolor") == false
StringUtils.containsAllWords("ab", "b") == false
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1685,7 +1685,7 @@ public class StringUtils {
      * StringUtils.containsAny("", *)              = false
      * StringUtils.containsAny(*, null)            = false
      * StringUtils.containsAny(*, [])              = false
-     * StringUtils.containsAny("abcd", "ab", "cd") = false
+     * StringUtils.containsAny("abcd", "ab", "cd") = true
      * StringUtils.containsAny("abc", "d", "abc")  = true
      * </pre>
      *
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -16,8 +16,11 @@
  */
 package org.apache.commons.lang3.text;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
+
+import java.util.regex.Pattern;
 
 /**
  * <p>Operations on Strings that contain words.</p>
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -1685,7 +1685,7 @@ public static boolean containsAny(final CharSequence cs, final CharSequence sear
      * StringUtils.containsAny("", *)              = false
      * StringUtils.containsAny(*, null)            = false
      * StringUtils.containsAny(*, [])              = false
-     * StringUtils.containsAny("abcd", "ab", "cd") = false
+     * StringUtils.containsAny("abcd", "ab", "cd") = true
```
agrava-luizalabs: It is good to point out that the documentation was wrong in this case

britter:
This would better fit into ```org.apache.commons.lang3.text.WordUtils```, because it deals with words. I could imagine having a StringUtils.containsAll(CharSequence, CharSequence...) method that just checks if all of the given CharSequences are all contained in the first CharSequence, but without handling word boundaries.

eduardomartins-luizalabs:
@britter i didn't saw this class, my bad.
Should I send another commit or amend this?

britter:
Usually we make additional commit so that the development of a feature can be tracked afterwards. Since in this case the commit would change the whole PR, I think amending is appropriate in this case.

eduardomartins-luizalabs:
Ok

src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -16,21 +16,16 @@
...
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-
-import java.util.Locale;
-
 import org.apache.commons.lang3.test.DefaultLocale;
 import org.hamcrest.core.IsNot;
 import org.junit.Test;
 
+import java.util.Locale;
```
eduardomartins-luizalabs: Sorry about that, Ide issue.
I will fix.

eduardomartins-luizalabs:
@britter updated.

britter:
Your placeholder Jira issue is https://issues.apache.org/jira/browse/LANG-1135. Please create a Jira for bigger contributions as documented in CONTRIBUTING.md. Thank you!

britter:
Hello @eduardomartins-luizalabs,

the following will fail:

```java
assertFalse(WordUtils.containsAllWords("lorem ipsum null dolor sit amet", "ipsum", null, "lorem", "dolor"));
```

This is because you only check for nulls at index 0 of the input array. IMHO null is not contained in any string. Have a look at for example ```StringUtils.containsAny(CharSequence, CharSequence...)``` and adc6bc4104514063b3e8ebe7a301d5dd4bce36d3. Further more I wonder whether we should let this method operate on CharSequences. WDYT?

eduardomartins-luizalabs:
Sorry about the JIRA card.
I think it's a great improvement, i will do that.

eduardomartins-luizalabs:
@britter updated.

PR #84, opened 2015-05-05
@SuppressWarnings("deprecation") to joinWith, because it uses the

PascalSchumacher:
deprecated ObjectUtils.toString(Object) method (which is o.k. to use as
long as commons-lang3 does not require java 7 or greater)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -4207,7 +4207,9 @@ public class StringUtils {
 
         final Iterator<Object> iterator = Arrays.asList(objects).iterator();
         while (iterator.hasNext()) {
-            result.append(ObjectUtils.toString(iterator.next()));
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4196,7 +4196,8 @@ public static String join(final Iterable<?> iterable, final String separator) {
      * @return the joined String.
      * @throws java.lang.IllegalArgumentException if a null varargs is provided
      */
-    public static String joinWith(final String separator, final Object... objects) {
+    @SuppressWarnings("deprecation")
```
britter: It's better to keep the scope of this annotation as small as possible. I'd rather assign the result of {{ObjectUtils.toString(iterator.next())}} and add the annotation there.

PascalSchumacher: good point

PascalSchumacher:
Thanks for merging! :)

PR #85, opened 2015-05-06
ReflectionToStringBuilder doesn't throw IllegalArgumentException when the constructor's object param is null

jacktan1991:
As described in it's [javadoc](http://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/builder/ReflectionToStringBuilder.html#ReflectionToStringBuilder(java.lang.Object)), ReflectionToStringBuilder constructor will throw IllegalArgumentException if the Object to build a toStringfor is null, while in fact it won't.

I had create the Issue [LANG-1132](https://issues.apache.org/jira/browse/LANG-1132) for this.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
@@ -1041,7 +1041,7 @@ public class ToStringBuilderTest {
         static final int staticInt2 = 67890;
     }
 
-    @Test
+    @Test(expected=IllegalArgumentException.class)
     public void testReflectionNull() {
         assertEquals("<null>", ReflectionToStringBuilder.toString(null));
     }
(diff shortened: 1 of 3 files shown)


britter:
Please fix test failures.

jacktan1991:
Have fixed it. But `ToStringBuilderTest.testReflectionNull` is desigined like

```java
@Test(expected=IllegalArgumentException.class)
    public void testReflectionNull() {
        assertEquals("<null>", ReflectionToStringBuilder.toString(null));
    }
```

so now I'm really confused wether ReflectionToStringBuilder constructor should throw IllegalArgumentException when Object passed in is null.

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
```diff
@@ -1041,7 +1041,7 @@ public void test_setUpToClass_invalid() {
         static final int staticInt2 = 67890;
     }
 
-    @Test
+    @Test(expected=IllegalArgumentException.class)
```
britter: This change is correct. The JavaDoc of the toString method says it throws an IllegalArgumentException for null inputs, but apparently it doesn't.

PR #87, opened 2015-05-10
Lang 1134

Lady-Stardust:
LANG-1134

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Validate.java
+++ b/src/main/java/org/apache/commons/lang3/Validate.java
@@ -45,6 +45,20 @@ import java.util.regex.Pattern;
  */
 public class Validate {
 
+    private static final String DEFAULT_NOT_NAN_EX_MESSAGE = 
+        "The validated value is not a number";
+    private static final String DEFAULT_FINITE_EX_MESSAGE =
+        "The value is invalid: %f";
+    private static final String DEFAULT_GREATER_EX_MESSAGE =
+        "The value %s is not greater than %s";
+    private static final String DEFAULT_GREATER_OR_EQUAL_EX_MESSAGE =
+        "The value %s is not greater than or equal to %s";
+    private static final String DEFAULT_SMALLER_EX_MESSAGE =
+        "The value %s is not smaller than %s";
+    private static final String DEFAULT_SMALLER_OR_EQUAL_EX_MESSAGE =
+        "The value %s is not smaller than or equal to %s";
+    private static final String DEFAULT_DIFFERENT_EX_MESSAGE =
+        "The value %s is invalid";
     private static final String DEFAULT_EXCLUSIVE_BETWEEN_EX_MESSAGE =
         "The value %s is not in the specified exclusive range of %s to %s";
     private static final String DEFAULT_INCLUSIVE_BETWEEN_EX_MESSAGE =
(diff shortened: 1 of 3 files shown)


britter:
Very nice, please see my comments.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+     * @param value  the value to validate
+     * @param message  the {@link String#format(String, Object...)} exception message if invalid, not null
+     * @param values  the optional values for the formatted exception message
+     * @throws IllegalArgumentException if the value is not a number
+     * @see #notNaN(double) 
+     * 
+     * @since 3.4
+     */
+    public static void notNaN(final double value, final String message, final Object... values) {
+        if (value != value) {
```
britter: Better use ```Double.isNaN(double)``` here. Reads better, IMHO.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+     * @param value the value to validate
+     * @param message  the {@link String#format(String, Object...)} exception message if invalid, not null
+     * @param values  the optional values for the formatted exception message
+     * @throws IllegalArgumentException if the value is infinite or {@code NaN}
+     * @see #finite(double) 
+     * 
+     * @since 3.4
+     */
+    public static void finite(final double value, final String message, final Object... values) {
+        if (value != value || Double.isInfinite(value)) {
```
britter: Better use ```Double.isNaN(double)``` ;)

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+     * 
+     * @param <T>  the type of the argument object
+     * @param min  the reference value
+     * @param value  the object to validate
+     * @throws IllegalArgumentException if {@code value} is smaller than or equal to {@code min}
+     * @see #greaterObj(java.lang.Object, java.lang.Comparable, java.lang.String, java.lang.Object...) 
+     * 
+     * @since 3.4
+     */
+    public static <T> void greaterObj(final T min, final Comparable<T> value) {
```
britter: Why not simply ```public static <T> void greater(final T min, final Comparable<T> value)```?

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+     * @param min  the reference value
+     * @param value  the object to validate
+     * @param message  the {@link String#format(String, Object...)} exception message if invalid, not null
+     * @param values  the optional values for the formatted exception message
+     * @throws IllegalArgumentException if {@code value} is smaller than or equal to {@code min}
+     * @see #greaterObj(java.lang.Object, java.lang.Comparable) 
+     * 
+     * @since 3.4
+     */
+    public static <T> void greaterObj(final T min, final Comparable<T> value, final String message, final Object... values) {
```
britter: see above

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+     * 
+     * @param <T>  the type of the argument object
+     * @param min  the reference value
+     * @param value  the object to validate
+     * @throws IllegalArgumentException if {@code value} is smaller than {@code min}
+     * @see #greaterOrEqualObj(java.lang.Object, java.lang.Comparable, java.lang.String, java.lang.Object...) 
+     * 
+     * @since 3.4
+     */
+    public static <T> void greaterOrEqualObj(final T min, final Comparable<T> value) {
```
britter: Again, the "Obj" can be dropped.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -874,7 +888,752 @@ public static void matchesPattern(final CharSequence input, final String pattern
...
+        if (!(value <= max)) {
+            throw new IllegalArgumentException(String.format(message, values));
+        }
+    }
+    
+    // different
+    //---------------------------------------------------------------------------------
+    
+    /**
+     * <p>Validates that the specified argument is not equal to a given value
```
britter: The wording here is a bit confusing since "not equal" usually means ```value.equals(reference) == false```. Maybe ```smallerOrGreater``` would be a better method name?

src/test/java/org/apache/commons/lang3/ValidateTest.java
```diff
@@ -1015,6 +1708,9 @@ public void testExclusiveBetweenDouble_withMessage()
         }
     }
 
+    //-----------------------------------------------------------------------
```
britter: Where are these changes coming from?

src/main/java/org/apache/commons/lang3/Validate.java
Lady-Stardust: Okay, I wanted to avoid having more method calls as the Validate methods might get called often in a program.
Should I edit it and make a new pull request? How does it work?

src/main/java/org/apache/commons/lang3/Validate.java
Lady-Stardust: (1)- If I call the method 'greater' I'm creating an ambiguous definition as longs and double can be boxed into comparable objects.

(2)- I agree about swapping the arguments, I wanted to do it the other way around from the start, but if you look at the methods inclusiveBetween and exclusiveBetween, they take the reference values first. I thought it'd be better to make my methods homogeneous with those. Can you confirm that you want me to make this change?

I don't have a fix for point (1).

src/main/java/org/apache/commons/lang3/Validate.java
Lady-Stardust: smallerOrGreater seems like a less natural name to me. Would there be a way to rephrase the documentation in a better way?
Also isn't ""value.equals(reference) == false"" equivalent to ""value.compareTo(reference) != 0"" in the general case?

src/test/java/org/apache/commons/lang3/ValidateTest.java
Lady-Stardust: I might have moved my methods and left that behind, my bad.

src/main/java/org/apache/commons/lang3/Validate.java
britter: Hello @Lady-Stardust,

calling a method is not that expensive (approximately 2ns), so we can live with that ;-)

You can simply add more commits to your branch and github will update the PR when you push your changes to your fork. Thank you!

Lady-Stardust:
Ok I pushes the changes to my fork. I made the changes you said, except for removing the "Obj" suffix because it wouldn't compile anymore.

britter:
Very nice, I'll have a look later this week!

britter:
@Lady-Stardust, what is the name you want to be listed with in [changes.xml](https://github.com/apache/commons-lang/blob/master/src/changes/changes.xml)?

Lady-Stardust:
I don't want to be listed

PR #88, opened 2015-05-11
Added new addListener method that allows checking for duplicate listeners.

ZenHarbinger:
Added new addListener() method that takes a boolean for if duplicates are allowed.

Added test method to make sure that duplicate listeners are not added.


britter:
@ZenHarbinger thank you, and don't forget to submit a JIRA for your changes ;-)

PR #89, opened 2015-05-15
LANG-1115 Add support for varargs in method utils

jimlloyd:
@britter I have been collaborating with Joe Ferner on this contribution and have prepared this PR which I believe addresses the feedback you provided in [LANG-1115](https://issues.apache.org/jira/browse/LANG-1115). I'm not entirely sure I am following all of you contribution guidelines so please let me know if I need to resubmit. I will also add a comment to the Jira issue.

FYI, the first three commits in this PR replicate the original patch submission provided by Joe. The following seven commits address your feedback.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/ConstructorUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/ConstructorUtils.java
@@ -18,6 +18,7 @@ package org.apache.commons.lang3.reflect;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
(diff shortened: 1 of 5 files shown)


cfranzen:
I have played around a little bit with that pull request here and I think I have found an issue with the current implementation. When building the varArgs array that will be used in method.invoke() the method System.arraycopy() is used. Unfortunately this method will not handle auto-(un)boxing correctly. So the following example will not work:

```java
class TestBean{
   public void doSomething(int... values){
      ...
   }
}

MethodUtils.invokeMethod(testBean, Integer.valueOf(1), Integer.valueOf(2));
```

My expectation would be that MethodUtils is able to handle that. Not sure if my expectation is wrong.

PR #91, opened 2015-05-28
[LANG-1138]add a static method to create a started stopwatch

WalterInSH:
Basically, I very next step after I new a stopwatch is starting it. Why don't we add a static method to get a started stopwatch. Save our strength.


WalterInSH:
Henri Yandell has fixed this issues

PR #93, opened 2015-06-01
fix StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 3

goni-hgs:
fix StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 3
to StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 2

worked with Gyucheol Park, Wookju Ha and Sanghee Lee of coupang.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
@@ -16,18 +16,15 @@
  */
 package org.apache.commons.lang3;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-
-import java.util.Locale;
-
-import org.apache.commons.lang3.test.SystemDefaultsSwitch;
 import org.apache.commons.lang3.test.SystemDefaults;
+import org.apache.commons.lang3.test.SystemDefaultsSwitch;
 import org.hamcrest.core.IsNot;
 import org.junit.Rule;
 import org.junit.Test;
+
+import java.util.Locale;
+
+import static org.junit.Assert.*;
 
 /**
  * Unit tests {@link org.apache.commons.lang3.StringUtils} - Substring methods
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -16,19 +16,16 @@
  */
 package org.apache.commons.lang3;
 
-import static org.junit.Assert.assertEquals;
```
britter: Please don't reorganize imports.

britter:
Shouldn't this result in ` StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 1` because the sequence 'aa' is at every index?

rousej:
I agree with britter.  StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 1 would be true because the sequence 'aa' is at every index. There seems to be confusion around this method, but it seems to me the original code had it correct.

........
int index = lastIndex ? str.length() : INDEX_NOT_FOUND;  
do {  
    if (lastIndex) {  
       index = CharSequenceUtils.lastIndexOf(str, searchStr, index - 1);  
   } else {  
       index = CharSequenceUtils.indexOf(str, searchStr, index + 1);  
   }  
........

I'm not sure why StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 3 would ever be true.
StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 2 is easier to see where it's coming from, but StringUtils.ordinalIndexOf("aaaaaa", "aa", 2) == 1 is still the correct answer, and it also works for other tests. I found my way here because I wanted to use this method in a project, but found that the current release is broken by commit e5a3039. With the current release if the first index is between 0 and searchStr.length() -1 the method will return the index for ordinal + 1.....in other words the wrong index.

PascalSchumacher:
I guess this pull request can be closed, now that https://github.com/apache/commons-lang/commit/e5a3039f7a1e727fca40db7357a9191b6a7cf41d has been reverted with https://github.com/apache/commons-lang/commit/d75fe46b8f1b0d5c27887052ee4714d6a9c7ea4b and additional tests and javadoc have been added with https://github.com/apache/commons-lang/commit/7fd021d82ff431fb31f42bc6c5c44a3b979cb426

PR #94, opened 2015-06-06
RFC: provide support for generating random length strings within an inclusive range

xenoterracide:
this patch isn't really intended for immediate merge, but proof of concept to see if patches to the random* methods that allowed a range to be specified instead of a total count would be accepted, and what changes before doing a lot of work on it would be required if any.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -95,6 +95,21 @@ public class RandomStringUtils {
     public static String randomAlphabetic(final int count) {
         return random(count, true, false);
     }
+
+    /**
+     * <p>Creates a random string whose length is the number of characters
+     * specified.</p>
+     *
+     * <p>Characters will be chosen from the set of alphabetic
+     * characters.</p>
+     *
+     * @param minLength the minimum length of random string to create
+     * @param maxLength the maximum length of random string to create
+     * @return the random string
+     */
+    public static String randomAlphabetic(final int minLength, final int maxLength) {
+        return random( RandomUtils.nextInt( minLength, maxLength ), true, false);
+    }
     
     /**
      * <p>Creates a random string whose length is the number of characters
--- a/src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
@@ -16,18 +16,18 @@
  */
 package org.apache.commons.lang3;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
```diff
@@ -16,18 +16,18 @@
  */
 package org.apache.commons.lang3;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import org.junit.Test;
```
britter: Please don't reorganize imports.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -95,6 +95,21 @@ public static String randomAscii(final int count) {
     public static String randomAlphabetic(final int count) {
         return random(count, true, false);
     }
+
+    /**
+     * <p>Creates a random string whose length is the number of characters
+     * specified.</p>
```
britter: Dosn't exactly describe what the method does. More like `Creates a random string of length between minLength and maxLength`.

britter:
Can you describe a use case for this functionality?

src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
xenoterracide: yeah sorry IDE did that, I decided to not correct it whilst waiting to see if this would be functionality the project would accept.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
xenoterracide: correct, copy pasta, oops, can fix.

xenoterracide:
I use methods like this (I've written for Numeric and AlphaNumeric as well) for generating test data. At work it's being used in our selenium tests, at home I use it for any tests. It's not meant to be used to tests the bounds, rather that any behavior in the bounds that is expected works. Also that there aren't hidden gotchas, like something fails because a string has a `.` or a `-` because someone decided to parse user input.

Managers actually watch our selenium tests replay, and they discovered that our Zip Code fields, perhaps improperly, take any alphanumeric string.

I think my first use case was testing that passwords worked between 8-20 characters.

Obviously much of that can be accomplished by using a static length random string, but there are a few things you can catch by using a random length, like improper or forgotten bounds, display length issues, weird sort issues, and of course it makes for just a little more random distribution of the data.

    protected void wireEntity( final JpaUser entity ) {
        entity.setName( RandomUtil.randomAlphanumeric( 5, 15 ) );
        entity.setEnabled( true );
    }

PR #98, opened 2015-06-19
Ability to throw checked exceptions without declaring them

gzak:
One of Java's most controversial features is checked exceptions, and most other languages&mdash;including newer JVM-based languages (Scala, Clojure, etc.)&mdash;have dispensed with them. Check out this [fantastic interview with Anders Hejlsberg](http://www.artima.com/intv/handcuffs.html) (creator of C#) on why checked exceptions are a bad idea.

This utility gives users a way to dispense with checked exceptions even in Java, in a minimally intrusive fashion. By standardizing such a utility and giving it widespread adoption, hopefully the folks over at Oracle will eventually give in and drop the feature in a later release of Java. Luckily, it wouldn't be backwards incompatible to do so.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
+++ b/src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
@@ -1,0 +1,35 @@
+package org.apache.commons.lang3.exception;
+
+/**
+ * @author <a href="mailto:thegzak@gmail.com">Gregory Zak</a>
+ *
+ * <p>Utility class containing a method to allow checked exceptions to be rethrown without explicitly
+ * declaring them.</p>
+ *
+ */
+public final class ThrowUtils {
+
+    private ThrowUtils() {
+    }
+
+    /**
+     * <p>Any method with uninteresting checked exceptions in the body can simply be wrapped in a
+     * try/catch block, and the exceptions can then be freely rethrown from the catch block using
+     * {@code rethrow(t)}. If a method requires a return value, simply use {@code return rethrow(t)}
+     * in the catch block and compiler type inference will do the rest (no need to manually capture
+     * a return variable in the try block and return it after the catch block).</p>
+     *
+     * @param throwable The throwable to rethrow
+     * @param <R> Used only for compiler type inference
+     * @return Never returns, always throws
+     */
+    public static <R> R rethrow(Throwable throwable) {
+        return ThrowUtils.<R, RuntimeException>rethrow0(throwable);
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <R, T extends Throwable> R rethrow0(Throwable throwable) throws T {
+        throw (T) throwable;
+    }
+
+}
--- a/src/test/java/org/apache/commons/lang3/exception/ThrowUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/exception/ThrowUtilsTest.java
@@ -1,0 +1,30 @@
+package org.apache.commons.lang3.exception;
+
+import org.junit.Test;
+
+import static org.apache.commons.lang3.exception.ThrowUtils.rethrow;
+
+/**
+ * @author <a href="mailto:thegzak@gmail.com">Gregory Zak</a>
+ */
+public class ThrowUtilsTest {
+
+    @Test(expected = Exception.class)
+    public void testVoid() {
+        rethrow(new Exception());
+    }
+
+    @Test(expected = Exception.class)
+    public void testReturning() {
+        Object obj = helper();
+    }
+
+    private static Object helper() {
+        try {
+            throw new Exception();
+        } catch (Exception e) {
+            return rethrow(e);
+        }
+    }
+
+}


src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.commons.lang3.exception;
+
+/**
+ * @author <a href="mailto:thegzak@gmail.com">Gregory Zak</a>
```
kinow: @author tag needs to be removed (project policy)

sjamesr:
Just for reference I think this is similar to http://docs.guava-libraries.googlecode.com/git/javadoc/src-html/com/google/common/base/Throwables.html in Guava

src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param throwable The throwable to rethrow
+     * @param <R> Used only for compiler type inference
+     * @return Never returns, always throws
+     */
+    public static <R> R rethrow(Throwable throwable) {
+        return ThrowUtils.<R, RuntimeException>rethrow0(throwable);
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <R, T extends Throwable> R rethrow0(Throwable throwable) throws T {
```
kinow: A JavaDoc comment is necessary or it will add a CheckStyle error.

src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @author <a href="mailto:thegzak@gmail.com">Gregory Zak</a>
+ *
+ * <p>Utility class containing a method to allow checked exceptions to be rethrown without explicitly
+ * declaring them.</p>
+ *
+ */
+public final class ThrowUtils {
+
+    private ThrowUtils() {
```
kinow: A JavaDoc comment is necessary or it will add a CheckStyle error.

src/test/java/org/apache/commons/lang3/exception/ThrowUtilsTest.java
```diff
@@ -0,0 +1,30 @@
+package org.apache.commons.lang3.exception;
```
kinow: Missing license header.

src/main/java/org/apache/commons/lang3/exception/ThrowUtils.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.commons.lang3.exception;
```
kinow: Missing license header.

kinow:
Checked out the pull request locally, it compiles and the tests pass. The minor checkstyle errors, the author tag and the missing license headers have to be fixed; you can either update the pull request or a committer will do so when the code is merged.

As @sjamesr commented it looks like Throwables#propagate from Guava. I'm +1 for the feature, though I'm not sure about the class name. Will leave it open for a while so that others can review it too.

kinow:
Place holder ticket created in JIRA https://issues.apache.org/jira/browse/LANG-1149

Feel free to write to the Commons Dev Mailing list (use the [lang] prefix) about your issue if this pull request stays idle for too long or if you want to discuss the implementation. Thank you for your contribution.

jochenw:
@sjamesr, @kinow: I think, there is a fundamental difference to Guavas propagate method: The latter wraps the Exception in a RuntimeException, if it must be exposed in the signature. This one doesn't. (Beats me, how it works, to be honest.)

netomi:
It works (if you can call it like that) by casting it to something this it isn't, which results in a ClassCastException. As a ClassCastException is a RuntimeException everything appears to be fine, but it is completely wrong.
The referenced util in guava makes sense, and if you like to include it in commons, just copy&paste it, but do not try to come up with a flawed version yourself.

jochenw:
@netomi So, in other words: It doesn't do what's documented, doesn't it? Or, to rephrase that again: Rejected,

netomi:
The code as is can only be rejected. The question is if commons-lang wants to include the same functionality as guava, which makes sense imho.

kinow:
@gzak feel free to update the pull request & comment here or in JIRA, otherwise we may close it as the current implementation needs rework.

Nice catch @netomi @jochenw !

gzak:
@netomi it is not flawed, but it is tricky to understand. There is no ClassCastException, the original exception does indeed get thrown, even if it is checked. Check out the unit test, it expects an Exception, not a ClassCastException.

It exploits type erasure to trick the compiler into permitting a checked exception to be thrown "as if" it were a RuntimeException. That cast operation you see in the private method ultimately gets erased, because the generic type argument itself gets erased. But it does the job of telling the compiler that it will be a RuntimeException, even if it's not.

It's a very clever workaround (I cannot claim credit, so I'm not bragging, I'm complimenting).

netomi:
My bad, I missed that the type argument will get erased, but in fact this makes it even worse. With this method you can outwit the compiler, throwing a checked exception further up the stack, leading to situations where code that would never expect a checked exception force to handle it.

I would prefer to explicitly throw the checked exception wrapped in a RuntimeException as guava does in its Throwables class.

gzak:
Why though? If you're using one of those methods, you're essentially saying "I don't care about this error, just raise it to the top-level handler". At the top, you generally just catch Throwables and log the error without actually doing anything (presumably, all the finally blocks along the way took care of recovery). This just cleans up the stack trace by removing one layer of unnecessary wrapping.

Another thing to keep in mind: Scala and other newer JVM languages already disregard checked exceptions, so you can already find yourself in situations where a checked exception gets to your top-level handler in Java when interfacing with other libraries (I've encountered it myself). This just brings Java closer to parity with other more modern languages.

gzak:
Give the interview I posted a read: one key point is to assume users actually care about the exception itself, when in reality they generally don't. This leads to tons of code which just appends a "throws Throwable" clause to method signatures, which defeats the purpose of the feature. Or worse, silently swallowing exceptions just to make them go away, leading to really hard to find issues. Our code base, for instance, is littered with throws Throwable clauses, which also gets rid of checked exceptions, but also adds a lot of unnecessary clutter.

netomi:
If scala has been designed like that, then developers know what to expect. Java was designed differently and there is tons of code out there that does a try {} catch (RuntimeException). If you allow checked exceptions to propagate as is you violate the java language specification and create potentially more problems than you are trying to solve.

btw. try {} catch (Throwable) is discouraged in various environments as it might also catch things that application code should not handle, like ThreadDead or any other error.

gzak:
Generally you don't do catch (RuntimeException) though, right? I would panic slightly if I saw that, to be honest, since that could be NullPointerException or DivideByZero, etc, which is generally masking a bug.

What will the handler for Guava's propagate method look like in practice? If you're using it, you are signaling that the exception doesn't matter, so just a catch (Exception) at the top will do (if you want to avoid Error).

It's true that this circumvents a Java language feature, but I would argue that the feature itself does more harm than good, so circumventing it will restore some semblance of order. Also, by making this technique more prevalent, it's more likely that Java will drop the feature in the future. There's two aspects: one is somewhat philosophical, which is to stay in tune with the design of the language. The other is practical, which appeals to the day-to-day costs of this design. I'm certainly trading philosophy for pragmatism here, but I would argue that the latter is more important. It's what the troops on the ground have to deal with every day.

Those are my final comments, I will accept whatever action you take at this point, as I understand both sides (had the same debate with a colleague).

netomi:
This technique is used already elsewhere, e.g. in Android (https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/java/libcore/util/SneakyThrow.java), okio (https://github.com/square/okio/blob/master/okio/src/main/java/okio/Util.java), lombok (https://projectlombok.org/features/SneakyThrows.html).

There are basically two patterns that I found where this would be acceptable and is actually used:

 * in code which offers a strict interface, like Runnable.run() but we know for sure that any thrown exception will be caught (if executed within a Thread for example).
 * in cleanup code, that first collects some throwable, and at the end rethrows it, but the method itself does declare the checked exception, Example:

```java
// The following code must enumerate several types to rethrow:
public void close() throws IOException {
    Throwable thrown = null;
    ...
    if (thrown != null) {
        if (thrown instanceof IOException) {
            throw (IOException) thrown;
        } else if (thrown instanceof RuntimeException) {
            throw (RuntimeException) thrown;
        } else if (thrown instanceof Error) {
            throw (Error) thrown;
        } else {
            throw new AssertionError();
        }
    }
}

// With SneakyThrow, rethrowing is easier:
public void close() throws IOException {
    Throwable thrown = null;
    ...
    if (thrown != null) {
        SneakyThrow.sneakyThrow(thrown);
    }
}
```

I am not sure if there are other use-cases that would be acceptable, but alone the fact that it is a dangerous technique and should only used with great care is a good indication to not include it in commons-lang.

jochenw:
@netomi: "I would prefer to explicitly throw the checked exception wrapped in a RuntimeException as guava does in its Throwables class."

That is (IMO) a valid point. OTOH, I'd bet my house on the fact that there are users out there who will take a different stance.

In other words: That doesn't mean that we must reject the stuff. If we document, what problems may, or may not, arise, then let the users decide whether to use that stuff or not.

And, besides: Nothing prevents us to add another method, which will use an UndeclaredThrowableException, or whatever, as a wrapper, and which you and me will likely prefer.

gzak:
So what's the verdict? Is there anything I should address?

jochenw:
I'd suggest the following:

- Fix missing license headers.
- Add a second method, which works like Throwables.propagate (returning, or throwing, a RuntimeException)
- Add documentation, which compares both methods and lists reasons, why one would prefer either over the other (this part being the most important.
- Change the tests in a manner, that demonstrates that we can control not only the Exception class to be thrown, but the instance as well. Roughly like this:

    OutOfMemoryError ooe = new OutOfMemoryError();
    Throwable th = null;
    try {
       Throwables.propagate(ooe);
       Assert.fail("Expected exception");
    } catch (OutOfMemory e) {
       th = e;
    }
   assertSame(ooe, th);

kinow:
+1 @jochenw. Do you think it would be a good idea to start a discussion thread in the dev mailing list as well?

jochenw:
kinow: Another one? No, don't think so. Being a committer gives the right to push in changes, if one thinks, they are right/acceptable. If others disagree, there's room for discussion later on.

chonton:
Refactored and committed.  See https://github.com/apache/commons-lang/commit/59022fb870c2c45a27e00943003c5acdeddaeec3 and https://github.com/apache/commons-lang/commit/03fe88ab71cf2eadaa39654786c3fba713742768

PascalSchumacher:
@gzak Can you please close the issue? Thanks!

PR #101, opened 2015-06-28
random length strings within range

xenoterracide:
at a high level, I have added  `randomGraph( int count )` and `randomPrint( int count )` which map to their corresponding regular expression classes. These are useful if you're needing to test arbitrary user input in various kinds of text fields. `randomAscii` was not sufficient due to it's inclusion of the DEL character.

I have also added variants of random Ascii, graph, print, alphabetic, alphanumeric, and numeric that have  `minimumInclusive` and `maximumInclusive` parameters. These can be used to create string of random length to ensure more random distribution of data when testing. The will also ensure that developers who use them are less likely to forget testing variable lengths within the bounds.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -81,6 +81,20 @@ public class RandomStringUtils {
     public static String randomAscii(final int count) {
         return random(count, 32, 127, false, false);
     }
+
+    /**
+     * <p>Creates a random string whose length is between the minimum and maximum.</p>
+     *
+     * <p>Characters will be chosen from the set of characters whose
+     * ASCII value is between {@code 32} and {@code 126} (inclusive).</p>
+     *
+     * @param minInclusive the minimum length of the string to generate
+     * @param maxInclusive the maximum length of the string to generate
+     * @return the random string
+     */
+    public static String randomAscii(final int minInclusive, final int maxInclusive  ) {
+        return randomAscii(RandomUtils.nextInt(minInclusive, maxInclusive + 1));
+    }
     
     /**
      * <p>Creates a random string whose length is the number of characters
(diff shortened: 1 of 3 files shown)


kinow:
#94, #100 and #101 are useful for generating random data for (for example) testing. I thought about suggesting these methods to the incubating [text] component; but on second thought, I think it could be better to have these features in a separate component.

Just food for thought, but maybe something like a commons for testing, like https://github.com/andygibson/datafactory, but also including general code for helping Java developers to write test.

xenoterracide:
I think the added `randomGraph( count )` and `randomPrint( count )` might be ok to stay here, since they resemble the existing methods so much. I can see putting the ranged variants somewhere else, I'm not sure if they're *only* useful for testing, that's just what I've used them for (well that and dumby output for creating new vaadin components).

Willing to move things if that decision is made.

PascalSchumacher:
+1 I would like to use these methods :)

xenoterracide:
https://bitbucket.org/xenworks/datafactory/src/ I've started work on a "replacement" for datafactory and this here, feedback welcome

PascalSchumacher:
I still think these methods should be added to commons-lang. They are a very useful extensions of RandomStringUtils feature set.

I created https://issues.apache.org/jira/browse/LANG-1224 and https://issues.apache.org/jira/browse/LANG-1225 to track them.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -81,6 +81,20 @@ public static String random(final int count) {
...
+     * <p>Creates a random string whose length is between the minimum and maximum.</p>
+     *
+     * <p>Characters will be chosen from the set of characters whose
+     * ASCII value is between {@code 32} and {@code 126} (inclusive).</p>
+     *
+     * @param minInclusive the minimum length of the string to generate
+     * @param maxInclusive the maximum length of the string to generate
+     * @return the random string
+     */
+    public static String randomAscii(final int minInclusive, final int maxInclusive  ) {
```
PascalSchumacher: I guess the intent would be even clearer if the parameters were called `minLengthInclusive` and `maxLengthInclusive`. What do you think?

PascalSchumacher:
To quote Gary Gregory from https://issues.apache.org/jira/browse/LANG-1224

>It seems really confusing that some new APIs would use an inclusive max while existing APIs use an exclusive max.

Current APIs with exclusive max:

    random(int, int, int, boolean, boolean)
    random(int, int, int, boolean, boolean, char...)
    random(int, int, int, boolean, boolean, char[], Random)

So for the sake of consistency with the existing methods and the java random api, an exclusive max should be used.

PascalSchumacher:
Thanks!

PR #102, opened 2015-07-02
Changes in Capitalize methods in WordUtils class.

kaching88:
This commit do some refactorization and improvements in Capitalization section methods. I removed overloaded methods with null parameter and add some code to prevent duplications in delimeters parameters. Commit passes all tests.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -1788,11 +1788,11 @@ public class ArrayUtils {
      * @param array
      *            the array to reverse, may be {@code null}
      * @param startIndexInclusive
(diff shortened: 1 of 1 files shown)


joehni:
You cannot simply remove methods, because it breaks binary compatibility. All you can do is deprecating them now and remove them in the next major version.

kaching88:
Removed methods was a overloads which points to capitalize and uncapitalize method. As we can see capitalize and uncapitalize methods have (String,char...) params and removed overloads methods had (String) and call (String,char..) by (String, null). Actually (String, char...) can behave as (String) so removing those overloads won't break anything. And last but not least you cant do any change without deleting those overloads due stackoverflow error.

joehni:
It does! Binary compatibility means, that you can take the new jar file and replace it in an existing project without recompile the sources. This is what Apache Common promises for its components. Therefore is is *not* possible to remove those methods, because existing classes will call the methods with the single argument and will break badly if it is missing.

PR #104, opened 2015-07-06
Changes in Capitalize methods in WordUtils class.

kaching88:
Some minor changes in capitalize and uncapitalize methods. I annotated useless methods as depreciated, do some refactorization and duplication elimination in delimeters varargs. This is second pull request after some useful advices from community.


joehni:
Your changes introduce the conversion of the char array into a HashSet and back again into an array. This is way too expensive. Keep the original array and iterate through it. If you want to check for duplicates, you may create a 2nd array and add each processed character. Then you can check any new delimieter character for the index in the 2nd array first.

kaching88:
Yes...You have right. I completely removed duplication elimination as too much overhead even with an Array.

PR #107, opened 2015-07-20
LANG-1155 - String unwrap capability

MSaifAsif:
LANG-1155 - Unwrapping a string from a character set


PascalSchumacher:
Thanks! :+1:

PR #108, opened 2015-08-21
LANG-781 Added methods to ObjectUtils class to check for null element…

vanta:
I've created this pull request to refresh the issue LANG-781 as I faced a lack of methods to check nulls in array.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -128,6 +128,45 @@ public class ObjectUtils {
         return null;
     }
 
+    /**
+     * <p></p>Checks if any value in the array is not {@code null}.
+     * If all the values are {@code null} or the array is {@code null}
+     * or empty then {@code false} is returned. Otherwise {@code true} is returned.</p>
+     *
+     * @param values  the values to test, may be {@code null} or empty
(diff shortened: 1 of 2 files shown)


vanta:
Any update on this?

garydgregory:
Yes:
- Thank you for the patch.
- The Javadoc has some bogus text, like starting with "<p></p>". You do not need that.
- Do not use a for-each loop to traverse an array. This creates an iterator when none is needed. Use a normal for loop, which will not create the extra Iterator to GC. This matters when a libraries is used on the critical path and/or environments trying to do low/no GC.

PascalSchumacher:
@vanta It would be nice if you could update the pull request according to garys comments.

Thanks!

vanta:
@garydgregory How bout `firstNonNull` method that exists since 2009 and uses for-each loop to iterate over array?

garydgregory:
There is a discussion going on right now on the dev mailing list about for-each vs. for. Let's see how that plays out. See https://commons.apache.org/mail-lists.html

PascalSchumacher:
Thanks! :+1: 

For each loop won the discussion, so I merged only the first commit.

@vanta Sorry for telling you to refactor to use a for loop.

vanta:
@PascalSchumacher Nothing to sorry for :)

PR #110, opened 2015-10-05
LANG-1171 Add null safe compare methods in StringUtils :

rikles:
Add null safe methods in `StringUtils` to compare 2 Strings :
```java
public static int compare(final String str1, final String str2);
public static int compare(final String str1, final String str2, final boolean nullIsLess);
public static int compareIgnoreCase(final String str1, final String str2);
public static int compareIgnoreCase(final String str1, final String str2, final boolean nullIsLess);
```
Those methods are null safe equivalent to :
```java
str1.compareTo(str2);
str1.compareToIgnoreCase(str2);
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
  *      - checks if a String contains text</li>
  *  <li><b>Trim/Strip</b>
  *      - removes leading and trailing whitespace</li>
- *  <li><b>Equals</b>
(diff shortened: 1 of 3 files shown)


rikles:
This unit test fails :
```
org.apache.commons.lang3.StringUtilsTest.testStringUtilsCharSequenceContract()
```

It said :
> The method public static int org.apache.commons.lang3.StringUtils.compare(String,String,boolean) appears to be immutable in spirit and therefore must not accept a String

I use String parameters instead of CharSequence because in the Java conception, only String are Comparable and not CharSequence.
I call `String#compareTo(String)` in the implementation.

So what should I do ?

PR #112, opened 2015-10-19
LANG-1175: removed ant-based build

wikier:


.gitignore
```diff
@@ -4,6 +4,7 @@ target
 maven-eclipse.xml
 build.properties
 site-content
+*~
```
britter: What does this mean?

.gitignore
```diff
@@ -4,6 +4,7 @@ target
 maven-eclipse.xml
 build.properties
```
britter: Can be dropped from .gitignore

src/changes/changes.xml
```diff
@@ -22,6 +22,7 @@
   <body>
 
   <release version="3.5" date="tba" description="tba">
+    <action issue="LANG-1175" type="fix" dev="wikier" dute-to="Benedikt Ritter">Remove Ant-based build</action>
```
britter: Copy and paste error. Should be `due-to`. I've corrected this for the other lines.

britter:
Looks good to me. But we should also drop:
* default.properties
* src/release-tools (it also depends on the main ant build)

.gitignore
wikier: commonly some editors (e.g., emacs) use the suffix `~` to create backup files

wikier:
updated from some comments by @britter

src/assembly/src.xml
```diff
@@ -32,7 +32,6 @@
                 <include>pom.xml</include>
                 <include>PROPOSAL.html</include>
                 <include>RELEASE-NOTES.txt</include>
-                <include>build.xml</include>
                 <include>default.properties</include>
```
britter: Should also be dropped

britter:
Looks good to me now. Feel free to merge to master.

wikier:
I don't have permissions on the main git repo, so please @britter merge it.

PR #115, opened 2015-11-10
Fix indexOfIgnoreCase javadoc and provide test for fixed example

Spikhalskiy:
Now javadoc for StringUtils#indexOfIgnoreCase contains example
```
StringUtils.indexOfIgnoreCase("abc", "", 9) = 3
```
This result is very unintuitive and incorrect - with current code it will be -1. Provided test for this example and fixed example's result in javadoc.


garydgregory:
Patch applied. Thank you.

PR #118, opened 2015-11-26
Optimization of getLevenshteinDistance with a threshold

jontejj:
If the string lengths differ more than the threshold,
there's no need for the algoritm to begin allocating
arrays etc.


PascalSchumacher:
I created a jira issue to track this: https://issues.apache.org/jira/browse/LANG-1234

PascalSchumacher:
Thanks! :+1:

PR #120, opened 2015-12-22
modify note at line 1230

BarkZhang:
add a missing right parenthesis


garydgregory:
In Git master, please verify and close. Offically tracked here: https://issues.apache.org/jira/browse/LANG-1200

PR #127, opened 2016-02-29
General cleanup

pivovarit:
I removed some of the unnecessary code constructs that were driving me mad for some time. I have some ideas for a new methods, but I will prepare separate PRs to make things clean

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BooleanUtils.java
+++ b/src/main/java/org/apache/commons/lang3/BooleanUtils.java
@@ -63,7 +63,7 @@ public class BooleanUtils {
         if (bool == null) {
             return null;
         }
-        return bool.booleanValue() ? Boolean.FALSE : Boolean.TRUE;
+        return bool ? Boolean.FALSE : Boolean.TRUE;
     }
 
     // boolean Boolean methods
@@ -155,7 +155,7 @@ public class BooleanUtils {
      * @return {@code true} or {@code false}, {@code null} returns {@code false}
      */
     public static boolean toBoolean(final Boolean bool) {
-        return bool != null && bool.booleanValue();
+        return bool != null && bool;
     }
 
     /**
@@ -175,7 +175,7 @@ public class BooleanUtils {
         if (bool == null) {
             return valueIfNull;
         }
-        return bool.booleanValue();
+        return bool;
     }
 
     // Integer to Boolean methods
@@ -238,7 +238,7 @@ public class BooleanUtils {
         if (value == null) {
             return null;
         }
-        return value.intValue() == 0 ? Boolean.FALSE : Boolean.TRUE;
+        return value == 0 ? Boolean.FALSE : Boolean.TRUE;
     }
 
     /**
@@ -426,7 +426,7 @@ public class BooleanUtils {
         if (bool == null) {
             return null;
         }
-        return bool.booleanValue() ? NumberUtils.INTEGER_ONE : NumberUtils.INTEGER_ZERO;
+        return bool ? NumberUtils.INTEGER_ONE : NumberUtils.INTEGER_ZERO;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


garydgregory:
Hello,

Thank you for your interest in Apache Commons Lang.

Changes like this are not clean ups because it will cause an unboxing compiler warning.

@@ -465,7 +465,7 @@ public static int toInteger(final Boolean bool, final int trueValue, final int f
         if (bool == null) {
             return nullValue;
         }
-        return bool.booleanValue() ? trueValue : falseValue;
+        return bool ? trueValue : falseValue;
     }

Would you be willing to update your patch?

Thank you!

pivovarit:
Well, sure :)

pivovarit:
I will prepare another PR in order to avoid unnecessary commit, ok? Or is this fine?

garydgregory:
I like to look at patch files, which ever way produced. ;-)

pivovarit:
@garydgregory so I guess this should be fine :)

garydgregory:
Sadly, this kind of change is not a "Good Thing":

@@ -7266,14 +7245,14 @@ public static int indexOfDifference(final CharSequence... css) {
         // find the min and max string lengths; this avoids checking to make
         // sure we are not exceeding the length of the string each time through
         // the bottom loop.
-        for (int i = 0; i < arrayLen; i++) {
-            if (css[i] == null) {
+        for (CharSequence cs : css) {
+            if (cs == null) {
                 anyStringNull = true;
                 shortestStrLen = 0;
             } else {
                 allStringsNull = false;
-                shortestStrLen = Math.min(css[i].length(), shortestStrLen);
-                longestStrLen = Math.max(css[i].length(), longestStrLen);
+                shortestStrLen = Math.min(cs.length(), shortestStrLen);
+                longestStrLen = Math.max(cs.length(), longestStrLen);
             }
         }

That is because the compiler converts use an Iterator to traverse an array with a for-each loop which is slower than traversing it with index access. Rule of thumb: don't use for-each over an array when performance matters. In low-level libraries like Commons, performance usually matters.

Please review your patch and remove such changes.

Thank you!

pivovarit:
@garydgregory well, I did not realize stuff like this matter in practice. I will have done this by tomorrow

pivovarit:
@garydgregory PR updated. I thought that those manual iterations are just left-overs from pre-java1.5 era ;)

garydgregory:
Ok, I'll look tomorrow. In the meantime, can you provide a diff file for all of this?

markt-asf:
Just add .diff to the end of the pull request URL

pivovarit:
https://patch-diff.githubusercontent.com/raw/apache/commons-lang/pull/127.diff

Wow, it works indeed. Thanks @markt-asf

pivovarit:
Are you guys planning on doing something with it or can I close it?

garydgregory:
@pivovarit:
One more thing: Please do not remove 'final' from anything. Send me another diff and I'll review again.
Thank you,
Gary

pivovarit:
@garydgregory Why is it important to maintain "finality" of static methods?

garydgregory:
Final methods is a bigger discussion than what this ticket suggests. It's NOT a code style clean up.

For a general discussion on final, but not specifically of final methods, please see https://garygregory.wordpress.com/2013/01/26/the-final-kiss-in-java/

pivovarit:
@garydgregory I am aware of the importance of the `final` keyword in the context your blogpost refers to, but this is something different. Static methods are implicitly final. Adding `final` keyword will only prevent those methods from being hidden in subclasses. Is this what you want to achieve?

garydgregory:
It is what it is for whatever reason the original author intended, that person must have thought it important for subclasses to NOT be able to hide the methods. The class itself should probably be final like any utility class, but we cannot change that in a minor release since it would break BC.

To cut to the chase: I am OK with apply the patch if it does not include the 'final' changes.

pivovarit:
@garydgregory consider it done

PR #135, opened 2016-04-24
Add tests for some missed branches in StringUtils

caseyscarborough:
This pull request adds tests for some missed branches in `StringUtils`.

Tests that the `join` methods (`long[]`, `char[]`, `byte[]`, `int[]`, `float[]`, `double[]`, and `short[]`) return `null` and `EMPTY` when appropriate.

Also tests that `IllegalArgumentException` is properly thrown from `replaceEach` when `searchList` and `replacementList` are different lengths.

Updates line coverage from 99% to 100% in and branch coverage from 97% to 98%.


PascalSchumacher:
Thanks! :+1:

PR #141, opened 2016-05-13
Lang 1195: Enhance MethodUtils to allow invocation of private methods

Derek-Ashmore:
Currently, MethodUtils is restricted to finding and invoking accessible methods. Frequently, developers have a need to test 'private' methods. What I see is that they escalate access to 'protected' in order to more easily provide test coverage for these methods. From a design perspective, this is bad.

I propose to enhance MethodUtils so that it can easily invoke private methods. I'm not suggesting that developers should do this in production code, merely test code. Much as FieldUtils provides access to private fields via the 'forceAccess' overload on many of its methods. I've copied a utility like this around for years. It would be much more convenient to simply include it with Commons Lang. See [LANG-1195](https://issues.apache.org/jira/browse/LANG-1195) for details.

I made sure that I used space indentation and tried to adhere closely to your standards.  I've also signed the Contributor License Agreement.

Assuming you like the enhancement, I'm willing to contribute additional work if you see issues with what I've done.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
@@ -93,6 +93,29 @@ public class MethodUtils {
             IllegalAccessException, InvocationTargetException {
         return invokeMethod(object, methodName, ArrayUtils.EMPTY_OBJECT_ARRAY, null);
     }
+    
+    /**
+     * <p>Invokes a named method without parameters.</p>
+     *
+     * <p>This is a convenient wrapper for
+     * {@link #invokeMethod(Object object,String methodName, Object[] args, Class[] parameterTypes)}.
+     * </p>
+     *
+     * @param object invoke method on this object
+     * @param forceAccess force access to invoke method even if it's not accessible
+     * @param methodName get method with this name
+     * @return The value returned by the invoked method
+     *
+     * @throws NoSuchMethodException if there is no such accessible method
+     * @throws InvocationTargetException wraps an exception thrown by the method invoked
+     * @throws IllegalAccessException if the requested method is not accessible via reflection
+     *  
+     * @since 3.5
+     */
+    public static Object invokeMethod(final Object object, final boolean forceAccess, final String methodName) 
+    		throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        return invokeMethod(object, forceAccess, methodName, ArrayUtils.EMPTY_OBJECT_ARRAY, null);
+    }
 
     /**
      * <p>Invokes a named method whose parameter type matches the object type.</p>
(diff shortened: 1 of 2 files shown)


sebbASF:
Note that an alternative solution for testing private methods is to make them package-protected.
(and document why they are not private!)

The test code needs to access it from the same package, but that is not usually a problem.
If necessary create a public method in the same package which can be used by all the test code.
A similar approach can be used for private fields; create a package-protected getter.

Derek-Ashmore:
Thanks for taking time to respond.  I should be challenged on the need 
for this.  I grant your point-- over-exposing methods is technically 
possible.  I've issues with that, however.

 From an design perspective, you should only expose fields (through 
accessors/mutators) and methods you intend to expose.  That is, that are 
being made available for use by other classes or by inheritance. 
Over-exposing methods and fields in the way you describe violates OO 
design principles.  Creating an easy way for test code to access 
privates is less of a principle breach; test classes presumably know 
exactly how a class is supposed to operate and are tightly coupled to 
that class anyway.

Why create FieldUtils?  That logic should also have been used to not 
implement the methods on FieldUtils that allow you to manipulate private 
fields. Doesn't that run into the same issue?

I don't like over-exposing fields and methods just to be able to cover 
them in testing.  I'm not alone.

Thanks for taking a look at this request and taking time to comment.  
Thanks, also, for the most interesting discussion.
Derek <https://www.linkedin.com/in/derekashmore> 
<https://twitter.com/Derek_Ashmore> <http://www.derekashmore.com/>
Please note that my email address has changed to 
derek.ashmore@dvtconsulting.com

PascalSchumacher:
I agree.

FieldUtils supports accessing private fields, so MethodUtils should support accessing private methods.

Derek-Ashmore:
I happened to notice that I inadvertently made unwanted formatting changes in a previous commit. I've take pains to reduce the number of changed lines dramatically. Hopefully, this will make it easier to merge this change into the trunk if it's decided that this is a reasonable enhancement.

I'm also willing to make additional changes to the if you would like; anything I can do to make this contribution easier to merge in. 

Please let me know if you have questions.

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -92,6 +92,29 @@ public static Object invokeMethod(final Object object, final String methodName)
             IllegalAccessException, InvocationTargetException {
         return invokeMethod(object, methodName, ArrayUtils.EMPTY_OBJECT_ARRAY, null);
     }
+    
+    /**
+     * <p>Invokes a named method without parameters.</p>
+     *
+     * <p>This is a convenient wrapper for
+     * {@link #invokeMethod(Object object,String methodName, Object[] args, Class[] parameterTypes)}.
```
PascalSchumacher: I guess this should be ``invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object[] args, Class[] parameterTypes)`?

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -126,13 +149,41 @@ public static Object invokeMethod(final Object object, final String methodName,
     /**
      * <p>Invokes a named method whose parameter type matches the object type.</p>
      *
-     * <p>This method delegates the method search to {@link #getMatchingAccessibleMethod(Class, String, Class[])}.</p>
+     * <p>This method supports calls to methods taking primitive parameters 
+     * via passing in wrapping classes. So, for example, a {@code Boolean} object
+     * would match a {@code boolean} primitive.</p>
+     *
+     * <p>This is a convenient wrapper for
+     * {@link #invokeMethod(Object object,String methodName, Object[] args, Class[] parameterTypes)}.
```
PascalSchumacher: I guess this should be ``invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object[] args, Class[] parameterTypes)`?

PascalSchumacher: I guess this should be ``invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object[] args, Class[] parameterTypes)`?

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
PascalSchumacher: I guess this should be ``invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object[] args, Class[] parameterTypes)`?

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -142,21 +193,57 @@ public static Object invokeMethod(final Object object, final String methodName,
...
-        final Method method = getMatchingAccessibleMethod(object.getClass(),
-                methodName, parameterTypes);
+        
+        final String messagePrefix;
+        final Method method;
+        if (forceAccess) {
+            messagePrefix = "No such method: ";
+            method = getMatchingMethod(object.getClass(), methodName, parameterTypes);
+            if (method != null) {
+                method.setAccessible(true);
```
PascalSchumacher: For `FieldUtils` there is this jira issue: https://issues.apache.org/jira/browse/LANG-965?jql=project%20%3D%20LANG%20AND%20status%20%3D%20Open%20ORDER%20BY%20priority%20DESC "FieldUtils methods leak accessible flags" with requests that the visibility of a field should be restored after the field is written. Maybe the visibility of the method should also be restored after it is invoked? What do think?

PascalSchumacher:
Thanks for updating the pull request. :+1: 

It would be nice if you could rebase on current master.

Thanks,
Pascal

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
Derek-Ashmore: Yeah -- my bad.   I'll correct the javadoc link.  Thanks for catching that.

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
Derek-Ashmore: Agreed -- I didn't realize that the issue was out there and wanted to be consistent with FieldUtils.   I'll make this change.   Thanks for making me aware of it.

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
Derek-Ashmore: Yeah -- my bad. I'll correct the javadoc link. Thanks for catching that.

Derek-Ashmore:
Agreed -- in addition to the other changes you mentioned, I'll re-base using the current master so that it's easier to merge in.   Thanks for looking at this.

PascalSchumacher:
Hi Derek,

you should be able to rebase this pull request by following these steps:
* `git remote add upstream https://github.com/apache/commons-lang.git`
* `git checkout master`
* `git pull --rebase upstream`
* `git checkout LANG-1195`
* `git rebase master`
* edit/resolve conflicts
* `git rebase --continue`
* `git push -f`

Now this pull request is rebased onto master.

Derek-Ashmore:
Hi Pascal,

I'm not sure I did the re-base part correctly.  Both my forked master and LANG-1195 have all changes we discussed and are based on the updated master.  All conflicts were resolved and "I think"; you've obviously done this more than I have and I trust your judgement.  Please let me know if you see problems with what I've done.  Thanks for looking at this.

PascalSchumacher:
Hi Derek,

the rebase is perfect. :+1: 

I think resetting the accessibility of should be done in a finally block, so that it works even when the method invocation throws an exception. What do you think?

Derek-Ashmore:
Good idea -- the change has been made and checked in.  Any other thoughts?

Derek-Ashmore:
It's the Objects.deepEquals() -- to new for the supported jdk versions -- I'll remove it

PascalSchumacher:
Hi Derek,

sorry for the delay.

I took another look at the code and compared it with `FieldUitls`. Maybe it is a good idea to change the method parameter order to `Object object, String methodName, boolean forceAccess`. I think `object` and `methodName` belong together and this order is similar to the order of the parameters of `FieldUtils` methods e.g. `FieldUtils#readDeclaredField(Object target, String fieldName, boolean forceAccess)`.

What do you think?

Derek-Ashmore:
I originally thought the same thing.  The problems is the "Object... args" support.  The fact that it's optional means that you can put an optional boolean at the end easily and tell the difference between its intended use to force access or as an argument to the method.  Fields don't have args, so there's no ... support. See example below. 

invokeMethod(final Object object, final String methodName, Object... args)
invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object... args)

We *could* put it at the end in the example below, but then it's inconsistent with the other overload:
invokeMethod(final Object object, final boolean forceAccess, final String methodName, Object[] args, Class<?>[] parameterTypes)

Am I missing something obvious?

PascalSchumacher:
You are right. I missed that. It is impossible to use the parameter order I suggested, because it would clash with the existing `invokeMethod` signature. Sorry.

PascalSchumacher:
Thanks! :+1:

Derek-Ashmore:
Thank you! Without your pull request suggestion and additional support, it never would have gone anywhere. Please let me know if I can ever return the favor.

PascalSchumacher:
@Derek-Ashmore Just today I learned that the resetting of the original accessibility of the method I asked you to add is not necessary, because `Method#setAccessible` only modifies the behavior of the `AccessibleObject` not of the actual method. 

See http://stackoverflow.com/questions/10638826/java-reflection-impact-of-setaccessibletrue and http://www.artima.com/weblogs/viewpost.jsp?thread=164042 for details.

Sorry that I caused you extra work. :(

Derek-Ashmore:
No worries – thanks for the information.

PR #143, opened 2016-05-16
LANG-1230 Remove unnecessary synchronization

marschall:
The synchronization in HashCodeBuilder and EqualsBuilder is not needed
and can safely be removed.

Issue: LANG-1230
https://issues.apache.org/jira/browse/LANG-1230


PascalSchumacher:
Thanks!

PR #122, opened 2016-01-24
Add method to DiffBuilder to allow appending from a DiffResult.

NickManley:

PR #151, opened 2016-05-21
LANG-1233: DiffBuilder add method to allow appending from a DiffResult

NickManley:
Suppose one has a Diffable object where one of the properties is also diffable. For example, given a Person object with an address property, it might be useful to be able to perform a diff on the Person and identify specific differences within the address. You might want the diff result of a Person to show that "address.city" is different rather than just the entire address. This feature allows for that.

Creating a new pull request after the original was closed by mistake.

JIRA: https://issues.apache.org/jira/browse/LANG-1233
Previous Pull Request: https://github.com/apache/commons-lang/pull/122

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
@@ -981,6 +981,9 @@ public class DiffBuilder implements Builder<DiffResult> {
      * @param diffResult
      *            the {@code DiffResult} to append
      * @return this
+     * @throws IllegalArgumentException
+     *             if field name is {@code null}
+     * @since 3.5
      */
     public DiffBuilder append(final String fieldName,
             final DiffResult diffResult) {


src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
```diff
@@ -942,6 +942,59 @@ public DiffBuilder append(final String fieldName, final Object[] lhs,
...
+     *       .build();
+     *   }
+     * }
+     * </pre>
+     * 
+     * @param fieldName
+     *            the field name
+     * @param diffResult
+     *            the {@code DiffResult} to append
+     * @return this
```
PascalSchumacher: Hi Nick,

please add `@since` and `@throws` tags.

Thanks!

NickManley:
I didn't realize the original pull request had already been merged, so things got out of sync with the pull request. I think everything is back to normal now.

PascalSchumacher:
Hi Nick,

sorry I did not mean to merge the pull request before giving you a change to update.

I will merge this one now.

Sorry again,
Pascal

PascalSchumacher:
Thanks! :+1:

pfried:
> Suppose one has a Diffable object where one of the properties is also diffable. For example, given a Person object with an address property, it might be useful to be able to perform a diff on the Person and identify specific differences within the address. You might want the diff result of a Person to show that "address.city" is different rather than just the entire address. This feature allows for that.

Actually imho this is not true since the Diffbuilder has a Type `<T>` the appendable `DiffResult` is also `DiffResult<T>`. If you have a `class Person` and a `class Address` this statement is therefore not true? Or am I missing something and you can give me an example on how to do this?

PR #152, opened 2016-05-21
LANG-1223: Add StopWatch#getTime(TimeUnit)

NickManley:


PascalSchumacher:
Hi Nick,

thanks for the pull request. :)

I think for completeness there should be at least one unit test.

NickManley:
The StopWatch times are private, so I can't create a mock StopWatch to test it comprehensively. I did however write a simple unit test to verify that sleeping for 600ms will output 0 seconds and that as soon as it passes 1000ms, it reports 1 second.

PascalSchumacher:
Thanks for updating the pull request.

Yes, the `StopWatch` implementation is not really testable. :(

I think the best approach would be to use reflection to set the times (e.g. by using `FieldUtils#writeField`).

NickManley:
I added a convenience method to the StopWatchTest for creating a suspended StopWatch with an arbitrary time using FieldUtils as you suggested.

PascalSchumacher:
Thanks! :+1:

PR #153, opened 2016-05-21
LANG-1236: Add StopWatch#getTime with String Format Argument

NickManley:
I wasn't sure if this kind of thing should be a part of the getTime method or if toString should be overloaded to take a format argument. Either way seems okay to me. I put it on getTime, but if there are disagreements about that, I would be okay with putting it in toString instead.


NickManley:
Closing the pull request. Ticket marked as Won't Fix.

PR #154, opened 2016-05-22
LANG-1225: Add RandomStringUtils#randomGraph and #randomPrint which m…

PascalSchumacher:
…atch corresponding regular expression class

These are useful over randomAscii because they do not contain the DEL character but otherwise contain the full range of ASCII printing characters, and optionally include whitespace. This is useful for testing user defined inputs where characters like spaces, angle brakets, semicolons, dashes, etc. can cause issues.

This is a part of https://github.com/apache/commons-lang/pull/101


PascalSchumacher:
Merged: https://github.com/apache/commons-lang/commit/1a002c67f2d3c198411a62165e53b31d6b414aa3

PR #155, opened 2016-05-23
LANG-1214 Handle "void" in ClassUtils.getClass()

henryptung:
`getClass()` previously deferred to an array type to get the class,
but this fails for `void`, which lacks an array type.
Instead, map explicitly from primitive name to primitive class.

Added unit test line fails before change, passes after change.


PascalSchumacher:
Thanks!

PR #156, opened 2016-05-24
LANG-1018: Fix precision loss on NumberUtils.createNumber(String)

NickManley:
**Warning: This patch will break backwards compatibility.**

This patch is designed to fix the loss of precision when using NumberUtils.createNumber(String). However, the fix is not 100% backwards compatible with the currently released commons-lang 3.4. To fix the loss of precision, values that were previously returning Floats are now returning Doubles. Some explicit casts may break as a result.

```java
// Works in 3.4, but causes ClassCastException in patch.
Float f = (Float) NumberUtils.createNumber("-160952.54");
System.out.println(f); // prints "-160952.55"

// Works with patch.
Double d = (Double) NumberUtils.createNumber("-160952.54");
System.out.println(d); // prints "-160952.54"
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -491,9 +491,7 @@ public class NumberUtils {
         // if both e and E are present, this is caught by the checks on expPos (which prevent IOOBE)
         // and the parsing which will detect if e or E appear in a number due to using the wrong offset
 
-        int numDecimals = 0; // Check required precision (LANG-693)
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
@NickManley Not sure if you checking jira, so I thought I let you know there is some feedback on your pull request there: https://issues.apache.org/jira/browse/LANG-1018?focusedCommentId=15316755&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15316755

PascalSchumacher:
Thanks! :+1:

PR #160, opened 2016-06-03
Eliminate redundant calculation of isDecimalPoint.

ewannema:
There was a redundant check of ```str.charAt(i) == '.'```.

I eliminated the second check in favor of using the already defined ```isDecimalPoint```.


PascalSchumacher:
Thanks!

PR #169, opened 2016-07-06
LANG-1248 FastDatePrinter Memory allocation regression

benbenw:
When the code was migrated from StringBuffer to Appendable in LANG-1152.
We've lost the ability to modify the buffer (setCharAt) 
The new implementation of appendFullDigits allocate a temporary char
array to work around that limitation.
This is a major source of memory allocation which is not present in
version 3.4.
Test case : 
Single fastdateformat instance
100000 format of long 
pattern : yyyy-MM-dd'T'HH:mm:ss.SSSZ
lang 3.4 : ~30 MB allocated
lang 3.5-snapshot : ~40 MB allocated
-> ~9.3 MB are coming from appendFullDigits called from
PaddedNumberField

This commit add a fast path for 1 to 4 digits which avoid the memory
allocation from the temporary work array.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/FastDatePrinter.java
+++ b/src/main/java/org/apache/commons/lang3/time/FastDatePrinter.java
@@ -684,24 +684,56 @@ public class FastDatePrinter implements DatePrinter, Serializable {
      * @param value the value to append digits from.
      */
     private static void appendFullDigits(final Appendable buffer, int value, int minFieldWidth) throws IOException {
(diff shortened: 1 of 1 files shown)


chonton:
Are you OK with using this [alternative implementation](https://github.com/chonton/commons-lang/commit/4d26fa6c107636c0f986c45379edcb18ac1ec3f5)?

benbenw:
Discussion was done on the jira
https://issues.apache.org/jira/browse/LANG-1248

PR #171, opened 2016-07-15
Removing test redundant for org.apache.commons.lang3.ClassUtilsTest.testWrapperToPrimitiveNoWrapper

TestSuiteReductionResearch:
We are researchers working on identifying redundant tests in a test suite. Our analysis of finding redundant tests involve each test's dynamic code coverage and their potential fault-detection capability.

Through our analysis, we found that the tests org.apache.commons.lang3.ClassUtilsTest.testWrapperToPrimitiveNoWrapper, org.apache.commons.lang3.ClassUtilsTest.testWrapperToPrimitiveNull are redundant with respect to one another. In this pull request, we are proposing to keep only the test org.apache.commons.lang3.ClassUtilsTest.testWrapperToPrimitiveNoWrapper while adding @Ignore annotations to the remaining test. However, as we believe these tests are identical, any one of them can be kept while skipping the remaining test.

If you do not believe any of these tests should be ignored, we would greatly appreciate it if you could follow up on this pull request and let us know your reasons.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
@@ -934,7 +934,7 @@ public class ClassUtilsTest  {
         assertNull("Wrong result for non wrapper class", ClassUtils.wrapperToPrimitive(String.class));
     }
 
-    @Test
+    @Test @org.junit.Ignore
     public void testWrapperToPrimitiveNull() {
         assertNull("Wrong result for null class", ClassUtils.wrapperToPrimitive(null));
     }


kinow:
Both tests use ClassUtils#wrapperToPrimitive. One passes a String.class, and the other a null value. Then the method looks at a map, created and filled in a static constructor. 

Since String.class is not in the map, it returns a null value.

Then trying to get a value for the key null, also returns a null value.

I agree that, considering code coverage, both tests are redundant. But they are assessing two different things, as defined in the method names. One assessing that asking for null will return a null. And the other that requesting an element that is not in the map also returns null.

Even though they may be redundant now, something may change in the future, and having tests covering scenarios, regardless of coverage, is a good thing too.

>If you do not believe any of these tests should be ignored, we would greatly appreciate it if you could follow up on this pull request and let us know your reasons.

So for this, I'm -1 for removing the test based only on coverage. They may be redundant, looking from a test coverage perspective, but as a developer using that method, I expect to be able to quickly understand what's going to happen in each scenario I use it. Tests and Javadoc with some examples, covering these scenarios, can be extremely useful.

garydgregory:
I agree, -1 to remove the test.

august782:
Thank you very much for the feedback! It does make sense that you want to keep tests for different scenarios even if they do have the same coverage.

Besides these two tests, our analysis actually did discover other tests that are redundant with one another in the test suite. Would you be open to us sending some more pull requests concerning removing redundant tests?

TestSuiteReductionResearch:
Thank you very much for the feedback! It does make sense that you want to keep tests for different scenarios even if they do have the same coverage.

Besides these two tests, our analysis actually did discover other tests that are redundant with one another in the test suite. Would you be open to us sending some more pull requests concerning removing redundant tests?

kinow:
@august782, thanks for being understanding. I think finding redundant tests, and simplifying it is a good idea, just not in the case of tests covering many different scenarios.

I think many tests in Apache Commons components are written with scenarios in mind, and method-oriented. As normally the components are libraries to be included in other systems, the tests aim at asserting that methods work as expected, under different conditions (i.e. with a diverse range of possible arguments).

Having good code coverage is important as well, but tests are not being written in Apache Commons components simply to have a high score in the reports - though normally covering well your scenarios, you end up with a good branch and line test coverage as well (citation needed :) ).

So if you have found redundant tests, where the same scenario is being covered in multiple tests, and you believe that reducing it will i) benefit users and developers in understanding the code, and ii) still cover the possible use cases for the library, or iii) make it easier to assert the same thing, then I think these pull requests would be useful.

For other pull requests that disable or remove tests based only on line and or branch coverage, I would still argue against removing them, and in favor to keep them; for the reasons I stated before.

TestSuiteReductionResearch:
Thank you again for providing more feedback. I believe our analysis did find some other tests that do seem to be covering the same scenarios, and we will be sending them as separate pull requests.

PR #174, opened 2016-07-20
[LANG-1250] SerializationUtils::deserialize has unnecessary code and a comment for that

Glease:
There is no real CCE possible, even the javadoc say so. Github's blame showed the code and javadoc is written by two indiviual person in two different years, so that may be the cause.


PascalSchumacher:
Thanks!

PR #180, opened 2016-07-27
LANG-1251: SerializationUtils.ClassLoaderAwareObjectInputStream should use static initializer to initialize primitiveTypes map.

ueshin:
`SerializationUtils.ClassLoaderAwareObjectInputStream` should use static initializer to initialize `primitiveTypes` map because initializing the map in the constructor of `ClassLoaderAwareObjectInputStream` would break thread safety. `java.util.HashMap` is not thread safe.


PascalSchumacher:
Thanks! :+1:

PR #184, opened 2016-08-25
Correct replacing 160 in normalizeSpace

phoenix384:
Normalizing spaces of a string containing a nbsp (160) followed by a regular whitespace results in a string containing a double space character instead. You have to execute normalize-space() twice to get a string with only single spaces.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -2593,9 +2593,6 @@ public class StringUtilsTest {
 
     @Test
     public void testNormalizeSpace() {
-        // Java says a non-breaking whitespace is not a whitespace.
-        assertFalse(Character.isWhitespace('\u00A0'));
-        //
         assertNull(StringUtils.normalizeSpace(null));
         assertEquals("", StringUtils.normalizeSpace(""));
         assertEquals("", StringUtils.normalizeSpace(" "));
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
@phoenix384 The problem is that starting with 3.4 StringUtils#normalizeSpace no longer normalizes unicode non-breaking spaces. I reported this as a bug and argued for reverting this behavior, but other people disagreed. :( 

For details see:
https://github.com/apache/commons-lang/pull/113
https://issues.apache.org/jira/browse/LANG-1184

phoenix384:
But it can't be the desired behaviour, that executing normalize-space twice normalizes non-breaking spaces while executing it once replaces them with a normal whitespace.
Either don't touch them or normalize them immediately.
I would prefer to normalize them, since they simply are white spaces.

PascalSchumacher:
Thanks for the information. I did not know that `normalizeSpace` replaces non-breaking spaces with normal whitespace.

Both these test succeed:

```java
assertEquals("a b", StringUtils.normalizeSpace("a\u00A0b"));
assertEquals("a  b", StringUtils.normalizeSpace("a\u00A0 b"));
```

The argument brought forward as to why `normalizeSpace` does not normalize non-breaking space anymore was that `Character#isWhitespace` returns `false`. Like I said I do not agree with this argument.

PascalSchumacher:
the non-breaking space replacement was added with https://issues.apache.org/jira/browse/LANG-910

PR #185, opened 2016-09-07
Fix ReflectionToStringBuilder toString javadoc

voyageth:
- as is
```
     * @param outputStatics
     *            whether to include transient fields
```

- to be
```
     * @param outputStatics
     *            whether to include static fields
```


britter:
Thank you!

PR #186, opened 2016-09-11
LANG-1252: rename isNumber, isCreatable. Accommodate for java 6 "+" handing.

chtompki:
Associated with: https://issues.apache.org/jira/browse/LANG-1252

@britter, do you have any thoughts on this. I think that it sufficiently covers the issue, but I wasn't sure about the mechanics about predicating functionality on java version. The main issue is that `new java.math.BigInteger("+2")` works for Java 1.7 and up, but not in Java 1.6. On the other hand `java.math.Float.valueOf("+2.0")` works on all versions of java. So I tried to accommodate for that in the `isCreatable` method.

Based on your email about releases and my being slightly deadline driven person, I figured I'd try to put something reasonable together for the 3.5 release candidate. That said, if the diff on the PR doesn't look solid enough, I'm willing to work on the changes some more.


src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1374,8 +1407,11 @@ public static boolean isNumber(final String str) {
         boolean hasDecPoint = false;
         boolean allowSigns = false;
         boolean foundDigit = false;
+        boolean isJava6 = StringUtils.startsWith(
+                System.getProperty("java.version"), "1.6");
```
britter: You can better use `SystemUtils`for this. But I'm about to extend `JavaVersion` in for [LANG-1263](https://issues.apache.org/jira/browse/LANG-1263) so that will make this even easier. I'll ping you when I have that finished.

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
```diff
@@ -1216,91 +1217,104 @@ public void testIsDigits() {
...
 
     /**
-     * Tests isNumber(String) and tests that createNumber(String) returns
-     * a valid number iff isNumber(String) returns false.
+     * Tests isCreatable(String) and tests that createNumber(String) returns
+     * a valid number iff isCreatable(String) returns false.
      */
     @Test
-    public void testIsNumber() {
-        compareIsNumberWithCreateNumber("12345", true);
```
britter: Don't remove this tests. Although we have deprecated `isNumber` we will probably have to maintain the code for a while before it can be dropped in Lang 4.0

chtompki:
I think the following commits satisfy what you were looking for there. Let me know your thoughts.

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1360,11 +1361,44 @@ public static boolean isDigits(final String str) {
...
+     *
+     * <p><code>null</code> and empty/blank {@code String} will return
+     * <code>false</code>.</p>
+     *
+     * <p>Note, {@link #createNumber(String)} should return a number for every
+     * input resuling in <code>true</code>.</p>
+     *
+     * @param str  the <code>String</code> to check
+     * @return <code>true</code> if the string is a correctly formatted number
+     * @since 3.3 the code supports hex {@code 0Xhhh} and octal {@code 0ddd} validation
```
britter: should be changed to simply `@since 3.5`

britter:
Please also add your change to `src/changes/changes.xml`. Thank you!

PR #187, opened 2016-09-11
LANG-1263 Add possibility to retrieve the current JavaVersion

britter:
Added a constant `CURRENT` to the `JavaVersion` enum which represents the JavaVersion for the `java.specification.version` system property. This allows for checks like this:

```java
if(JavaVersion.CURRENT.atLeast(JAVA_1_6) {
  // do something only possible in Java 1.6+
}
```

### DIFF
--- a/src/test/java/org/apache/commons/lang3/JavaVersionTest.java
+++ b/src/test/java/org/apache/commons/lang3/JavaVersionTest.java
@@ -21,7 +21,9 @@ package org.apache.commons.lang3;
 import org.junit.Test;
 
 import static org.apache.commons.lang3.JavaVersion.JAVA_RECENT;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.apache.commons.lang3.JavaVersion.JAVA_0_9;
 import static org.apache.commons.lang3.JavaVersion.JAVA_1_1;
 import static org.apache.commons.lang3.JavaVersion.JAVA_1_2;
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/lang3/JavaVersionTest.java
```diff
@@ -72,4 +73,28 @@ public void testToString() {
...
+            assertEquals(JAVA_1_4, JavaVersion.CURRENT);
+        } else if (SystemUtils.JAVA_SPECIFICATION_VERSION.equals("1.5")) {
+            assertEquals(JAVA_1_5, JavaVersion.CURRENT);
+        } else if (SystemUtils.JAVA_SPECIFICATION_VERSION.equals("1.6")) {
+            assertEquals(JAVA_1_6, JavaVersion.CURRENT);
+        } else if (SystemUtils.JAVA_SPECIFICATION_VERSION.equals("1.7")) {
+            assertEquals(JAVA_1_7, JavaVersion.CURRENT);
+        } else if (SystemUtils.JAVA_SPECIFICATION_VERSION.equals("1.8")) {
+            assertEquals(JAVA_1_8, JavaVersion.CURRENT);
+        } else if (SystemUtils.JAVA_SPECIFICATION_VERSION.equals("9.0.0")) {
```
PascalSchumacher: Not sure if this is correct. Shouldn't it be something like `9.` if I understand 

http://openjdk.java.net/jeps/223

correctly?

britter:
@PascalSchumacher I've changed the implementation to use the `IS_JAVA_XX` fields from `SystemUtils`. However, Java 9 detection is still broken, since it is broken in `SystemUtils`

britter:
We already have `SystemUtils.isJavaVersionAtLeast(JavaVersion)` so there is no reason to add this API.

PR #188, opened 2016-09-13
LANG-1266 Add alphabet converter

eyala:
(as described in [the mailing list](http://mail-archives.apache.org/mod_mbox/commons-dev/201609.mbox/%3c289983494.3057706.1472720010277@mail.yahoo.com%3e))

This is a utility class I wrote for converting from one alphabet to another - for example, from unicode to latin, without using some of the chars in latin. The usage looks like this:

```
Set<Character> originals; // a, b, c, d
Set<Character> encoding; // 0, 1, d
Set<Character> doNotEncode; // d

AlphabetConverter ac = AlphabetConverter.createConverter(originals, encoding, doNotEncode);

ac.encode("a"); // 00
ac.encode("b"); // 01
ac.encode("c"); // 0d
ac.encode("d"); // d
ac.encode("abcd"); // 00010dd
```
Of course, 

`x.equals(ac.decode(ac.encode(x)))`

 should always be true.

The implementation provided makes the encodings of fixed length, other than the "do not encode" chars, which remain as they are (length one).

In addition, in order to make it easier to preserve the encoding scheme, I've added a human-readable toString implementation, and a constructor that can recreate an AlphabetConverter from the encoding map, such that:
```

AlphabetConverter ac;

ac.equals(AlphabetConverter.createConverterFromMap(ac.getOriginalToEncoded())); // always should be true

```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/AlphabetConverter.java
+++ b/src/main/java/org/apache/commons/lang3/text/AlphabetConverter.java
@@ -1,0 +1,356 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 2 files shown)


britter:
The issue has been moved to https://issues.apache.org/jira/browse/TEXT-19 Please reference TEXT-19 in your PR against the Commons Text repository. Thank you!

PR #191, opened 2016-09-15
ToStringExcludeNullValue

gabriel-amaral:
I added an annotation to exclude a field from toString when the value is null.


britter:
@amaral7 do you really want to configure this on a per field basis or would it be enough to add a configuration option to `ToStringBuilder` to configure the general behavior for all fields?

amaral7:
@britter when I made the changes, I was thinking to cover the largest number of possible cases, and for this reason I implemented this configuration per field that would cover even who wants the general behavior for all fields, but more painful. I think there may be cases that need this behavior per field, but I know that most cases would be more comprehensive, for this reason have made the changes. And that wasn´t so painful to me ;).

PascalSchumacher:
Support for excluding `null` values was added in https://github.com/apache/commons-lang/commit/8147cc5b3de5fa7a3a3e8116355efa44367dc3c5

@gabriel-amaral Thanks for the pull request and sorry that it was not merged. Can you please close it? Thanks!

PR #193, opened 2016-09-23
Add isAnyNotEmpty() and isAnyNotBlank() to StringUtils

ptemplier:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -258,6 +258,35 @@ public static boolean isAnyEmpty(final CharSequence... css) {
...
+     * StringUtils.isAnyNotEmpty("foo", "bar")     = true
+     * </pre>
+     *
+     * @param css  the CharSequences to check, may be null or empty
+     * @return {@code true} if any of the CharSequences are empty or null
+     * @since 3.5
+     */
+    public static boolean isAnyNotEmpty(final CharSequence... css) {
+      if (ArrayUtils.isEmpty(css)) {
+        return true;
```
britter: Shouldn't this return false? An empty Array does not contain any non empty CharSequences. WDYT?

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -358,6 +387,36 @@ public static boolean isAnyBlank(final CharSequence... css) {
...
+     * StringUtils.isAnyNotBlank("foo", "bar")     = false
+     * </pre>
+     *
+     * @param css  the CharSequences to check, may be null or empty
+     * @return {@code true} if any of the CharSequences are not blank or null or whitespace only
+     * @since 3.5
+     */
+    public static boolean isAnyNotBlank(final CharSequence... css) {
+      if (ArrayUtils.isEmpty(css)) {
+        return true;
```
britter: See above

src/main/java/org/apache/commons/lang3/StringUtils.java
ptemplier: I took the same behaviour as isAnyBlank()/isAnyEmpty() but maybe those should return false too. It would make more sense.

PascalSchumacher:
I agree with both of you.

An empty array does not contain any non-empty/non-blank `CharSequence`s, so `isAnyNotEmpty`/`isAnyNotBlank` should return `false` for an empty array.

An empty array does not contain any empty/blank `CharSequence`s, so `isAnyEmpty`/`isAnyBlank` should return `false` for an empty array.

I created two jira issues: one for the addition: https://issues.apache.org/jira/browse/LANG-1270 and one the purposed change: https://issues.apache.org/jira/browse/LANG-1271

PascalSchumacher:
Thanks! :+1:

igaidai4uk:
This type of change is very painful, and not easy to track.

PR #194, opened 2016-09-29
add isAllBlank,isNotAllBlank method for String "null"   is also blank

wangdongxun:
add isAllBlank,isNotAllBlank method for String "null"   is also blank

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -305,7 +305,6 @@ public class StringUtils {
         }
         return true;
     }
-
     /**
      * <p>Checks if a CharSequence is not empty (""), not null and not whitespace only.</p>
      *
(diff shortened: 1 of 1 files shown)


wangdongxun:
if("null".equals(lowerCase("Null"))){
            return true;
}

britter:
@wangdongxun thanks for this PR, but I don't see this as a feature in `StringUtils`. Can you elaborate some more what use case you have for this?

PascalSchumacher:
Thanks for the pull request.

The method body can be shortened to `StringUtils.isBlank(cs) || StringUtils.equalsIgnoreCase(cs, "null")`.

The method would be very confusing, as the String `"null"` is not all bank.

I'm sorry, but I'm -1 to adding this.

kinow:
Thanks for your contribution @wangdongxun !

>The method would be very confusing, as the String "null" is not all bank.

Agree with @PascalSchumacher . Adding this could indeed be very confusing. Future requests could ask for "'null'" (with single quotes inside), or "NULL", etc. Easier done with two calls as @PascalSchumacher suggested IMO.

Unfortunately -1 as well.

wangdongxun:
Thank you for your reply @kinow @PascalSchumacher @britter 
    I feel very honore to receive your reply. I completely agree with @PascalSchumacher . You are right!

kinow:
Likewise @wangdongxun :-) are you albe to close this pull request? Otherwise I believe there is some integration in our infrastructure to let us close it.

Please do not hesitate to submit other pull requests, issues or comment in our mailing lists. 

Thank you
Bruno

PR #195, opened 2016-10-13
LANG-1160 StringUtils.abbreviate() to support 'custom ellipses' parameter

kinow:
From [LANG-1160](https://issues.apache.org/jira/browse/LANG-1160).

Before this pull request the code contained only two abbreviate methods. One abbreviate(String, int) that would call the other one abbreviate(String, int, int) with a default value for one of the int parameters.

This pull request adds two other methods with the same name, but that receive a custom abbreviation marker. The existing methods were update to call the new complete method with a default abbreviation marker of "...", which guarantees backward compatibility (we are keeping binary compatibility, and also feature backward compatibility).

Tests were update, test coverage is the same as before. No issues found building the site locally and from looking at CPD, PMD, FindBugs, CheckStyle and Surefire reports; all tests passing locally, with additional two tests in this PR.

Cheers
Bruno


PascalSchumacher:
+1

PR #196, opened 2016-10-17
site: Nav, javadoc now says GIT latest as opposed to SVN latest

chtompki:
@britter any thoughts on taking this without a jira?


britter:
@chtompki we don' need jira for these kind of changes, since they do not affect the behavior of Commons Lang in any way. Feel free to merge.

PR #197, opened 2016-10-19
ObjectUtils::xorNull added

cezarykluczynski:
Seems useful. What's your opinion?

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -194,6 +194,26 @@ public class ObjectUtils {
         return true;
     }
 
+    /**
+     * Exclusive or. Returns true only if exactly one of two values is {@code null}.
+     *
+     * <pre>
+     * ObjectUtils.xorNull(a, b)       = false
+     * ObjectUtils.xorNull(null, null) = false
+     * ObjectUtils.xorNull(a, null)    = true
+     * ObjectUtils.xorNull(null, b)    = true
+     * </pre>
+     *
+     * @param object1  the first object, may be {@code null}
+     * @param object2  the second object, may be {@code null}
+     * @return {@code false} if both values are {@code null}, or if no values is null,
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Thanks for the pull request!

I'm not sure this is generally useful. I do not remember ever needing/using a method like this.

dmjones500:
@cezarykluczynski Can you suggest a few use cases for this functionality in a normal application?

cezarykluczynski:
@dmjones500 I'm not sure what you mean by normal application, but here are two examples: https://github.com/cezarykluczynski/stapi/search?utf8=%E2%9C%93&q=xorNull

dmjones500:
@cezarykluczynski Commons Lang solves frequently encountered programming problems by adding methods that are arguably "missing" from `java.lang`. I'm trying to imagine the sort of problem this solves.

You've provided two examples. Looking at them, I'm not sure whether `xorNull` is the best way to solve them.

Based on the name of the method, I would have thought this might work:

```java
private void removeBirthNameIfItEqualsName(ActorTemplate actorTemplate) {
    String name = actorTemplate.getName();
    String birthName = actorTemplate.getBirthName();

    if (birthName == null) {
        return;
    }

    if (birthName.equals(name)) {
        actorTemplate.setBirthName(null);
    }
}
```

In your other example, you could rearrange the conditional statements and remove the need for your method:

```java
private LocalDate getIfEqualSkipNulls(LocalDate left, LocalDate right, Page item, boolean dateOfBirth) {

    if (left == null && right == null) {
        return null;
    } else if (left != null && right != null) {
        if (left.equals(right)) {
            return left;
        } else {
            String dateType = dateOfBirth ? "birth" : "death";
            log.error(
                    "Date of {} {} from for PageToLifeRangeProcessor does not equal date of {} {} from "
                            + "ActorTemplateToLifeRangeProcessor for {} - setting date to null",
                    dateType, left, dateType, right, item.getTitle());
            return null;
        }
    } else {
        return ObjectUtils.firstNonNull(left, right);
    }
}
```

PR #199, opened 2016-10-22
LANG-1258: Add ArrayUtils#toStringArray(Object[]) method

PascalSchumacher:
patch supplied by IG

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -7967,4 +7967,27 @@ public class ArrayUtils {
 
         return removeAll(array, Arrays.copyOf(indices, count));
     }
+
+    /**
+     * Returns an array containing the string representation of each entry in the argument array.
+     *
+     * <p>
+     * A {@code null} input Object[] returns {@code null}.
+     * </p>
+     *
+     * @param source the Object[] to be processed, may be null
+     * @return String[] of the same size as the source with its element's string representation,
+     * {@code null} if null Object[] input
+     * @since 3.6
+     */
+    public static String[] toStringArray(final Object[] source) {
+        if (null == source) return null;
+
+        String[] al = new String[source.length];
+        for (int i = 0; i < source.length; i++) {
+            al[i] = source[i].toString();
+        }
+
+        return al;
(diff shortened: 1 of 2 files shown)


kinow:
Patch looks good. I wonder if the inline if will raise a warning in checkstyle. Other than that, +1 :D

Xaerxess:
[I added null elements handling in PR to this branch](https://github.com/PascalSchumacher/commons-lang/pull/1), because in current state NPE would be thrown in case of null in array.

PascalSchumacher:
I have updated the pull request with @Xaerxess changes.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -7967,4 +7967,29 @@ public static boolean isSorted(boolean[] array) {
 
         return removeAll(array, Arrays.copyOf(indices, count));
     }
+
+    /**
+     * Returns an array containing the string representation of each entry in the argument array.
+     *
+     * <p>
+     * A {@code null} input Object[] returns {@code null}.
```
Xaerxess: One more thing came to my mind - we can document `null` elements handling. I don't have exact words for that, but something like `A {@code null} input Object[] returns {@code null}. {@code null} elements are preserved.`.

Xaerxess:
Sure, consistency in API is the key. But `stringIfNull` is used when array itself is `null`, not element, so it wouldn't be consistent with the API.

PascalSchumacher:
@Xaerxess Sorry for the confusion, I wasn't talking about `toString`, but about the `toPrimitive` methods e.g.

https://github.com/apache/commons-lang/blob/96c8ea2fb3719e2f6e3d7a4d7b46718f26515a86/src/main/java/org/apache/commons/lang3/ArrayUtils.java#L4497

As these also convert the type of the array I think they are the most similar existing methods (compared to the new method in this pull request).

PascalSchumacher:
Merged: https://github.com/apache/commons-lang/commit/8d95ae41975a2307501aa0f4a7eba296c59edce9 and https://github.com/apache/commons-lang/commit/8d601ab71228f7c3dff950540e7ee6e4043e9053

Thanks everybody!

PR #200, opened 2016-10-23
LANG-1276

andyklimczak:
Fix issue of buf using nonupdated buffer in StrBuilder replaceImpl
Avoid array OoB error by keeping variable buf consistent with buffer


PascalSchumacher:
Thanks! :+1:

PR #202, opened 2016-10-23
LANG-1034: Recursive and reflective EqualsBuilder

PascalSchumacher:
patch by yathos UG


PascalSchumacher:
I haven't locked at this in-depth yet, but seems like a useful enhancement.

I would remove the support for including transient fields, because that seems like a bad practice.

douddle:
I start to use your feature and I discover a bug.
String are now compared with internal data (hash and char[]). But sometimes hash value is not initialized, this is probably a cache for the hashcode. So the comparison fails even if the char[] content is the same.

PascalSchumacher:
@douddle I created https://issues.apache.org/jira/browse/LANG-1356 to track the bypass issue.

PR #205, opened 2016-11-03
LANG-1281: Javadoc of StringUtils.ordinalIndexOf is contradictory

PascalSchumacher:


kinow:
+1

aioobe:
But the implementation of the method does NOT increment by 1 if given the empty string. So this change causes the documentation to contradict the implementation.

PascalSchumacher:
`...incrementing the starting index by one after each successful match, so matches may overlap.` For empty String there is no match, so no incrementing is done.

Please do not hesitate to suggest better solution for this issue. Thanks!

PascalSchumacher:
I guess we should hide the implementation details and just state `Note: Matches may overlap.`.

What do you think?

aioobe:
Of course there are matches for the empty string. a) there is both a leading and trailing empty string in "a" for instance. b) the implementation doesn't return INDEX_NOT_FOUND when given the empty string.

As far as I can tell, the implementation increments the position with 1 each time a match is found, except if the sought string is the empty string, in which case it increments with 0 (conceptually that is; in practice it obviously returns 0 right away for that case, since if it increments by 0 each match, it will never move away from 0).

Personally I think the implementation is messy and hard to formalize. It would have been much better if the implementation consistently stuck the idea of incrementing by 1 for each match. This would not rule out the optimization because one would simply return n for the empty search string case.

Given the situation I guess it's not an option to change the implementation though, so the only option left is to adjust the documentation to reflect the current behavior. Here are two suggestions:

"increments by 1 unless the search string is the empty string in which case the position is never incremented"

or something like

"always returns 0 if the search string is the empty string, otherwise it increments the position by 1 for each match found"

kinow:
Oh, now it makes sense @aioobe 

I like 

"increments by 1 unless the search string is the empty string in which case the position is never incremented"

PascalSchumacher:
Updated the pull request using "increments by 1 unless the search string is the empty string in which case the position is never incremented".

aioobe:
Looks good to me, even though I think it's a terrible idea to view empty strings as some form of special strings and treat them differently than other strings.

PascalSchumacher:
Thanks everybody!

PR #208, opened 2016-11-04
LANG-1066: Add shell/XSI escape/unescape support

PascalSchumacher:
patch supplied by Mark

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
@@ -272,6 +272,40 @@ public class StringEscapeUtils {
         }
     }
 
+    /**
+     * Translator object for escaping Shell command language.
+     *
+     * @see <a href="http://pubs.opengroup.org/onlinepubs/7908799/xcu/chap2.html">Shell Command Language</a>
+     */
+    public static final CharSequenceTranslator ESCAPE_XSI =
+          new LookupTranslator(
+            new String[][] {
+              {"|", "\\|"},
+              {"&", "\\&"},
+              {";", "\\;"},
+              {"<", "\\<"},
+              {">", "\\>"},
+              {"(", "\\("},
+              {")", "\\)"},
+              {"$", "\\$"},
+              {"`", "\\`"},
+              {"\\", "\\\\"},
+              {"\"", "\\\""},
+              {"'", "\\'"},
+              {" ", "\\ "},
+              {"\t", "\\\t"},
+              {"\r\n", ""},
+              {"\n", ""},
+              {"*", "\\*"},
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -272,6 +272,40 @@ public int translate(final CharSequence input, final int index, final Writer out
...
+              {"\\", "\\\\"},
+              {"\"", "\\\""},
+              {"'", "\\'"},
+              {" ", "\\ "},
+              {"\t", "\\\t"},
+              {"\r\n", ""},
+              {"\n", ""},
+              {"*", "\\*"},
+              {"?", "\\?"},
+              {"[", "\\["},
```
britter: What about the closing braked?

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -413,6 +447,47 @@ public int translate(final CharSequence input, final int index, final Writer out
...
     }
 
+    public static final CharSequenceTranslator UNESCAPE_XSI = new BackslashUnescaper();
+
+    /**
+     * Translator object for unescaping backslash escaped entries.
+     *
+     * @since 3.6
+     */
+    static class BackslashUnescaper extends CharSequenceTranslator {
```
britter: Why not XsiUnescaper?

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -801,4 +876,61 @@ public static final String unescapeCsv(final String input) {
...
+    }
+
+    /**
+     * <p>Alias for {@link #escapeXSI(String)}.</p>
+     *
+     * @param input String to escape values in, may be null
+     * @return String with escaped values, {@code null} if null string input
+     * @since 3.6
+     */
+    public static final String escapeShell(final String input) {
```
britter: I'm not sure whether it is a good idea to have aliases for methods. Does this help users to find what they are looking for or does this really cause confusion?

PascalSchumacher: Not sure. I'm fine with removing the methods.

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
PascalSchumacher: No sure why the author of the patch choose this naming. I'm fine with changing it.

PascalSchumacher:
As discussed on the mailing list string escaping will move to commons-text, so I'm closing this pull request.

PR #211, opened 2016-11-15
 Fixed possible infinite loop in random of RandomStringUtils

allquantor:
Hi,
I faced that the `random` method can cause an infinite loop. It can be caused if the **end** parameter is too low for generating either numbers (< 48) or letters (<65). I added some exception tests to make the cases clear. Basically, if we call the random function with for example `numbers = true && end < 48
`
the validation for  `Character.isDigit(ch)` will fail and we increment count  `count++`.

Moreover:

- Removed the white spaces in RandomStringUtilsTest.java 
- Added the terms (inclusive) (exclusive) to the Javadoc of the random method to make it more clear.


PascalSchumacher:
Thanks for the pull request. :)

Please do not mix formatting/white-space changes with functional changes in commit.

allquantor:
@PascalSchumacher got you. PR is updated.

PascalSchumacher:
related issue: https://issues.apache.org/jira/browse/LANG-806 

I guess this pull request fixes the most important part of the issue.

PascalSchumacher:
Thanks! :+1:

PR #213, opened 2016-11-18
LANG-1288: Add a builder to StringEscapeUtils

PascalSchumacher:
patch supplied by Mark (for LANG-1066)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
@@ -429,6 +429,65 @@ public class StringEscapeUtils {
       super();
     }
 
+    /**
+     * <p>Convenience wrapper for {@link java.lang.StringBuilder} providing escape methods.</p>
+     *
+     * <p>Example:</p>
+     * <pre>
+     * new Builder(ESCAPE_HTML4)
+     *      .append("&lt;p>")
+     *      .escape("This is paragraph 1 and special chars like &amp; get escaped.")
+     *      .append("&lt;/p>&lt;p>")
+     *      .escape("This is paragraph 2 &amp; more...")
+     *      .append("&lt;/p>")
+     *      .toString()
+     * </pre>
+     * 
+     * @since 3.6
+     */
+    public static class Builder {
+
+        private final StringBuilder sb;
+        private final CharSequenceTranslator translator;
+
+        public Builder(final CharSequenceTranslator translator) {
+            this.sb = new StringBuilder();
+            this.translator = translator;
+        }
+
+        /**
+         * <p>Escape {@code input} according to the given {@link CharSequenceTranslator}.</p>
+         *
+         * @param input the String to escape
+         * @return {@code this}, to enable chaining
+         */
+        public Builder escape(final String input) {
+            sb.append(translator.translate(input));
+            return this;
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -429,6 +429,65 @@ public StringEscapeUtils() {
...
+     *      .escape("This is paragraph 2 &amp; more...")
+     *      .append("&lt;/p>")
+     *      .toString()
+     * </pre>
+     * 
+     * @since 3.6
+     */
+    public static class Builder {
+
+        private final StringBuilder sb;
```
britter: How about making it possible to pass in any `java.lang.Appendable` and use a `StringBuilder` as default value?

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -429,6 +429,65 @@ public StringEscapeUtils() {
...
+        private final CharSequenceTranslator translator;
+
+        public Builder(final CharSequenceTranslator translator) {
```
britter: I'd rather have a private constructor and then put a static factory method into `StringEscapeUtils` instead of having the inner class that one needs to construct like `new StringEscapeUtils.Builder(ESCAPE)`

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -429,6 +429,65 @@ public StringEscapeUtils() {
...
+     *      .escape("This is paragraph 1 and special chars like &amp; get escaped.")
+     *      .append("&lt;/p>&lt;p>")
+     *      .escape("This is paragraph 2 &amp; more...")
+     *      .append("&lt;/p>")
+     *      .toString()
+     * </pre>
+     * 
+     * @since 3.6
+     */
+    public static class Builder {
```
britter: Is `Builder` really a good name? It's more of a wrapper currently.

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
britter: Actually, I'm now thinking that it may be good to have a real decorator for `Appendables` with decorates any `Appendable` by piping input to the append methods through the translator first. WDYT?

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
PascalSchumacher: In general I would have implemented it the same way, but this style is consistent with `TypeUtils.WildcardTypeBuilder` and `BasicThreadFactory.Builder`.

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
PascalSchumacher: The problem with `Appendable` is that its `append` method can throw an `IOException`. That would make the builder more cumbersome to use. Of course it could just rethrow the `IOException` as a `RuntimeException`, but I do not know if this is o.k. for commons-lang.

PascalSchumacher:
As discussed on the mailing list string escaping will move to commons-text, so I'm closing this pull request.

PR #215, opened 2016-11-22
Separate test cases by a new line.

Abrasha:
As you can see on screenshots it was quiet hard to maintain each test case separately. But now it is more readable.

before
after
P.S. rename local variable name to more applicable.


PascalSchumacher:
Thanks! 👍

PR #217, opened 2016-11-24
LANG-1290: Increase test coverage of org.apache.commons.lang3.ArrayUtils class

Abrasha:
- fix an issue with not getting Annotation in AnnotaionUtilsTest
- add tests to cover ArrayUtils

Increased coverage to ~98% lines of code and 100% methods.

PR #218, opened 2016-11-24
Fix/different annotations test

Abrasha:
```
@Test
public void testAnnotationsOfDifferingTypes() {
    assertFalse(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), field4.getAnnotation(NestAnnotation.class)));
    assertFalse(AnnotationUtils.equals(field4.getAnnotation(NestAnnotation.class), field1.getAnnotation(TestAnnotation.class)));
}
```

This test is successful, but checks a different behaviour than expected.

`field4.getAnnotation(NestAnnotation.class)` is expected to return an annotation of type `NestAnnotation`, but actually it returns `null` because the Retention of annotation is set to default.

As a result, test did not cover quite important case.

See discussion: [#217](https://github.com/apache/commons-lang/pull/217)

### DIFF
--- a/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
@@ -350,7 +350,8 @@ public class AnnotationUtilsTest {
         NestAnnotation nest();
         NestAnnotation[] nests();
     }
-
+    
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Thanks! 👍

PR #221, opened 2016-12-04
Add IsAllEmpty and isAllBlank methods

martin-tarjanyi:
* Fix javadoc

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -433,7 +433,7 @@ public class StringUtils {
      * StringUtils.isAnyNotBlank("bob", "")        = true
      * StringUtils.isAnyNotBlank("  bob  ", null)  = true
      * StringUtils.isAnyNotBlank(" ", "bar")       = true
-     * StringUtils.isAnyNotBlank("foo", "bar")     = false
+     * StringUtils.isAnyNotBlank("foo", "bar")     = true
      * StringUtils.isAnyNotBlank(new String[] {})  = false
      * </pre>
      *
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -478,6 +478,31 @@ public class StringUtils {
       return !isAnyBlank(css);
     }
 
+    /**
+     * <p>Checks if all of the CharSequences are empty (""), null or whitespace only.</p>
+     *
+     * </p>Whitespace is defined by {@link Character#isWhitespace(char)}.</p>
+     *
+     * <pre>
+     * StringUtils.isAllBlank(null)             = true
+     * StringUtils.isAllBlank(null, "foo")      = false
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Thanks for the pull request. :)

I guess we better class theses methods `areAllEmpty`/`areAllBlank`. What do you think?

martin-tarjanyi:
The names you suggest express the use cases better, I agree.  

The only thing I'm afraid of is that people might have a difficult time finding them since most of the blank/empty methods start with 'is'. 

I'll change it, though. 

Other topic: what do you think of the handling of empty array or null as a single parameter? Is it right to return true in these cases?

Can I use git rebase to add changes or brand new commits are the preferred way?

dmjones500:
I vote strongly for `is*` method naming. That's the convention for Java methods of this type, even if it slightly upsets the English ear.

PascalSchumacher:
>I vote strongly for is* method naming. That's the convention for Java methods of this type, even if it slightly upsets the English ear.

Then let's keep is* names.

>Other topic: what do you think of the handling of empty array or null as a single parameter? Is it right to return true in these cases?

I'm not sure what the behavior should be in these cases. @dmjones500 / anybody else: any opinions on this?

PascalSchumacher:
I have created an jira issue for this: https://issues.apache.org/jira/browse/LANG-1293 and also asked for opinions on the empty/null array question there.

dmjones500:
I commented on the naming conventions without actually reviewing the change...

I don't think we should add methods that are simply the inverse of another method. It bloats the code base for no real gain.

I wouldn't be in favour of committing these new methods. The Javadoc fix is welcome, of course!

martin-tarjanyi:
So if this method is not added to the codebase but I would like to have this functionality in my client code, then I should call ``!isAnyNotEmpty(...)``. It would be very difficult to read. That's why I thought these methods would be useful. 

However, if I should choose between ``isAllEmpty`` and ``isAnyNotEmpty`` to be in StringUtils, I would support ``isAllEmpty`` because the negated version ``!isAllEmpty`` is more readable in a client code.

What do you think?

PascalSchumacher:
If only one method is added it should be `isAll*` instead of `isAnyNot*`. Since the `isAnyNot*` are not released yet, they can be removed.

I personally would have no problem with adding both types of methods. Making client code more readable is imho enough to justify the addition.

PascalSchumacher:
On further consideration, there are already methods with are "simply the inverse of another method": `isEmpty/isNotEmpty`, `isAnyEmpty/isNoneEmpty`, `containsAny/containsNone`. I will start a discussion on the mailing list.

martin-tarjanyi:
Hi @PascalSchumacher,
Is there any update on this matter? Should I change something?

PascalSchumacher:
@martin-tarjanyi Sorry for the delay. :( I finally send a mail to the mailing list:

https://mail-archives.apache.org/mod_mbox/commons-dev/201702.mbox/%3C13256228-1650-e98e-fe93-22801e530c88%40gmx.net%3E

PR #222, opened 2016-12-25
LANG-1303 : Escape HTML characters once

SampannaKahu:
Hi, this is my first time contributing to any open source project.
The feature goes as follows:

If already escaped HTML characters are in the input test, they get escaped again using StringEscapeUtils.escapeHtml4().

For example:
If the input is:
100 kg \&lt; 1000kg
Then the output of escapeHtml4() becomes:
100kg \&amp;lt; 1000kg

At my workplace, we felt the need for a method in StringEscapeUtils which does not escape already escaped characters.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
@@ -16,12 +16,10 @@
  */
 package org.apache.commons.lang3;
 
-import java.io.IOException;
-import java.io.Writer;
-
 import org.apache.commons.lang3.text.translate.AggregateTranslator;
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
Thanks for the pull request. 👍 

StringEscapeUtils was ported to commons-text and will soon deprecated in commons-lang.

It would be nice if you could open a pull request at https://github.com/apache/commons-text

Thanks!

SampannaKahu:
Thanks Pascal.
I have raised a pull request at commons-text: https://github.com/apache/commons-text/pull/23

PascalSchumacher:
Thanks Sampanna!

It would be nice if you would close this pull request. (I can not do it easily as this is a read-only repository.)

Thanks!

PR #223, opened 2016-12-30
[LANG-1304] StringUtils method to check for mixed case in string

andyklimczak:
[jira link](https://issues.apache.org/jira/browse/LANG-1304)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -7303,6 +7303,46 @@ public class StringUtils {
         return true;
     }
 
+    /**
+     * <p>Checks if the CharSequence contains mixed casing of both uppercase and lowercase characters.</p>
+     *
+     * <p>{@code null} will return {@code false}.
+     * An empty String (length()=0) will return {@code false}.</p>
+     *
+     * <pre>
+     * StringUtils.isMixedCase(null)    = false
+     * StringUtils.isMixedCase("")      = false
+     * StringUtils.isMixedCase("ABC")   = false
+     * StringUtils.isMixedCase("abc")   = false
+     * StringUtils.isMixedCase("aBc")   = true
+     * StringUtils.isMixedCase("A c")   = true
+     * StringUtils.isMixedCase("A1c")   = true
+     * StringUtils.isMixedCase("a/C")   = true
+     * StringUtils.isMixedCase("aC\t")  = true
+     * </pre>
+     *
+     * @param cs the CharSequence to check, may be null
+     * @return {@code true} if contains both uppercase and lowercase characters, and is non-null
+     */
+    public static boolean isMixedCase(final CharSequence cs) {
+        if (cs == null || isEmpty(cs) || cs.length() <= 1) {
+            return false;
+        }
+        boolean containsUppercase = false;
+        boolean containsLowercase = false;
+        final int sz = cs.length();
+        for (int i = 0; i < sz; i++) {
+            if (containsUppercase && containsLowercase) {
+                return true;
+            } else if (Character.isUpperCase(cs.charAt(i))) {
+                containsUppercase = true;
+            } else if (Character.isLowerCase(cs.charAt(i))) {
+                containsLowercase = true;
+            }
+        }
+        return containsUppercase && containsLowercase;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7193,6 +7193,45 @@ public static boolean isAllUpperCase(final CharSequence cs) {
...
+     * @return {@code true} if contains both uppercase and lowercase characters, and is non-null
+     */
+    public static boolean isMixedCase(final CharSequence cs) {
+        if (cs == null || isEmpty(cs)) {
+            return false;
+        }
+        boolean containsUppercase = false;
+        boolean containsLowercase = false;
+        final int sz = cs.length();
+        for (int i = 0; i < sz; i++) {
```
bindul: Wouldn't it be more efficient to track the case of the first character then break and return false on the first occurrence of the other case? Saves iterating the rest of the character sequence if it has mixed case.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -2563,6 +2563,24 @@ public void testIsAllUpperCase() {
...
+        assertFalse(StringUtils.isMixedCase(StringUtils.EMPTY));
+        assertFalse(StringUtils.isMixedCase(" "));
+        assertTrue(StringUtils.isMixedCase("aBc"));
+        assertFalse(StringUtils.isMixedCase("abc"));
+        assertFalse(StringUtils.isMixedCase("ABC"));
+        assertFalse(StringUtils.isMixedCase("aBc "));
+        assertFalse(StringUtils.isMixedCase("A c"));
+        assertFalse(StringUtils.isMixedCase("aBc\n"));
+        assertFalse(StringUtils.isMixedCase("A1c"));
+        assertFalse(StringUtils.isMixedCase("a/C"));
```
bindul: Can you please add a test for a single character string? I think the implementation breaks when there is only a single character in the String/CharSequence

src/main/java/org/apache/commons/lang3/StringUtils.java
andyklimczak: Almost, because consider the string "aB\n":
It does contain a lowercase, then an uppercase, but it also contains the character '\n', which is a neither lower or uppercase. Breaking early would miss this case.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
andyklimczak: Done, and I added a quick return case to skip going into the loop unnecessarily because a char sequence of length 1 cannot have both an upper and lowercase.

src/main/java/org/apache/commons/lang3/StringUtils.java
bindul: In which case the name of the method should probably be changed? With a method name like `isMixedCase`, I would expect a return value of `true`.
With your test string, yes there is the `\n` which is neither upper case or lower case, but the string has `a` and `B`, so it is mixed case.

andyklimczak: This method seems like a hybrid/sister method to ```isAllUppercase()``` and ```isAllLowercase()``` and treats non-alpha characters similarly.

PascalSchumacher: @arbasha I believe you requested this feature, what is your opinion on this?

arbasha: Hi, @PascalSchumacher I second @bindul's point, this method should return true if the CharSequence has at least one upperCase and one lowerCase character. As a requester of this feature I would like to see this method behave that way.

andyklimczak: I missed this message before, will correct the logic sometime this week

andyklimczak:
hello friends, I've updated `isMixedCase()` to be much more accepting of special characters, numbers, and spaces. Please double check that my tests assert the strings correctly, and match the desired logic 

Let me know if anything seems wrong ! 🐤 🔥

PascalSchumacher:
Thanks! :+1:

PR #226, opened 2017-01-04
[LANG-1308] Move CvsTranslators out of StringEscapeUtils and make them DRY

jarst:
CvsEscaper/CvsUnescaper are currently inner classes of StringEscapeUtils and it does not seem like it is a proper place for them.
Since they are package-private, they can be safely moved lang.text.translate package.
Moreover it's possible make them more DRY by introducing SinlgePassTranslator as suggested in TODO.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
@@ -16,11 +16,9 @@
  */
 package org.apache.commons.lang3;
 
-import java.io.IOException;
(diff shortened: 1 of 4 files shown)


PascalSchumacher:
Thanks for the pull request. 👍 

StringEscapeUtils has been moved to commons-text and the commons-lang version will soon be deprecated.

It would be nice if you would create a pull request for https://github.com/apache/commons-text instead.

Thanks in advance!

jarst:
I will do so :)

PR #233, opened 2017-02-04
Fix a few broken @literal tags

sebkur:
The Javadoc of SystemDefaultsSwitch contains some broken `@literal` tags.
Instead of `{@literal @}` to insert an '@' into the HTML, `{@literal@}`
without the space is used, which does not work.


PascalSchumacher:
Thanks!

PR #235, opened 2017-02-12
Remove redundant casts

mureinik:
Remove redundant casts throughout the codebase to make it cleaner and
easier to read.


PascalSchumacher:
Thanks! 👍 

I restored the casts in `ArrayUtils#testGetLength` before merging, because otherwise these tests make no sense. (The test probably make no sense anyway, as you can cast to `Date` and the tests still pass.)

PR #237, opened 2017-02-18
(doc) Use the full alphabet for the well-known example text.

gesellix:
This change adds the missing `s` character to complete the `[a-z]` namespace.

Since the commit only affects Javadoc and test code, I refrained from creating a Jira issue. If I should create one, please leave a comment.

I have signed the ICLA via email.

Signed-off-by: Tobias Gesellchen <tobias@gesellix.de>


chtompki:
Wait, I should have read that more completely. I agree with this change.

chtompki:
I should say. Many thanks for the contribution. Keep up the good work. Cheers, -Rob.

gesellix:
You're certainly right that the sentence makes more sense in a context related to font types. Yet, regardless of the context, since that sentence _is_ a standard, we should at least copy it without typos ;-)
Thanks for the quick review + merge!

chtompki:
Yes of course. My first read had it moving from "s" to "ed." Clearly a bad read on my part. Doh! :-)

PR #238, opened 2017-02-19
Validate's String.format without arguments

mureinik:
While calling `String.format("some string")` without any additional
arguments is not technically wrong, it's redundant, as it just
returns the same string.

Removing these calls and just using the string instead both cleans up
the code and offers a (very slight) performance gain.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Validate.java
+++ b/src/main/java/org/apache/commons/lang3/Validate.java
@@ -1052,7 +1052,7 @@ public class Validate {
     public static void inclusiveBetween(final long start, final long end, final long value, final String message) {
         // TODO when breaking BC, consider returning value
         if (value < start || value > end) {
-            throw new IllegalArgumentException(String.format(message));
+            throw new IllegalArgumentException(message);
         }
     }
 
@@ -1096,7 +1096,7 @@ public class Validate {
     public static void inclusiveBetween(final double start, final double end, final double value, final String message) {
         // TODO when breaking BC, consider returning value
         if (value < start || value > end) {
-            throw new IllegalArgumentException(String.format(message));
+            throw new IllegalArgumentException(message);
         }
     }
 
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Thanks for the pull request (and you other pull requests).

This pull request actually slightly changes the behavior of the methods. There is at least one use for `String.format` without parameters: inserting platform specific line separators with `%n`.

As the javadoc does not say that `String.format` is used on the message and I believe is is unlikely that somebody want would want platform specific line separators here I think we can still merge the pull request.

But maybe somebody else disagrees?

kinow:
Didn't know about this use of `String.format` @PascalSchumacher , thanks :-)

I'm +1 for merging the pull request. Agree it's unlikely, and not sure if its use was intentional.

mureinik:
Actually, it's worse  - If your string contains any other formatter other than `%n` these methods would throw a `MissingFormatArgumentException`, as you have no way to pass the arguments to the formatted string.

PascalSchumacher:
Yes, currently the user has to escape single `%`s in the error message with `%%`.

mureinik:
Now that I think about it, a cleaner solution might be to add `Object...` arguments to these methods' signatures, and pass it on to the `String.format` calls.
This way, we keep the code 100% backwards compatible, regardless of any `%n`s anyone might have used - any existing calls just pass empty varargs, which is perfectly legal.

The downside is slightly less elegant signatures.  If the maintainers prefer this approach, I'd be happy to push such a patch.

kinow:
@mureinik I think removing these calls should be fine. Specially given that other classes in lang use a normal String, and not `String.format` when there are no parameters, e.g. https://github.com/apache/commons-lang/blob/954ade4c1ae2adc0aaac3a1dbe800495c519520c/src/main/java/org/apache/commons/lang3/ThreadUtils.java#L54

PascalSchumacher:
Yes, lets keep the pull requests as is is. Sorry if I caused the impression of being overtly critical of this change.

PR #242, opened 2017-02-22
Fix FastDateParser#getStrategy(char, int, Calendar) javadoc

mureinik:
The javadoc refers to a `formatField` parameter, which the method doesn't
have.
Reading the description and the method's code, this documentation
clearly refers to the `f` parameter.

This patch fixes the javadoc and aligns it with the method's parameters.


PascalSchumacher:
Thanks!

PR #243, opened 2017-02-22
Fix JsonToStringStyleTest.NestingPerson javadoc

mureinik:
A `{@link}` javadoc can only reference a resolvable class name. The
`{@link}` tag in `NestingPerson`'s javadoc is thus broken, as it directly
references `JsonToStringStyle` without qualifying it with the enclosing
`ToStringStyle` class.

This patch adds the enclosing class to the javadoc, thus "unbreaking"
the reference.


PascalSchumacher:
Thanks!

PR #245, opened 2017-02-24
Use foreach

mureinik:
Since the project no longer supports Java versions older than 6, it's
safe to use the enhanced for loop syntax introduced in Java 5.

This patch employs this syntax where possible to clean up the code.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharSet.java
+++ b/src/main/java/org/apache/commons/lang3/CharSet.java
@@ -175,9 +175,8 @@ public class CharSet implements Serializable {
      */
     protected CharSet(final String... set) {
         super();
-        final int sz = set.length;
-        for (int i = 0; i < sz; i++) {
-            add(set[i]);
+        for (String s : set) {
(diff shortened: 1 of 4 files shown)


PascalSchumacher:
For time to time there are discussion concerning the use of the for-each loop on the mailing list. (e.g. http://markmail.org/thread/wmedbgqmewmpiptk#query:+page:1+mid:gbo5cpywwfxl5p4w+state:results) As far as I remember the consensus was to use it only for arrays, because there it does not have any negative impact on performance. When used with iterables it has some impact on performance (because it requires the creation of an iterator). Therefore some contributors were against using it with iterables in a low level library like commons-lang.

Most of your changes concern arrays, so if you remove the iterable parts I think this is safe to merge.

mureinik:
@PascalSchumacher Missed that discussion somehow. I've reduced the scope of the patch to handle only for loops that go over arrays.

PascalSchumacher:
Thanks! :+1:

PR #246, opened 2017-02-27
DateUtilsTest asserts

mureinik:
Use JUnit's `assertFalse` for assertions with conditions instead of
re-implementing the logic here by testing the condition and throwing an
`AssertionFailureException` if the condition is met.


kinow:
:+1:

PascalSchumacher:
Thanks!

PR #247, opened 2017-02-28
JUnit imports

mureinik:
The `junit.framework` package has been deprecated, and should no longer
be used.

This patch fixes the one remaining import from it to statically import
`org.junit.Assert.assertNull` instead.


kinow:
:+1:

PR #248, opened 2017-03-04
Longs

mureinik:
Style improvement for handling long literals throughout the project.

This patch contains:
- Using long literals instead of casting int literals
- Using uppercase L to signify long literals


PascalSchumacher:
Thanks! :+1:

mureinik:
@PascalSchumacher Using uppercase L for long literals can also easily be enforced by checkstyle. Should I add this check, or would that be more of a nuisance than it's worth?

PascalSchumacher:
+1 to adding the check

mureinik:
@PascalSchumacher I posed PR #249 to do so, thanks.

PR #249, opened 2017-03-04
Checkstyle for long literals

mureinik:
PR #248 corrected all the long literals to use the upper case L
notation.
This patch finishes the job as per the discussion there, and adds a
checktysle check to ensure no code that introduces long literals with
lowercase l are introduced.


PascalSchumacher:
Thanks!

PR #250, opened 2017-03-04
LANG-1300: Convert CharSequence to String for int-methods in StringUtils

MarkDacek:
@chtompki 
[JIRA link](https://issues.apache.org/jira/browse/LANG-1300)


chtompki:
I'll give this a look this evening.

MarkDacek:
If this is too simplistic of a change, consider the minor change I did here. I based it off of the JDK implementation that handles supplementary characters. 

https://github.com/MarkDacek/commons-lang/commit/cde4c529034e182a982fd842252ec410d213c34c

PR #251, opened 2017-03-05
LANG-1300: Change CharSequenceUtils to support supplementary characters

MarkDacek:
@chtompki 
I'm tackling the root cause now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jdk:
   - oraclejdk8
 
 script:
-  - mvn test checkstyle:check javadoc:javadoc -B
+  - mvn test checkstyle:check javadoc:javadoc apache-rat:check -B
 
 after_success:
-  - mvn clean apache-rat:check cobertura:cobertura coveralls:report
+  - mvn clean cobertura:cobertura coveralls:report
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jdk:
   - oraclejdk8
 
 script:
-  - mvn test javadoc:javadoc -B
+  - mvn test checkstyle:check javadoc:javadoc -B
 
 after_success:
   - mvn clean apache-rat:check cobertura:cobertura coveralls:report
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,8 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 
+script:
+  - mvn test javadoc:javadoc -B
+
 after_success:
   - mvn clean apache-rat:check cobertura:cobertura coveralls:report
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -42,6 +42,6 @@ limitations under the License.
     <module name="JavadocMethod">
       <property name="allowUndeclaredRTE" value="true"/>
     </module>
-    <module name="UpperEll">
+    <module name="UpperEll" />
  </module>
 </module>
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -36,9 +36,6 @@ limitations under the License.
     <module name="RedundantImport"/>
     <module name="UnusedImports"/>
     <module name="NeedBraces"/>
-    <module name="RedundantThrows">
-      <property name="allowUnchecked" value="true"/>
-    </module>
     <module name="JavadocMethod">
       <property name="allowUndeclaredRTE" value="true"/>
       <property name="scope" value="public" />
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -41,6 +41,7 @@ limitations under the License.
     </module>
     <module name="JavadocMethod">
       <property name="allowUndeclaredRTE" value="true"/>
+      <property name="scope" value="public" />
     </module>
     <module name="UpperEll" />
  </module>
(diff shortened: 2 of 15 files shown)


chtompki:
Definitely looking better than #250. I'm fiddling around with this locally currently. I'll keep you posted.

@dmjones500 - Since you logged LANG-1300, you have any thoughts on this one?

dmjones500:
Sorry @chtompki, I've been crazy busy lately and haven't moved many issues forward at all.

These changes look good. One thing I've noticed (which is no fault of this change) is the method behaves differently if the input is a String. `String.indexOf(int ch)` returns the index, as measured by _code points_. However the remainder of the method will return the index measured in _code units_.

I haven't had the time to prove this with a unit test, but might be worth looking into this at the same time.

dmjones500:
@chtompki  I'm far from a complexity expert, but if we keep a running tally of the current index in terms of code points, I don't see us going from _O(n)_ to _O(n<sup>2</sup>)_. Wouldn't _O(n<sup>2</sup>)_ imply that for each item in the input, we traverse the entire input again?

chtompki:
@dmjones500 - You're correct there, I wasn't thinking about it properly because I was thinking in terms of `lastIndexOf` moving in from the end of the `CharSequence` and then having to traverse in from the beginning to sort out the number of surrogate pairs in the prefix to the matching character. But, if you simply start with that traversal, then you remain at *O(n)*.

MarkDacek:
Just a question I have on this. 
Are we to do this for ALL calls to this method, or only those where the search char is a supplementary char?

MarkDacek:
@chtompki Please look this over. We have some interesting conflicts here.
1. The issue mentioned above. indexOf("a") will double-count surrogate-pairs if they exist. We can fix this if need be.
2. A string with two surrogate-pairs will read as length 4, though we will treat it as length 2 for the purpose of these operations. E.G. : 
       ```
        final int CODE_POINT = 0x2070E;
    	StringBuilder builder = new StringBuilder();
    	builder.appendCodePoint(CODE_POINT);
        builder.appendCodePoint(CODE_POINT);
       StringUtils.lastIndexOf(builder, CODE_POINT, 2) 
```
the bottom line will return -1, wheras the builder.toString()... version returns 2. I suspect this is just that they disregarded our issue and start at charAt(2) instead of our iterative approach here.

**Comment by dmjones500** [comment]
We need to agree on the desired behaviour of the method. Based on the signature, I think we can adjust it to support supplementary characters without violating the implied contract of the method. The issue for me is about indexes.

Based on the JavaDoc description, I suggest we return the code unit index. This is because the input is a `CharSequence` and the JavaDoc states:

> Finds the first index within a `CharSequence`, handling null.

Also, a `CharSequence` only has methods which operate on code unit indexes. So returning a code unit index would be helpful for further operations on the sequence.

What does everyone else think? We should obviously update the JavaDocs to make it crystal clear what we decide, either way.

**Comment by chtompki** [comment]
@dmjones500 - you think we should bubble this up to the dev list?

**Comment by chtompki** [comment]
Okay. It seems that we'll want some clearer documentation. Note that the javadoc on `indexOf` in the jdk follows:

```java
/**
     * Returns the index within this string of the first occurrence of the
     * specified character, starting the search at the specified index.
     * <p>
     * If a character with value <code>ch</code> occurs in the
     * character sequence represented by this <code>String</code>
     * object at an index no smaller than <code>fromIndex</code>, then
     * the index of the first such occurrence is returned. For values
     * of <code>ch</code> in the range from 0 to 0xFFFF (inclusive),
     * this is the smallest value <i>k</i> such that:
     * <blockquote><pre>
     * (this.charAt(<i>k</i>) == ch) && (<i>k</i> &gt;= fromIndex)
     * </pre></blockquote>
     * is true. For other values of <code>ch</code>, it is the
     * smallest value <i>k</i> such that:
     * <blockquote><pre>
     * (this.codePointAt(<i>k</i>) == ch) && (<i>k</i> &gt;= fromIndex)
     * </pre></blockquote>
     * is true. In either case, if no such character occurs in this
     * string at or after position <code>fromIndex</code>, then
     * <code>-1</code> is returned.
     *
     * <p>
     * There is no restriction on the value of <code>fromIndex</code>. If it
     * is negative, it has the same effect as if it were zero: this entire
     * string may be searched. If it is greater than the length of this
     * string, it has the same effect as if it were equal to the length of
     * this string: <code>-1</code> is returned.
     *
     * <p>All indices are specified in <code>char</code> values
     * (Unicode code units).
     *
     * @param   ch          a character (Unicode code point).
     * @param   fromIndex   the index to start the search from.
     * @return  the index of the first occurrence of the character in the
     *          character sequence represented by this object that is greater
     *          than or equal to <code>fromIndex</code>, or <code>-1</code>
     *          if the character does not occur.
     */
```

Let's see if we can get something along these lines on both CharSequenceUtils as well as StringUtils. Furthermore, I think we should stick as closely to the implementation in `String` as possible simply generalizing to `CharSequence`'s.

MarkDacek:
Chiming in here. I'd recommend using code units so that it's compatible with the other operations like substring. Of course, this would still break if you called substring(index+1) since you'd be taking the second byte of a supplementary character, but I'd still argue that this is the path to take.

chtompki:
I'm going to make some documentation changes here. I'll keep you apprised as I make them.

chtompki:
Going to fix the javadoc build and then pull this in.

PR #252, opened 2017-03-05
[LANG-1314] Fix javadoc creation on with Java 8

mureinik:
Java 8's javadoc seems to be stricter than Java 7's, and the current head current fails to generate javadoc (`mvn javadoc:javadoc`) with Java 8.

This PR contains a series of fixes to the various classes' javadoc which are required for Java 8 (but would also improve the javadoc generated with earlier java versions) and then adds the `javadoc:javadoc` target to Travis CI so we can guarantee it won't break again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,8 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 
+script:
+  - mvn test javadoc:javadoc -B
+
 after_success:
   - mvn clean apache-rat:check cobertura:cobertura coveralls:report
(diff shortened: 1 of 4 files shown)


PascalSchumacher:
Thanks!

PascalSchumacher:
@mureinik Sorry it wasn't my intention to steal the authorship of your commits with the javadoc fixes. :( I wasn't aware that a `git rebase` changes the author.

mureinik:
@PascalSchumacher no harm done. I'd rather have commons-lang's javadoc build properly than one patch more or less to my so-called name.

PR #253, opened 2017-03-09
Added a restart method for convenience

esend7881:
The restart function will allow lazy people like me to simple call `stopwatch.restart()` instead of manually calling the `stop`, `restart` `start` methods one by one, or recreating the object itself.


PascalSchumacher:
Hi @esend7881 

thanks for the pull request. 

I'm not sure it is a good idea to add at method that is nearly identical to `StopWatch#createStarted`.

esend7881:
@PascalSchumacher good point. I would say the main advantage is my method would not have the JVM re-create an object. So if you are restarting the timer over and over, it may be less costly. In the end, same outcome.

PR #256, opened 2017-03-13
fix for LANG-1310

drajakumar:


src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
```diff
@@ -110,32 +114,32 @@ private void privateStuff() {
 
         @SuppressWarnings("unused")
         private String privateStringStuff() {
-        	return "privateStringStuff()";
+         return "privateStringStuff()";
```
jvz: Think you could fix the formatting here?

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
```diff
@@ -357,16 +369,23 @@ public void testInvokeMethod() throws Exception {
...
                 5, "a", "b", "c"));
-
+        assertEquals("foo(long...)", MethodUtils.invokeMethod(testBean, "foo",
+                1L, 2L));
+        
+        try {
+            MethodUtils.invokeMethod(testBean, "foo",
+                    1, 2);
+            fail("should throw NoSuchMethodException");
+        } catch (final NoSuchMethodException e) {
```
Abrasha: Maybe it will be better to extract it to another test method with expected exception?

```
@Test(expected = NoSuchMethodException.class)
public void methodName(){

}
```

PascalSchumacher:
@drajakumar It's really great to get a pull request for this. :+1:  

I took a look at this issue some time ago, but gave up trying to solve it. Sadly I'm too tired for a review at the moment. I will get back to you in a few days, if nobody has reviewed/merged the pull request already.

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
Abrasha: @PascalSchumacher what do you think about refactoring existing tests? Or it is fine to community? I can take care about it

Claudenw: I find that the expect exception works well only with very short tests.  In this case if any of the earlier method calls throws the NoSuchMethodException the test will not fail.  In my opinion the current construct is the better one.

drajakumar:
the comment on indentation from@jvz is fixed now.

PascalSchumacher:
Thanks! :+1:

PR #209, opened 2016-11-06
LANG-1164: allow ToStringStyle to omitNulls

PascalSchumacher:

PR #259, opened 2017-03-18
LANG-1167: Add null filter to ReflectionToStringBuilder

MarkDacek:
@chtompki 
First attempt at this. In order to avoid a large refactor or breaking existing implementations, the boolean has to be a negative (isExcludeNullValues instead of appendNullValues).

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
@@ -300,6 +300,64 @@ public class ReflectionToStringBuilder extends ToStringBuilder {
         return new ReflectionToStringBuilder(object, style, null, reflectUpToClass, outputTransients, outputStatics)
                 .toString();
     }
+    
+    /**
+     * <p>
+     * Builds a <code>toString</code> value through reflection.
+     * </p>
+     *
+     * <p>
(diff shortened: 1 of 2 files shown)


chtompki:
@PascalSchumacher - Do you think we should take this approach, or the approach you went after in [LANG-1164](https://issues.apache.org/jira/browse/LANG-1164).

PascalSchumacher:
@chtompki Sorry for the delay, other open source projects and vacations interfered.

As the ToStringStyle implementations are singletons, I do not see a save way to use #209 as a basis for this pull request. Maybe I'm missing something?

I think #209 and this can coexists (as long as the naming is aligned).

@MarkDacek Thanks for creating this pull request. I think it would be great to have this feature in commons-lang. :+1:

HiAscend:
hello ,
on V3.8 V3.8.1 V3.9
when i use
`ReflectionToStringBuilder.toString(student, ToStringStyle.JSON_STYLE)`
return some like
{"id":666,"age":22,"name":"\u4E2D\u56FD","birthday":"Thu Jul 04 13:29:26 CST 2019","book":{"name":null}}
the student name is "中国" but now it is in unicode charset

however
on V3.7 it is ok
see：https://bbs.csdn.net/topics/392557654?page=1#post-404105720

PR #261, opened 2017-03-25
LANG-1317 Adds MethodUtils#findAnnotation and extend MethodUtils#getMethodsWithAnnotation for non-public, super-class and interface methods

yasserzamani:
In order to fix [WW-4744](https://issues.apache.org/jira/browse/WW-4744) ,

findAnnotation will be an extension for Method.getAnnotation that also searches interfaces and super classes.

findMethodsWithAnnotation will be an extension for getMethodsWithAnnotation that also supports non public methods, super class and interface methods.

NOTES:
To keep changes simple, currently no cache provided. Let's do that when needed 😃

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -53,14 +53,6 @@ public class ClassUtils {
     }
 
     /**
-     * Inclusivity literals for {@link #getAllSuperclassesAndInterfaces(Class, Priority)}.
-     * @since 3.6
-     */
-    public enum Priority {
-        SUPERCLASS, INTERFACE
-    }
-
-    /**
      * The package separator character: <code>'&#x2e;' == {@value}</code>.
      */
     public static final char PACKAGE_SEPARATOR_CHAR = '.';
@@ -470,11 +462,10 @@ public class ClassUtils {
      * from interfaces, and so on in a breadth first way.</p>
      *
      * @param cls  the class to look up, may be {@code null}
-     * @param p  determines what to peek in same breadth, the superclass or interface
      * @return the {@code List} of superclasses in order going up from this one
      *  {@code null} if null input
      */
-    public static List<Class<?>> getAllSuperclassesAndInterfaces(final Class<?> cls, Priority p) {
+    public static List<Class<?>> getAllSuperclassesAndInterfaces(final Class<?> cls) {
         if (cls == null) {
             return null;
         }
@@ -496,7 +487,7 @@ public class ClassUtils {
             } else if (sci < ifi) {
                 acls = allSuperclasses.get(sci++);
             } else {
-                acls = (p == Priority.SUPERCLASS ? allSuperclasses.get(sci++) : allInterfaces.get(ifi++));
+                acls = allInterfaces.get(ifi++);
             }
             classes.add(acls);
         }
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -53,6 +53,14 @@ public class ClassUtils {
     }
 
     /**
+     * Inclusivity literals for {@link #getAllSuperclassesAndInterfaces(Class, Priority)}.
+     * @since 3.6
+     */
+    public enum Priority {
+        SUPERCLASS, INTERFACE
+    }
+
+    /**
      * The package separator character: <code>'&#x2e;' == {@value}</code>.
      */
     public static final char PACKAGE_SEPARATOR_CHAR = '.';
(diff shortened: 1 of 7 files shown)


PascalSchumacher:
Thanks for the pull request!

It is really unfortunate  that the existing method `getMethodsWithAnnotation` is not called `getAccessibleMethodsWithAnnotation` : (, because then the new method could be called `getMethodsWithAnnotation`.

I would not have been able to tell the difference between `getMethodsWithAnnotation` and `findMethodsWithAnnotation` without reading the javadoc.

I believe we need a more intention revealing method name or maybe  we should just add an overload of `getMethodsWithAnnotation` e.g. `getMethodsWithAnnotation(Class cls, annotationCls, boolean ignoreAccessiblitiy)`.

What do you think?

yasserzamani:
@PascalSchumacher , thank you! I was overloaded by another job and forgot that I also can use method overloads. Yes, strongly I agree. Just I have some doubts about name `ignoreAccessiblitiy` because the new method is not only about accessibility, but also about traverse super classes and implemented interface and super interfaces while the annotation maybe is not present on the given class but is present on lower layers. Maybe we should also add another param named `searchSupers` for example.

PascalSchumacher:
You are right. The method signature I suggested omits the important super part. I agree it should be something like `getMethodsWithAnnotation(Class cls, annotationCls, boolean searchSupers, boolean ignoreAccess)`. (I'm not a fan of boolean parameters (especially multiple ones), so if somebody has an idea for a good method name, that is welcome.)

yasserzamani:
Another bad thing is that `getMethodsWithAnnotation(cls, annotationCls, false, false)` will be a duplicate for `getMethodsWithAnnotation(cls, annotationCls)`.

Another solution is to keep my current changes not modified but add following javadoc:

```
 *
 * <p>To lookup annotations on the given class level only choose {@code get*()} methods
 * and to lookup in the entire inheritance hierarchy of the given class and ignore 
 * accessibility, choose {@code find*()} methods</p>
 *
 * @since 2.5
 */
public class MethodUtils {
```

PascalSchumacher:
Well we have to keep `getMethodsWithAnnotation(cls, annotationCls)` for compatibility reasons.  In my opinion it is not problem if is equal to `getMethodsWithAnnotation(cls, annotationCls, false, false)`. It can be refactored to call `getMethodsWithAnnotation(cls, annotationCls, false, false)` and then it is just some sugar for the most? common use case.

yasserzamani:
👌 I am working on a new commit. thanks!

yasserzamani:
@PascalSchumacher , I made it better 😃 thanks a lot!

Please feel free to make any recommendation if you think I can make it better.

> Coverage decreased (-0.03%) to 94.571%

😞 Do you know why? 👆

Abrasha:
@yasserzamani I thinks because you added more conditional branches in the latest commit and that is why less LOC are covered

yasserzamani:
Thank you @Abrasha ,

I signed up with [coveralls.io](http://coveralls.io) and see that all of this PR's changes are covered with unit tests: [MethodUtils](https://coveralls.io/jobs/24310263/source_files/797094203#L845) , [ClassUtils](https://coveralls.io/jobs/24310263/source_files/797094115#L477).

Also these files show enhancement in coverage at [COVERAGE CHANGED](https://coveralls.io/jobs/24310263) :thinking:

PascalSchumacher:
Concerning the coverage: I think it's because the coverage differs between java versions. The build used to determine coverage depends on the order in which the travis builds finish.

PascalSchumacher:
@yasserzamani Sorry for the delay, other open source projects and vacations interfered.

I'm not sure if the priority parameter of `getAllSuperclassesAndInterfaces` is necessary? Are you aware of any use-cases where it is necessary? Maybe we should keep things simple and remove it?

Otherwise the pull request is fine. :=) :+1:

yasserzamani:
> Sorry for the delay, other open source projects and vacations interfered.

You're welcome, I understand and as an open source fun contributor, I'm patient enough :)

> I'm not sure if the priority parameter is necessary?

No, is not. I just thought I should make it more general with more options to prevent future needs for other user's PRs. Now, with thanks to your recommendations I remove it :)

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -457,6 +457,44 @@ public static String getAbbreviatedName(final String className, final int len) {
     }
 
     /**
+     * <p>Gets a combination of {@link #getAllSuperclasses}(Class)} and
+     * {@link #getAllInterfaces}(Class)}, one from superclasses, one
+     * from interfaces, and so on in a breadth first way.</p>
+     *
+     * @param cls  the class to look up, may be {@code null}
+     * @return the {@code List} of superclasses in order going up from this one
```
PascalSchumacher: This should also mention Interfaces.

Can you add a since tag to this method?

Thanks!

PascalSchumacher:
Created https://issues.apache.org/jira/browse/LANG-1321 for the `ClassUtils#getAllSuperclassesAndInterfaces` addition.

PascalSchumacher:
If I'm not mistaken the list returned by `getAllSuperclassesAndInterfaces` can contain the same interface multiple times if it is implemented by more than one super class. To prevent this the result should be collected in `LinkedHashSet` first (similar to https://github.com/yasserzamani/commons-lang/blob/90e252f571377cac39f5e2a3fc73749f589c24de/src/main/java/org/apache/commons/lang3/ClassUtils.java#L448). What do you think?

yasserzamani:
@PascalSchumacher , sorry for my bad decisions. Again I thought I should keep it public to provide same functionality for someone who needs in future :)

Now, I removed it to MethodUtils as a private method. I also added @since tag and mentioned interfaces in @return java doc.

Thanks a lot!

PascalSchumacher:
@yasserzamani Thanks!

No need to apologize. It is just that we try to be conservative with new additions (If we add a public method we have to support it for a long time and can not do any incompatible changes.). If the method is deemed generally useful it can always be moved and made public later on.

PascalSchumacher:
Merged. Thank you very much.

PR #262, opened 2017-04-01
Findbugs in Travis CI

mureinik:
This PR adds FindBugs to Travis CI so it can be used to automatically evaluate new patches instead of having to do so retroactively.

It contains a series of patches that either fix existing FindBugs errors or excludes them (either because the violation is intentional or because FindBugs cannot properly assess the code) - individual commit messages explain individual decisions.
At the end of that series of patches is a patch that adds the `findbugs:check` goal to Travis CI, so it would now be applied on any new PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jdk:
   - oraclejdk8
 
 script:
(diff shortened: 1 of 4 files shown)


PascalSchumacher:
Thanks! :+1:

PR #263, opened 2017-04-22
Use String#contains where possible

mureinik:
Since the project defines a JDK 7 source compatibility, it's safe to use JDK 5's features.
This patch replaces usages of `String#indexOf` with `String#contains` where possible to make the code easier to read and maintain.


PascalSchumacher:
Thanks!

PR #268, opened 2017-05-25
.travis.yml: add oraclejdk9

PascalSchumacher:


kinow:
+1 lgtm!

britter:
Travis build failed. Any idea why? Java 8 has an heap space error, while the Java 9 build reports an unrecognized VM option.

kinow:
Ooh, looked at the wrong green icon. JDK 7 complained about JAVA_HOME location... weird. Related issue in travis-ci repo https://github.com/travis-ci/travis-ci/issues/5520

PascalSchumacher:
Another problem is that the maven version used by travis does not work with jdk 9 as far as I know.

ameyjadiye:
As per Travis-ci documentation doesnt seems they support oraclejdk9 

https://docs.travis-ci.com/user/languages/java/ 

>The Travis CI environment provides Oracle JDK 7 (default), Oracle JDK 8, OpenJDK 6, OpenJDK 7, Gradle 2.0, Maven 3.2 and Ant 1.8, and has sensible defaults for projects that use Gradle, Maven or Ant.

Thats why I added Docker Image for ibmjdk8 to Commons text https://github.com/apache/commons-text/pull/45

PascalSchumacher:
It is not officially supported, but it is possible.

For example https://github.com/jacoco/jacoco and https://github.com/apache/logging-log4j2 use jdk 9 on travis.

ameyjadiye:
Hmm, how about using docker service for this ? , I have raised same for ibmjdk8 https://github.com/apache/commons-lang/pull/271

PascalSchumacher:
The build works using java 9 locally. I guess its only a few months at most until jdk 9 will be officially supported on travis, so I won't invest any more time into this for now.

PR #269, opened 2017-06-06
LANG-1337: Fix test failures in IBM JDK 8 for ToStringBuilderTest by specifying the ArrayList initial capacity.

kinow:
See https://issues.apache.org/jira/browse/LANG-1337 for issue description and explanation of the changes here.

In summary, we specify the ArrayList's initial capacity, this way the String built through reflection has the expected value. Tested with Oracle JDK 7 and 8, and IBM JDK 8.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
@@ -36,7 +36,8 @@ import org.junit.Test;
  */
 public class ToStringBuilderTest {
 
-    private final Integer arraylistInitialCapacity = Integer.valueOf(10);
+    // See LANG-1337 for more.
+    private static final int ARRAYLIST_INITIAL_CAPACITY = 10;
     private final Integer base = Integer.valueOf(5);
     private final String baseStr = base.getClass().getName() + "@" + Integer.toHexString(System.identityHashCode(base));
 
@@ -317,7 +318,8 @@ public class ToStringBuilderTest {
         // representation different for IBM JDK 1.6.0, LANG-727
         assumeFalse("IBM Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".equals(SystemUtils.JAVA_SPECIFICATION_VERSION));
         assumeFalse("Oracle Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".compareTo(SystemUtils.JAVA_SPECIFICATION_VERSION) < 0);
-        final List<Object> list = new ArrayList<>(arraylistInitialCapacity);
+        // LANG-1337 without this, the generated string can differ depending on the JVM version/vendor
+        final List<Object> list = new ArrayList<>(ARRAYLIST_INITIAL_CAPACITY);
         final String baseString = this.toBaseString(list);
         final String expectedWithTransients = baseString + "[elementData={<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>},size=0,modCount=0]";
         final String toStringWithTransients = ToStringBuilder.reflectionToString(list, null, true);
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
```diff
@@ -316,7 +316,7 @@ public void testReflectionHierarchyArrayList() {
         // representation different for IBM JDK 1.6.0, LANG-727
         assumeFalse("IBM Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".equals(SystemUtils.JAVA_SPECIFICATION_VERSION));
         assumeFalse("Oracle Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".compareTo(SystemUtils.JAVA_SPECIFICATION_VERSION) < 0);
-        final List<Object> list = new ArrayList<>();
+        final List<Object> list = new ArrayList<>(10);
```
sebbASF: I think that needs a comment.
Is the magic number 10 significant?
If so, what determines the value?
Could it ever change?

andyklimczak:
probably related to [this](https://stackoverflow.com/a/34250231)?

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
andyklimczak: probably related to [this](https://stackoverflow.com/a/34250231)?

sebbASF: Thanks!

If the test fails when the initial size arg is omitted, does that not also affect the behaviour of the method being tested? i.e. do apps also have to ensure that they specify the min size when using ToStringBuilder.reflectionToString() ?

kinow: Fair enough on the magic number. I'd thought about that, then noticed a few other tests with numbers. But one broken window doesn't mean I can break another one :-) fixing in another commit.

kinow: >If the test fails when the initial size arg is omitted, does that not also affect the behaviour of the method being tested?

Not really. The test simply checks the string built for an arraylist through reflection. The issue was caused for believing that the lazy initialization (as @andyklimczak) would work in the same independent of the JVM.

What the test is verifying is correct, the current approach could be improved to make the test less flaky.

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
```diff
@@ -316,7 +317,7 @@ public void testReflectionHierarchyArrayList() {
         // representation different for IBM JDK 1.6.0, LANG-727
         assumeFalse("IBM Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".equals(SystemUtils.JAVA_SPECIFICATION_VERSION));
         assumeFalse("Oracle Corporation".equals(SystemUtils.JAVA_VENDOR) && "1.6".compareTo(SystemUtils.JAVA_SPECIFICATION_VERSION) < 0);
-        final List<Object> list = new ArrayList<>();
+        final List<Object> list = new ArrayList<>(arraylistInitialCapacity);
```
britter: Sorry to be nitpicking, but this should be a constant and wie should add a comment referencing JIRA-1337 with an explanation why we need to pass the initial capacity.

kinow: Roger that. Will add a note to myself to fix the other final member variables later... trying to be concise, but I'm clearly missing the point here :-) was supposedly to be a very simple fix for this issue. Pushing a new commit in a few minutes, just finishing to review commons-fileupload vote.

kinow:
Comments added, received some feedback, but would still be useful someone with the last IBM JDK 8 to give it a try and confirm it works for him/her :)

PR #274, opened 2017-07-01
Fix ExceptionUtils#getStackFrame(String) javadoc

mureinik:
`ExceptionUtils#getStackFrame(String)`'s javadoc contains a broken reference to `SystemUtils#LINE_SEPARATOR` (as the class name is not fully qualified, and there's no import of `org.apache.commons.lang3.SystemUtils`).

This patch fixes the broken reference by replacing it it with a reference to `System#lineSeparator()`, which the method actually uses.


PascalSchumacher:
Thanks!

PR #275, opened 2017-07-09
[WIP] LANG-1339: replace java.beans.PropertyListener by java.util.Observable

kinow:
For discussion, regarding [LANG-1339](https://issues.apache.org/jira/browse/LANG-1339) (TL;DR: remove depedency on java.desktop from [lang], preparing for Java 9). Thought having some code to discuss would be helpful.

This pull request contains a possible solution, where java.beans.{PropertyChangeListener,PropertyChangeListener} are replaced by java.util.{Observer,Observable}.

Users would still have the same feature, but alas binary compatibility would not be kept. So perhaps this could be in a 4.x release. The other methods are deprecated in the next 3.x releases, and we remove the dependency to java.desktop on 4.x, and also offer a Java9 module with no dependency java.desktop.

Would this be a viable alternative?

Cheers
Bruno


britter:
Removing the dependency in 4.x is the way to go. I don't see a way to get this into 3.x :-(

kinow:
Yup @britter you are definitely correct. At least if we agree on the solution, we can think about marking some methods as deprecated in the current circuit breaker (though I prefer to mark as deprecated when there is an alternative), and/or add notes in the next release notes, alerting users about the change to come.

jodastephen:
The PR is indeed backwards incompatible.

My suggestion would be to add two new classes with the fixed code (different class names), and deprecated the old classes. That way there is no need to have commons-lang v4.

kinow:
>My suggestion would be to add two new classes with the fixed code (different class names), and deprecated the old classes. That way there is no need to have commons-lang v4.

Thought a bit about that after seeing the vote for [lang] 3.7. In any way, the current implementation will be removed only in 4.x, and we will have to include the require static for Java 9 module.

So right now I am thinking about not marking the class or fields as deprecated, nor adding another class with a different name. Just keep this PR and the ticket open. Then mark the ticket as FixVersion 4.0, use the require static trick for the module-info.java for Java 9.

Does that sound like a good plan?

kinow:
@jodastephen done. Kept the existing classes, but added new ones where the only modification is replacing `java.beans` by `java.util` equivalent classes.

Existing classes were annotated with `@deprecated` with a comment pointing to the new class.

WDTY? I'd like to sort it out as soon as possible to sort out the issue with dependencies & Java 9 in lang.

Thanks!
Bruno

ps: old code was removed from commit line. Moved instead to a branch at https://github.com/kinow/commons-lang/tree/LANG-1339-old, just in case we need to compare it or someone wants to see what it was before

kinow:
Posted to the mailing list and will wait to see if there is any further feedback. Otherwise I will merge it in the next days. Thank you @jodastephen !

PascalSchumacher:
+1

kinow:
Back to WIP. Please avoid merging it for now. Gilles noted in the mailing list that the Observable/Observer classes have been deprecated in Java 9. So we will probably go with the deprecation/removal path instead.

Link to the discussion in the mailing list: https://markmail.org/message/efy52lul5pzbznut

kinow:
Going to have another go at it in the next days. Just deprecate everything, create new classes with Java 8 Consumer that is called for every change of state in the event breaker, instead of PropertyListener or Observable.

src/main/java/org/apache/commons/lang3/concurrent/BaseCircuitBreaker.java
```diff
@@ -0,0 +1,152 @@
...
+    /** The current state of this circuit breaker. */
+    protected final AtomicReference<State> state = new AtomicReference<>(State.CLOSED);
+
+    /**
+     * Consumer called every time the circuit breaker state changes if not {@code null}.
+     */
+    private final Consumer<State> consumer;
+
+    /**
+     * Creates an {@code AbstractCircuitBreaker}. It also creates an internal {@code PropertyChangeSupport}.
```
oheger, line 40: There is still an outdated reference to PropertyChangeSupport in the commend.

kinow: Argh, good spotting @oheger ! Let me fix that. Thanks!

kinow: Done

PR #277, opened 2017-07-11
run maven defaultGoal from Travis

ameyjadiye:
Hi @garydgregory ,

As we have added defaulGoals in pom.xml, putting them again in travis make things redundant, by simply specifying ```mvn``` we can trigger default goals from pom.xml, we already did this for commons-text

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ jdk:
   - oraclejdk8
 
 script:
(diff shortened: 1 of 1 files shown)


garydgregory:
This PR has no description. Can you please specify why we want this change?

ameyjadiye:
@garydgregory , I think this ia good chahe to accept .

garydgregory:
Patch applied. Thank you!

PR #279, opened 2017-07-22
Fraction debug printings

mureinik:
Removed commented out debug printing from the Fraction class.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -269,7 +269,6 @@ public final class Fraction extends Number implements Comparable<Fraction> {
         double delta1, delta2 = Double.MAX_VALUE;
         double fraction;
         int i = 1;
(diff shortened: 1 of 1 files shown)


mureinik:
How is this code useful? It adds nothing to the functionality, nor does it help explain the code. It seems like the original author used it while developing this class, but once it's done, I see no value in keeping it.

ameyjadiye:
debugging can be much effectively done with IDE, adding these kind of console output is discouraged.
I would recommend to use logging _logger.debug()_, but that's not the guidelines in commons. 
so +1 to remove it.

kinow:
I think what Gary meant is that that comment is harmless, and can be used by other developers that want/need to debug the code, and see what is going on. I've used it in some projects myself, to indicate to fellow developers parts of the code that may be a bit fuzzy or complicated. I'm OK if it's kept, but OK too if others prefer to remove it (if we replace it by a comment like "Here we can see variable x" I'm OK as well).

PR #282, opened 2017-08-10
SwappedPair constructed as Pair.of(rhs,lhs)

namannigam:
There seems to be a bug in the current implementation of the `isRegistered` implementation, where the `swappedPair` is constructed similarly to the existing `pair` to check their existence in registry.

Source of finding - [StackOverflow#45603317](https://stackoverflow.com/questions/45603317/)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
@@ -16,16 +16,16 @@
  */
 package org.apache.commons.lang3.builder;
 
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.ClassUtils;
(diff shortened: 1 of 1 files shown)


kinow:
Makes sense, good catch. However, would be nice to have a unit test at least to cover that case, and make sure we don't have any regression for this behaviour. Would you be able to create one @namannigam ?

Thanks!!!
Bruno

garydgregory:
+1 on the unit test request, we need it to avoid regressions.

PascalSchumacher:
created https://issues.apache.org/jira/browse/LANG-1349 to track this

namannigam:
@kinow @garydgregory Sure would try to come back soon with a unit test for the use-case.

PascalSchumacher:
I added a test. Thanks for the pull request! :+1:

PR #283, opened 2017-08-20
LANG-1350: Fix varargs array invokeConstructor

bjkail:
If ConstructorUtils.invokeConstructor(Class, Object...) is invoked with an array of arguments whose classes do not match a constructor exactly (for example, a subclass is used) and an array is used for a varargs parameter, then the array is not matched to the varargs parameter type.


PascalSchumacher:
Thanks for the pull request! :+1: 

It was merged by @garydgregory in https://github.com/apache/commons-lang/commit/cc94767e7eabdfcf9d1cab1d8d1d8556864394c6

It would be nice if you could close the pull request.

Thanks!

PR #285, opened 2017-09-08
make checkstyle config more portable (no maven coupling)

osiegmar:


kinow:
Checked out pull request branch, built with the change, got no checkstyle errors. Then removed the suppression (without re-adding the previous value in pom.xml) and also got the same result, without any errors. Maybe it is aware that the suppression file must be ignored?

+1 for the change. Will wait for others to review :+1: 

Documentation about the SuppressionFilter module.

http://checkstyle.sourceforge.net/config_filters.html#SuppressionFilter

Thank you for your contribution @osiegmar

osiegmar:
Removing the SuppressionFilter results in thousands of errors. Maybe you have a stale checkstyle cache file `target/cachefile`?

kinow:
Not sure. Here's what I just tried.

```
$ git reset --hard
$ git status 
    On branch pr/285
    ....
$ vim checkstyle.xml
$ # removed <property name="file" value="checkstyle-suppressions.xml"/> line
$ mvn clean site
```

And then opening the Checkstyle report (via file:///home/.../commons-lang/target/site/index.html) there are no warnings or error in the checkstyle page.

Are you able to spot which step I am missing to reproduce it?

osiegmar:
I haven't checked the configuration of the maven site plugin in detail. Does it really generates a site for the current branch? I doubt that, as I've seen subversion commands in the config...

Anyway, to run checkstyle, just invoke `mvn clean checkstyle:check`.

kinow:
>Does it really generates a site for the current branch?

It should. That is one step in the release process. Probably due to the site using another configuration (under the reporting tag).

https://github.com/osiegmar/commons-lang/blob/01a820275fc7bd66cb699dd081a432c7dde2645e/pom.xml#L711

Could you update the pull request with your original change duplicated in that portion, please?

osiegmar:
> Could you update the pull request with your original change duplicated in that portion, please?

Done. Thanks for pointing that out!

kinow:
Thank you for updating the pull request.

Tested locally. Using the changes in the pull request I get the same result as we get when the settings are in the Maven settings.

Commenting out the suppression filter in the checkstyle configuration, the report in the web site now gives me **2828** errors. So the suppression filter is working indeed.

Looks good to me, merging it now.

PR #287, opened 2017-09-20
Clean up findbugs exclusions for old XYZRange classes

mureinik:
Commit 56e830a removed all the old `XYZRange` classes from
org.apache.lang3.math, and left only the generic `Range<T>` class.

This patch removes the left-over findbugs exclusions for those
classes.


PascalSchumacher:
Thanks! :+1:

PR #289, opened 2017-09-22
Boolean comparisons in CharRange

mureinik:
Cleaned up comparisons to false to just use the boolean negation operator (`!`).


PascalSchumacher:
Thanks!

PR #290, opened 2017-09-26
Added string methods

iashok22:
1. indexOfAll - Finds and returns all the indexes of given character
2. isAnagram - Tells whether given string is anagram with another string


garydgregory:
The anagram method belongs in commons-text IMO.

britter:
I agree with @garydgregory. @chtompki WDYT?

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -9245,5 +9245,39 @@ public static String unwrap(final String str, final char wrapChar) {
...
+    public static List<Integer> indexOfAll(final String source, final Character searchKey) {
+    	if(source == null || source.length() == 0 || searchKey == null ) {
+    		return null;
+    	}
+    	List<Integer> indexList = new ArrayList<>();
+		for(int i = 0 ; i < source.length() ; i++) {
+			if(searchKey.equals(source.charAt(i))) {
+				indexList.add(i);
+			}
+		}
```
Abrasha: Could you fix the indentation please? Let it be properly aligned.

PR #294, opened 2017-10-05
Added indexesOf method

iashok22:
indexesOf () - Finds the indexes of all the occurrences of given search key found in the given string.


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -9245,25 +9245,28 @@ public static String unwrap(final String str, final char wrapChar) {
...
+     * @param searchKey 
+     *           search character
      * @return list of integer of indexes.
      */
     public static List<Integer> indexesOf(final CharSequence source, final Character searchKey) {
     	if(isEmpty(source) || searchKey == null ) {
-    	    return null;
-    	}
+			return null;
+		}
```
Abrasha: Can you fix the formatting, please?

vsTianhao: Ha-ha，They refused to merge only two lines of code modification

garydgregory:
The question here is whether to use the more grammatically correct "indices" or the more casual "indexes".

kedar-joshi:
Also, length of the `source` can be cached for better performance e.g. 


```java
for (int i = 0, length = source.length(); i < length; i++)
{
    if (searchKey.equals(source.charAt(i)))
    {
	indexList.add(i);
    }
}
```

PR #296, opened 2017-10-09
LANG-1355: Add FastTimeZone to decrease TimeZone.getTimezone latency

chonton:
Adding FastTimeZone to decrease latency of timezone lookups.   I'll hold this PR open for 48 hours for commons committers' comments.  Thanks!

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
+++ b/src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
@@ -29,6 +29,10 @@ public class FastTimeZone {
 
     private static final TimeZone GREENWICH = new GmtTimeZone(false, 0, 0);
 
+    // do not instantiate
+    private FastTimeZone() {
+    }
+
     /**
      * Get the GMT TimeZone.
      * @return A TimeZone with a raw offset of zero.
@@ -39,10 +43,11 @@ public class FastTimeZone {
 
     /**
      * Get a TimeZone, looking first for GMT custom ids, then falling back to Olson ids.
-     * A GMT custom id has an optional prefix of GMT, followed by sign, hours digit(s), optional
-     * colon(':'), and optional minutes digits: <em>[GMT] (+|-) Hours [[:] Minutes]</em>
+     * A GMT custom id can be 'Z', or 'UTC', or has an optional prefix of GMT,
+     * followed by sign, hours digit(s), optional colon(':'), and optional minutes digits.
+     * i.e. <em>[GMT] (+|-) Hours [[:] Minutes]</em>
      *
-     * @param id A GMT custom id or Olsen id
+     * @param id A GMT custom id (or Olson id
      * @return A timezone
      */
     public static TimeZone getTimeZone(String id) {
@@ -56,8 +61,8 @@ public class FastTimeZone {
     private static final Pattern GMT_PATTERN = Pattern.compile("^(?:(?i)GMT)?([+-])?(\\d\\d?)?(:?(\\d\\d?))?$");
 
     /**
-     * Get a TimeZone with GMT offsets.  A GMT offset must be either 'Z' or match
-     * (GMT)? hh?(:?mm?)?, where h and m are digits representing hours and minutes.
+     * Get a TimeZone with GMT offsets.  A GMT offset must be either 'Z', or 'UTC', or match
+     * <em>(GMT)? hh?(:?mm?)?</em>, where h and m are digits representing hours and minutes.
      *
      * @param pattern The GMT offset
      * @return A TimeZone with offset from GMT or null, if pattern does not match.
--- a/src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
+++ b/src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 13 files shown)


PascalSchumacher:
Looks good imho. 👍

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -0,0 +1,90 @@
...
+            return tz;
+        }
+        return TimeZone.getTimeZone(id);
+    }
+
+    private static final Pattern GMT_PATTERN = Pattern.compile("^(?:(?i)GMT)?([+-])?(\\d\\d?)?(:?(\\d\\d?))?$");
+
+    /**
+     * Get a TimeZone with GMT offsets.  A GMT offset must be either 'Z' or match
+     * (GMT)? hh?(:?mm?)?, where h and m are digits representing hours and minutes.
```
kinow: Maybe instead of

> A GMT offset must be either 'Z' or match (GMT)? hh?(:?mm?)?

It should be

> A GMT offset must be 'Z', or 'UTC', or match (GMT)? hh?(:?mm?)?

?

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Faster methods to produce custom time zones.
+ *
+ * @since 3.7
+ */
+public class FastTimeZone {
```
kinow: Do we need a private constructor to prevent instantiation of FastTimeZone? It seems to contain only static methods. Not sure if that's the intended design. But noticed it while looking at the cobertura report (which is looking great).

src/main/java/org/apache/commons/lang3/time/GmtTimeZone.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.Date;
+import java.util.TimeZone;
+
+/**
+ * Custom timezone that contains offset from GMT.
+ *
+ * @since 3.7
+ */
+class GmtTimeZone extends TimeZone {
```
kinow: TimeZone is Serializabe. Do we need to add a serialVersionUID here?

src/test/java/org/apache/commons/lang3/time/GmtTimeZoneTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test
+    public void hoursInRange() {
+        Assert.assertEquals(23 * 60 * 60 * 1000, new GmtTimeZone(false, 23, 0).getRawOffset());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void minutesOutOfRange() {
+        Assert.assertEquals(0, new GmtTimeZone(false, 60, 0));
```
kinow: This test is wrong. Its title states that the minutes will be out of range, but the hour is actually out of range (60). Minute is 0, but never gets checked.

src/test/java/org/apache/commons/lang3/time/GmtTimeZoneTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    public void useDaylightTime() {
+        Assert.assertFalse(new GmtTimeZone(false, 0, 0).useDaylightTime());
+    }
+
+    @Test
+    public void inDaylightTime() {
+        Assert.assertFalse(new GmtTimeZone(false, 0, 0).useDaylightTime());
+    }
```
kinow: Maybe add something like

```
@Test
    public void testToString() {
        Assert.assertEquals("[GmtTimeZone id=\"GMT+23:00\",offset=82800000]", new GmtTimeZone(false, 23, 0).toString());
    }

    @Test
    public void testGetOffset() {
        Assert.assertEquals(82800000, new GmtTimeZone(false, 23, 0).getOffset(1, 1, 1, 1, 1, 1));
    }
```

With these two tests we reach 100% for GmtTimeZone.

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
chonton: done

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
chonton: done

src/main/java/org/apache/commons/lang3/time/GmtTimeZone.java
chonton: great catch! done

src/test/java/org/apache/commons/lang3/time/GmtTimeZoneTest.java
chonton: great catch! thanks.
done.

src/test/java/org/apache/commons/lang3/time/GmtTimeZoneTest.java
chonton: done

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -39,10 +43,11 @@ public static TimeZone getGmtTimeZone() {
...
     /**
      * Get a TimeZone, looking first for GMT custom ids, then falling back to Olson ids.
-     * A GMT custom id has an optional prefix of GMT, followed by sign, hours digit(s), optional
-     * colon(':'), and optional minutes digits: <em>[GMT] (+|-) Hours [[:] Minutes]</em>
+     * A GMT custom id can be 'Z', or 'UTC', or has an optional prefix of GMT,
+     * followed by sign, hours digit(s), optional colon(':'), and optional minutes digits.
+     * i.e. <em>[GMT] (+|-) Hours [[:] Minutes]</em>
      *
-     * @param id A GMT custom id or Olsen id
+     * @param id A GMT custom id (or Olson id
```
PascalSchumacher: Nitpick: either a superfluous `(` or a missing `)`

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Get the GMT TimeZone.
+     * @return A TimeZone with a raw offset of zero.
+     */
+    public static TimeZone getGmtTimeZone() {
+        return GREENWICH;
+    }
+
+    /**
+     * Get a TimeZone, looking first for GMT custom ids, then falling back to Olson ids.
```
garydgregory: In future patches, you can use the active voice for Javadocs: "Gets a TimeZone..." instead of "Get...". :-)

PR #297, opened 2017-10-09
Add a rule of Locale.ENGLISH to String.toUpperCase() method.

BruceKuiLiu:
A String is being converted to upper case, using the platform's default encoding.
This may result in improper conversions when used with international characters.
Use the String.toUpperCase( Locale l ) version should be better.
http://findbugs.sourceforge.net/bugDescriptions.html#DM_CONVERT_CASE

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -5501,7 +5501,7 @@ public class StringUtils {
          }
          String searchText = text;
          if (ignoreCase) {
-             searchText = text.toLowerCase();
+             searchText = text.toLowerCase(Locale.ENGLISH);
              searchString = searchString.toLowerCase();
          }
          int start = 0;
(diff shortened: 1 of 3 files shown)


BruceKuiLiu:
Thanks.

garydgregory:
Shouldn't this be done with Locale.ROOT?

BruceKuiLiu:
Right, Locale.ROOT should be the most appropriate.

src/main/java/org/apache/commons/lang3/time/FormatCache.java
```diff
@@ -240,7 +240,7 @@ public boolean equals(final Object obj) {
             // Eliminate the usual boilerplate because
             // this inner static class is only used in a generic ConcurrentHashMap
             // which will not compare against other Object types
-            return Arrays.equals(keys, ((MultipartKey)obj).keys);
+            return obj instanceof MultipartKey && Arrays.equals(keys, ((MultipartKey)obj).keys);
```
PascalSchumacher: Please do not include unrelated changes in a single pull request. Thanks!

garydgregory:
I am splitting out the FastDateParser changes here: https://issues.apache.org/jira/browse/LANG-1357

garydgregory:
I fixed some tests and code main code here and there and I think all the changes we need are in. If you feel otherwise, please update this PR based on git master. Thank you!

PR #298, opened 2017-10-09
Add an instanceof test in the implementation of equals(Object obj).

BruceKuiLiu:
The equals(Object obj) method shouldn't make any assumptions about the type of obj. It should simply return false if obj is not the same type as this.
http://findbugs.sourceforge.net/bugDescriptions.html#BC_EQUALS_METHOD_SHOULD_WORK_FOR_ALL_OBJECTS


kinow:
I think it makes sense. Any chance to include a unit test to protect against regressions in the future? Thanks

PascalSchumacher:
The comment above the statement explains why it is done this way.

In case this is merged the comment should be removed.

BruceKuiLiu:
I am sorry, I cannot provide any test cases.  We found it is a potential vulnerability.

PR #299, opened 2017-10-10
Add module-info for Java 9

jodastephen:
Add the module-info.java file
Make the appropriate changes to pom.xml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,6 @@ install:
 script:
   - mvn -e -B
 
+# coveralls does not work on Java 9 at present
 after_success:
   - if [ $TRAVIS_JDK_VERSION == "openjdk7" ] || [ $TRAVIS_JDK_VERSION == "oraclejdk8" ]; then mvn -e -B clean cobertura:cobertura coveralls:report -Ptravis-cobertura; fi
-  - if [ $TRAVIS_JDK_VERSION == "oraclejdk9" ]; then mvn -e -B clean cobertura:cobertura -Ptravis-cobertura; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,10 @@ language: java
 sudo: false
 
 jdk:
-  - openjdk7
-  - oraclejdk8
   - oraclejdk9
 
 script:
   - mvn
 
 after_success:
-  - mvn clean cobertura:cobertura coveralls:report -Ptravis-cobertura
+  - mvn clean cobertura:cobertura -Ptravis-cobertura
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
   - mvn --version
 
 script:
-  - mvn clean install -e -B
+  - mvn -e -B
 
 after_success:
   - if [ $TRAVIS_JDK_VERSION == "openjdk7" ] || [ $TRAVIS_JDK_VERSION == "oraclejdk8" ]; then mvn -e -B clean cobertura:cobertura coveralls:report -Ptravis-cobertura; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,23 @@ language: java
 sudo: false
 
 jdk:
+  - openjdk7
+  - oraclejdk8
   - oraclejdk9
 
+cache:
+  directories:
+    - "$HOME/.m2/repository"
+
+before_cache:
+  - rm -rf $HOME/.m2/repository/org/apache/commons/commons-lang3
+
+install:
+  - mvn --version
+
 script:
-  - mvn
+  - mvn clean install -e -B
 
 after_success:
-  - mvn clean cobertura:cobertura -Ptravis-cobertura
+  - if [ $TRAVIS_JDK_VERSION == "openjdk7" ] || [ $TRAVIS_JDK_VERSION == "oraclejdk8" ]; then mvn -e -B clean cobertura:cobertura coveralls:report -Ptravis-cobertura; fi
+  - if [ $TRAVIS_JDK_VERSION == "oraclejdk9" ]; then mvn -e -B clean cobertura:cobertura -Ptravis-cobertura; fi
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -20,7 +20,10 @@
  */
 module org.apache.commons.lang3 {
 
-    // see AbstractCircuitBreaker
+    // the AbstractCircuitBreaker class relies on java.desktop
+    // this is a large dependency, and unsuited for most users of commons-lang
+    // as such, the dependency is declared to be "static", making it optional
+    // if you use AbstractCircuitBreaker, then you must add "requires java.desktop" to your module file
     requires static java.desktop;
 
     // all packages are exported
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Apache Commons Lang provides utility classes for the core of Java.
+ */
+module org.apache.commons.lang3 {
+
+    // see AbstractCircuitBreaker
+    requires static java.desktop;
+
+    // all packages are exported
+    exports org.apache.commons.lang3;
+    exports org.apache.commons.lang3.arch;
+    exports org.apache.commons.lang3.builder;
+    exports org.apache.commons.lang3.concurrent;
+    exports org.apache.commons.lang3.event;
+    exports org.apache.commons.lang3.exception;
+    exports org.apache.commons.lang3.math;
+    exports org.apache.commons.lang3.mutable;
+    exports org.apache.commons.lang3.reflect;
+    exports org.apache.commons.lang3.text;
+    exports org.apache.commons.lang3.text.translate;
+    exports org.apache.commons.lang3.time;
+    exports org.apache.commons.lang3.tuple;
+
+}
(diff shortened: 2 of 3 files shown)


britter:
Awesome! Would be create if you could create and reference a JIRA ticket as described in CONTRIBUTING.md, so this will show up in our release notes.

.travis.yml
```diff
@@ -17,12 +17,10 @@ language: java
 sudo: false
 
 jdk:
-  - openjdk7
```
stokito: Why did you removed the build for jdk7 and 8?
If lang3 still support them then it should be runned tests on 7 and 8

jodastephen: With the proposed change, the build can only occur on Java 9. (Java 9 is needed to compile `module-info.java`).

stokito: Sounds sad :( Especially because, I guess, jdk8 compiler willn't just ignore the `module-info.java`.
Maybe some other projects will even try to generate the `module-info.java`

britter: I think we should discuss this on the developer mailing list.

michaelsavich:
Is there any way to rewrite AbstractCircuitBreaker to eliminate the dependency on java.desktop? Java.desktop is quite a large dependency to have for the sake of one class.

kinow:
@michaelsavich I have a proposal in a branch, submitted some months ago https://github.com/apache/commons-lang/pull/275

Waiting for feedback to prepare deprecated annotations in place, and merge in a major release (I believe it breaks BC)

jodastephen:
@michaelsavich , this PR does not create a full dependency on `java.desktop`. The dependency is expressed as `requires static`, which means that it will not be pulled in by default. Users that want to use the two broken classes on Java 9 in module mode (a very small percentage of commons-lang users) will need to manually include the `java.desktop` dependency. As such, there is no urgency to get `AbstractCircuitBreaker` removed.

michaelsavich:
Ah, gotcha. That makes sense.

src/main/java/module-info.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Apache Commons Lang provides utility classes for the core of Java.
+ */
+module org.apache.commons.lang3 {
+
+    // see AbstractCircuitBreaker
+    requires static java.desktop;
```
namannigam: a small difference and I am just curious to know, why is this not `transitive` which was what I saw using `jdeps --generate-module-info`?

jodastephen: Because this is a massive dependency, and we need to avoid dragging it all in for just a small percentage of users. Using `requires static` is an abuse of the module system, but the best option available. `requires transitive` would be the right solution normally.

britter:
I'll bring this up a last time on the ML to make sure nobody has objections against merging this.

.travis.yml
```diff
@@ -21,8 +21,19 @@ jdk:
...
+  - mvn --version
+
 script:
-  - mvn
+  - mvn -e -B
 
 after_success:
-  - mvn clean cobertura:cobertura coveralls:report -Ptravis-cobertura
+  - if [ $TRAVIS_JDK_VERSION == "openjdk7" ] || [ $TRAVIS_JDK_VERSION == "oraclejdk8" ]; then mvn -e -B clean cobertura:cobertura coveralls:report -Ptravis-cobertura; fi
+  - if [ $TRAVIS_JDK_VERSION == "oraclejdk9" ]; then mvn -e -B clean cobertura:cobertura -Ptravis-cobertura; fi
```
PascalSchumacher: Nitpick: I think we can remove this line, because running `cobertura` without uploading to `coveralls` mostly just wastes cpu cycles.

jodastephen: The right answer will be to move from Cobertura to JaCoCo (as JaCoCo is the more alive project). But right now, Coveralls also doesn't work with JaCoCo.

jodastephen:
`mvn install` works fine on each JDK version AFAICT. `mvn site` still breaks on cobertura on Java 9 and I'm not sure I can work around it without removing it from the parent pom.

namannigam:
@jodastephen What's the error that you get with site? Any links to the build?

jodastephen:
@namannigam You need to check it out and run `mvn clean install site` on Java 9.

sigursoft:
Hi guys,
What's the status on this pull request? 😄

jodastephen:
Adding `module-info` causes some older/Android projects to fail as they can't handle the class format. So, projects are left with the choice of moving forwards or breaking these older/Android projects.

This issue is also tied up in the need to modernise the commons parent pom.xml for Java 9 and later.

jodastephen:
Its not just compiling, there are issues at runtime still AFAIK, especially with Java EE and older bytecode libraries.

src/main/java/module-info.java
lploom: Isn't the best option to refactor the concurrent package to separate project? Doesn't seem like a great fit here anyhow.

johnou:
@jodastephen can you be more specific?

jodastephen:
There is a lot of complexity in commons-parent and commons generally. Getting a build to work for Joda was hard enough, and there I am willing to insist on releases being done on Java 9 or later. There needs to be a lot more work on the commons build before `module-info` should be considered, however at the moment its not clear as to whether doing that work would be positive (due to Android, Jakarta EE issues etc)

jfallows:
@jodastephen [moditect](https://github.com/moditect/moditect) can be configured to generate `module-info.class` into `META-INF/versions/9/module-info.class` instead, maintaining compatibility with Java8 and earlier runtimes.

GedMarc:
@jodastephen @jfallows 
I have successfully managed to get moditect on lang3, io, fileupload and text - 

https://github.com/GedMarc/commons-lang
https://github.com/GedMarc/commons-io
https://github.com/GedMarc/commons-fileupload

Using source and target of 1.6 is still supported to build for that target JDK (I have done this successfully with the org.json project), however the actual build has to happen in JDK 8 for the plugin to work.

When building the module-info, no import statements are allowed as yet

I place them directly in the root, @ClassGraph (Duke Choice Winner), places the module info files in the meta-inf/versions/ correctly with moditect. He also participates in the moditect development.

https://github.com/GedMarc/commons-lang/blob/master/src/jre11/java/module-info.java

jfallows:
@GedMarc if you add the following to `<configuration>` section of `moditect-maven-plugin` version `1.0.0.Beta2`, it will place `module-info.class` into `META-INF/versions/9/module-info.class` instead.
```
<jvmVersion>9</jvmVersion>
```

GedMarc:
@jfallows Yeah I've gone through a couple iterations and different mechanisms, about a year and half ago there was an issue in a few of the older ones which I'm sure this project will care about, in things like Wildfly 10/EAP 7 using Jandex the scanner would warning out on the class structure and items like that, so "hiding" the class there _seemed_ like a good idea, or the older reflections etc,

For me though I'm planning on discontinuing JDK 8 soon though, so for my purposes and to build cleanly with a full distribution cycle jdk 12 13 and up, the root definitely seems like the best idea and 190 odd projects later (Assisting Jackson JSON, having to use local builds of Guice no replies on the pull requests, stopped making pull requests because of that actually, commons).

JLink, and JPMS, has it's requirements though, but the results are spectacular.

GedMarc:
@jfallows for interests sake though - 
https://github.com/GedMarc/JPMSThirdPartyParent/tree/86537e0f18de56e10c3d9da4cb5a68f9273b7e87 is a listing of all the libraries that I'm expecting full JPMS compliance with in the short term 

The list of modules that are either finalized/closed/owner's non-contactable, or the repositories abandoned are here
https://github.com/GedMarc/JPMSThirdPartyParent/tree/86537e0f18de56e10c3d9da4cb5a68f9273b7e87

If you can help in getting a hold of these guys, specially javax.inject, man that would be fantastic

jodastephen:
I've seen error reports from using multi-version jar files as well as from just placing the `module-info.class` in the root. It is perfectly possible to create a fully modularized commons-lang provided the team are willing to exclude some users who can't handle `module-info.class` or multi-version jar files.

jfallows:
@GedMarc thanks for the links, will help however I can.

I'm intrigued by the _italics_ of "_seemed_ like a good idea". 😄 

Libraries designed with a base version of Java9 can handle `module-info.class` at the root, whereas older libraries designed with base version of `Java8` or lower might invalidate assumptions about class file contents when placing `module-info.class` at root, but that should be largely mitigated by placing `module-info.class` under `META-INF/versions/9` instead for runtime discovery by Java9 and higher.

Did you find that there were remaining issues where libraries scanned for all class files inside a JAR, even if they were under `META-INF/versions/9` and still tripped up on `module-info.class` due to their historical assumptions about class file structure? If so, is the proposed solution for those libraries to make them resilient to the presence of `module-info.class` in general, whether at the root or under `META-INF/versions/9`, i.e. treat like a bug and fix them?

@jodastephen please share details about those error reports if possible, just in case there are additional relevant pitfalls to navigate 👍

garydgregory:
The last time I tried to use a jar with a Java 9 class file properly embedded in META-INF it blew up the Android toolchain. It's been a while so I am not sure what the current state is...

GedMarc:
Ping https://github.com/apache/commons-fileupload/pull/26

GedMarc:
Basic Update - 

Maven Coordinates have changed, all libraries are up to date and modular, can simply include them to build up JLink artifacts

http://www.guicedee.com

Sineaggi:
@kinow I have an alternative branch that would allow users to continue to use `AbstractCircuitBreaker` with or without the `java.desktop` module, available here https://github.com/apache/commons-lang/compare/master...Sineaggi:commons-lang:make-java.desktop-optional.

In this case we provide two ways of specifying a listener, one from `java.desktop` and on provided by apache. For users that still use the `java.beans.PropertyChangeListener`, they'll need a `requires java.desktop` in their module-info.java file, but for other users they can use the new `org.apache.commons.lang3.concurrent.StateChangeListener` without issue. I could also do some extra work to make the new classes closely mirror the ones in `java.beans`.

Sineaggi:
@jodastephen Will you have any time to work on this branch and get it ready to merge?

jodastephen:
I have no plans to work further on this, and in my 2023 view adding `module-info.java` is a mistake for most projects, as it creates more problems than it solves. Having a module name reserved in the manifest is, of course, fine.

garydgregory:
Why the edit to change the import of PropertyChangeListener to a FQCN?
If the class still has a reference to PropertyChangeListener how is it different?

Sineaggi:
> Why the edit to change the import of PropertyChangeListener to a FQCN? If the class still has a reference to PropertyChangeListener how is it different?

You're right, that change is unnecessary. I originally used the same names for the classes in `java.beans`.

GedMarc:
I have all the commons libraries with module-info's for natively building artifacts

https://github.com/GedMarc/GuicedEE-Services/tree/2a56ec5e79010c8214b6a398d3cf5b4e09f844ce

These are always up to date and contain the required configurations for operation

https://mvnrepository.com/artifact/com.guicedee.services/ - prefix "commons"

e.g. https://mvnrepository.com/artifact/com.guicedee.services/commons-logging

garydgregory:
Closing: This is currently done through the moditect Maven plugin.

PR #304, opened 2017-10-27
Fix DateUtilsTest to work reliably on Java 9

jodastephen:
Fixes failures due to MST time-zone not being formatted as "MST"
Also sets and unsets time-zones in finally blocks
to avoid pollution global state when tests fail

See [LANG-1362](https://issues.apache.org/jira/browse/LANG-1362)


jodastephen:
Test failure looks like a timing issue unrelated to this PR

garydgregory:
The current git master build OK for me with 'mvn clean verify' and Oracle Java 9.0.1.

PascalSchumacher:
@garydgregory For me too and the build on travis with java 9 has been sucessfully running for months now. 

This seems to be a locale related thing. I can replicated the 8 test failures by adding `-Duser.country=GB -Duser.language=en` to surefire `argLine`.

PR #306, opened 2017-11-07
.travis: add IBM JDK 8

PascalSchumacher:


PascalSchumacher:
Looks like the IBM JDK is not yet supported by the travis jdk-switcher, so let's wait some time.

PR #308, opened 2017-11-14
LANG-1366 : Add Feature for No ClassName and MultiLine StringStyle

pckeyan:
Added sub class to print with No Class Name in Multiline.
Added Junit.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
@@ -2648,6 +2648,8 @@ public abstract class ToStringStyle implements Serializable {
 
     }
 
+    // ----------------------------------------------------------------------------
+
     /**
      * <p>
      * <code>ToStringStyle</code> that outputs with No Class Name in a Multi Line format.
@@ -2688,6 +2690,5 @@ public abstract class ToStringStyle implements Serializable {
         private Object readResolve() {
             return ToStringStyle.NO_CLASS_NAME_MULTI_LINE_STYLE;
         }
-
     }
 }
--- a/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
@@ -158,6 +158,20 @@ public abstract class ToStringStyle implements Serializable {
     public static final ToStringStyle JSON_STYLE = new JsonToStringStyle();
 
     /**
+     * The no class name Multi Line toString style. Using the <code>Person</code>
+     * example from {@link ToStringBuilder}, the output would look like this:
+     *
+     * <pre>
+     * name=John Doe
+     * age=33
+     * smoker=false
+     * </pre>
+     *
+     * @since 3.7
+     */
+    public static final ToStringStyle NO_CLASS_NAME_MULTI_LINE_STYLE = new NoClassNameMultiLineToStringStyle();
+
+    /**
      * <p>
      * A registry of objects used by <code>reflectionToString</code> methods
      * to detect cyclical object references and avoid infinite loops.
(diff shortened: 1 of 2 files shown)


smolnar82:
It would be great to add this new TSS. Thanks!

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -2633,4 +2647,48 @@ private Object readResolve() {
...
+            this.setContentEnd(System.lineSeparator());
+        }
+
+        /**
+         * <p>Ensure <code>Singleton</code> after serialization.</p>
+         *
+         * @return the singleton
+         */
+        private Object readResolve() {
+            return ToStringStyle.NO_CLASS_NAME_MULTI_LINE_STYLE;
```
kinow: Really nit-picking here, but other classes in this file are referencing it without the class name (as this is the same class...). So if possible, for consistency it could be the same, i.e. `return NO_CLASS_NAME_MULTI_LINE_STYLE;`.

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -2633,4 +2647,48 @@ private Object readResolve() {
...
+     * <code>ToStringStyle</code> that outputs with No Class Name in a Multi Line format.
+     * </p>
+     *
+     * <p>
+     * This is an inner class rather than using
+     * <code>StandardToStringStyle</code> to ensure its immutability.
+     * </p>
+     *
+     * @since 3.7
+     * @see <a href="http://json.org">json.org</a>
```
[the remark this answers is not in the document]
pckeyan: No, I removed the comment.

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -2633,4 +2647,48 @@ private Object readResolve() {
...
+     * <p>
+     * <code>ToStringStyle</code> that outputs with No Class Name in a Multi Line format.
+     * </p>
+     *
+     * <p>
+     * This is an inner class rather than using
+     * <code>StandardToStringStyle</code> to ensure its immutability.
+     * </p>
+     *
+     * @since 3.7
```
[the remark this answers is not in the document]
pckeyan: I updated the version to 3.10

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
pckeyan: Updated the return type.

src/test/java/org/apache/commons/lang3/builder/NoClassNameMultiLineToStringStyleTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Unit tests {@link org.apache.commons.lang3.builder.NoClassNameMultiLineToStringStyleTest}.
+ */
+public class NoClassNameMultiLineToStringStyleTest {
```
Stzx: As mentioned in #461 , maybe consider doing this work ahead of time?

@kinow , what do you think?

kinow: Hi @Stzx 

Unfortunately I haven't had much experience with JUnit 5 yet. Looks to me like it will support multiple modifiers. But I couldn't find a recommendation, or even any deprecation, that would require dropping the `public` modifier. Do you know if that's a best practice, recommendation, or if it will be deprecated/required later?

I found at least one file in their repo using the `public` modifier: https://github.com/junit-team/junit5/blob/f52ded2d2f331ca7dba738b53afa5b1c0f22a85d/platform-tests/src/jmh/java/org/junit/jupiter/jmh/AssertionBenchmarks.java

kinow: Found one comment that appears to be from someone involved with the project: https://stackoverflow.com/a/55230350

Doesn't seem to be required to use private, public, etc. It's up to the project.

Stzx: From the links you provided, as well as what I saw from the official examples ([junit5-jupiter-starter-maven](https://github.com/junit-team/junit5/blob/f52ded2d2f331ca7dba738b53afa5b1c0f22a85d/platform-tests/src/jmh/java/org/junit/jupiter/jmh/AssertionBenchmarks.java)) and [user guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests) seems to be determined by the project?

Looks different styles of different projects.

kinow: Yeah, and if it's not something that can be checked with checkstyle or with the other build tools, we have no clear way to enforce... and having to ask users to alter their PR's can be a bit frustrating for the contributors.

For this reason I think it might be easier to just keep things as they are, and accept new submissions using either public, private, or package-protected.

garydgregory:
TBH (and sorry for coming in a bit late), I am against a PR like this one. 

My concern here is that there is a huge number of configuration combinations possible for `ToStringStyle`. As I am sure we do not plan on adding all possible options, adding this one just feels like fulfilling some narrow use case while Commons Lang is a general purpose library. 

I'd hope not to see this class grow and grow and grow overtime; but, if this changes comes in, it becomes harder to say no to other niche configuration combinations.

This specific PR is so, well, specific, that while I am sure the author feels it is general purpose enough for inclusion here, I would claim it is not and that it belongs in their application, not Commons Lang.

kinow:
@garydgregory I don't have a use case for this, so can't argue with your review.

@pckeyan if you could elaborate more on your use case, or maybe others could chime in, then maybe your pull request could still be merged.

Otherwise, this is probably something that will have to be implemented on your applications as a small utility class :+1:

garydgregory:
FWIW, I have a bunch of these application specific ToStringStyles in various projects but I do not think it appropriate to add these here.

pckeyan:
My usecase is to print multiline without classname. Usage I had was in MapReduce application. I thought it would be a contribution to reuse as there is no available option without classname.

PR #310, opened 2017-12-20
A small mistake in javadoc description.

nickbarban:
Javadoc says that:
`StringUtils.getDigits("1123~45") = "12345"`
but really:
`StringUtils.getDigits("1123~45") = "112345"`


PascalSchumacher:
Thanks! 👍

PR #311, opened 2018-01-03
LANG-1373 Stopwatch based capability for nested, named, timings 

ottobackwards:
There are times when you want to do a number or related timings across a sequence of calls or operations.  This is difficult to do with just the StopWatch.

StackWatch provides an abstraction over the  StopWatch class that allows callers to create multiple named and possibly nested timing operations.

StackWatch uses a combination of Deque and a custom Tree implementation to create, start and end timing operations.

A Visitor pattern is also implemented to allow for retrieving the results after the completion of the operation, and timings may be tagged to allow the consumer to filter the results.

I have built this in my personal travis and all three jobs pass

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/StackWatch.java
+++ b/src/main/java/org/apache/commons/lang3/time/StackWatch.java
@@ -1,0 +1,597 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.lang3.time;
+
+import org.apache.commons.lang3.Validate;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * <p>
+ * The {@code StackWatch}, provides a wrapper around the {@code StopWatch} for creating multiple and
+ * possibly nested named timings.
+ * </p>
+ * <p>
+ * While the {@link StopWatch} provides functionality to time the length of operations, there is no
+ * context or name to go with the time tracked. It is also not possible to time nested calls with
+ * the {@code StopWatch}.
+ * </p>
+ * <p>
+ * {@code StackWatch} provides that functionality, allowing successive calls to
+ * {@link StackWatch#startTiming} to track nested calls.
+ * </p>
+ * <p>
+ * At the end of a timing 'run', a visitor interface provides the ability to visit all the timing
+ * 'nodes' and capture their output, including the level of the call if nested. The visitor is passed
+ * the {@code level} of the current node, it's path elements, and a {@link Timing} instance.
+ * </p>
+ * <p>
+ * Through the {@link Timing} the visitor can access the {@link StopWatch} instance for the timing.
+ * </p>
+ * <p>
+ * The {@link TimingNodeInternal} provide a tree structure in support of nesting.
+ * A {@code Deque} is use to track the current time node.
+ * </p>
+ * <p>
+ * The generic type {@code N} is used to type the {@link TimingNodeInternal#name}.
+ * The generic type {@code T} is used to type the {@link TimingNodeInternal#tags}.
+ * </p>
+ * <pre>
+ *   <code>
+ *    private void outerFunction() {
+ *      try {
+ *        StackWatch{@code <String,String> watch = new StackWatch<>("OuterFunction");}
+ *        watch.start();
+ *        functionOne(watch);
+ *        watch.stop();
+ *        watch.visit({@code new StackWatch.TimingVisitor<String,String>()} {
+ *         {@literal @}Override
+ *          public void visitTiming({@code int level, List<String> path,  StackWatch.Timing<String,String>} timing) {
+ *            System.out.println("Visit level " + level + " timing: " + timing.getName()) + " path: " + path + " time: " + timing.getStopWatch().getNanoTime() + "ns");
+ *          }
+ *        });
+ *      } catch (Exception e){}
+ *    }
+ *
+ *    private void functionOne({@code StackWatch<String,String>} watch) throws Exception {
+ *      watch.startTiming("One", "tagOne");
+ *      functionOneOne(watch);
+ *      watch.stopTiming();
+ *    }
+ *
+ *    private void functionOneOne({@code StackWatch<String,String>} watch) throws Exception {
+ *      watch.startTiming("OneOne", "tagTwo");
+ *      functionOneTwo(watch);
+ *      watch.stopTiming();
+ *    }
+ *
+ *    private void functionOneTwo({@code StackWatch<String,String>} watch) throws Exception {
+ *      watch.startTiming("OneTwo", "tagThree");
+ *      watch.stopTiming();
+ *    }
+ *   </code>
+ * </pre>
+ * <p>
+ * The example above would result in the following output.
+ * </p>
+ * <pre>
+ *   <code>
+ *    Visit level 0 timing: OuterFunction path: [OuterFunction] time: 316032287ns
+ *    Visit level 1 timing: One path: [OuterFunction, One] time: 156596831ns
+ *    Visit level 2 timing: OneOne path: [OuterFunction, One, OneOne] time: 106207892ns
+ *    Visit level 3 timing: OneTwo path: [OuterFunction, One, OneOne, OneTwo] time: 52851139ns
+ *   </code>
+ * </pre>
+ * <p>
+ * This class is not thread safe, and is meant to track timings across multiple calls on the same
+ * thread
+ * </p>
+ *
+ * @since 3.8
+ */
+public class StackWatch<N,T> {
+    /**
+     * The Deque used to track the timings
+     */
+    private Deque<TimingNode> deque = new LinkedList<>();
+
+    /**
+     * The root {@code TimingNodeInternal}.
+     * The root node represents the root of a tree structure, and contains all {@code TimingNodeInternal}
+     * instances.
+     */
+    private TimingNode<N,T> rootNode;
+
+    /**
+     * <p>
+     * Constructor
+     * </p>
+     * <p>
+     * The top most timing will be created with the rootName on {@link StackWatch#start()} ()}
+     * </p>
+     * @param rootName the root name
+     * @throws NullPointerException if rootName is null
+     */
+    public StackWatch(N rootName) {
+        Validate.notNull(rootName,"The rootName cannot be null");
+        rootNode = new TimingNodeInternal<>(rootName);
+    }
+
+    /**
+     * <p>
+     * Returns the root name.
+     * </p>
+     *
+     * @return the root name.
+     */
+    public N getRootName() {
+        return this.rootNode.getName();
+    }
+
+    /**
+     * <p>
+     * Starts the {@code StackWatch}.
+     * </p>
+     * <p>
+     * A root timing will be created named for the rootName and started.
+     * </p>
+     * <p>
+     * If not called before the first {@link  StackWatch#startTiming} call, then the
+     * {@code StackWatch} will be started at that time.
+     * </p>
+     *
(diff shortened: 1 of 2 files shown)


ottobackwards:
bump?

ottobackwards:
bump

kinow:
Ack :-) haven't had time to review it as we have a short summer around here, so have added a note to have a look at StopWatch (which I'm not familiar with) and at this variation.

Said that, had a very brief peek at the code from the browser without using the IDE. The code looks great! Only small minor things I could see were a duplicated white line (which doesn't matter tbh) and the the formatting. If I recall correctly, [lang] uses 4 spaces instead of 2? Though I could be wrong.

Thanks for being so patient. I intend to review it as soon as I get some spare time (IOW once the weather gets back to our normal 10-17C cloudy days with with windy rains). But happy if anyone beats me to it.

Cheers
Bruno

kinow:
Ah, and might be a good idea to run `mvn clean site`, then have a look at the reports output. Some pull requests are delayed merging due to the introduction of findbugs/checkstyle/etc issues. Quickly running it and fixing any reported issues might make it easy and faster to get it merged :)

ottobackwards:
You bet, thanks.  If there is any guide that I should be following let me know.

kinow:
Oh, good point. There's the [CONTRIBUTING.md](https://github.com/apache/commons-lang/blob/master/CONTRIBUTING.md). There are several links in that doc too, apologize for not having a shorter version.

The tabs/spaces is mentioned at the first link under *Additional Resources*, but there are heaps of other useful information in there. Hopefully not too different from other ASF projects, so you can probably filter through 90% or more of the contents in those files, and just recognize one of two differences in the way [lang]/Commons expects patches and pull requests.

Ta

ottobackwards:
Thanks, I did try to follow that, I use travis so the different java builds worked.  I used the local checkstyle xml, and through I had caught everything.

ottobackwards:
Do I have to do something to get my stuff added to the reports in the local site?  I see my tests run in the cli, but they are not in the reports, or the java doc etc

ottobackwards:
maybe  i didn't run clean, i'll try again

ottobackwards:
OK, I wasn't looking in target, I have the reports. smh.

ottobackwards:
The reports actually look good for my stuff, except the coverage stuff, if I am reading this correctly

kinow:
Perfect. So maybe it is not enforcing the 4 spaces? Haven't looked at the xml settings we have in lang in a while.

So if you fixed the 2/4 spaces, we won't have any other nit-picks like this, and be able to focus on the feature. Especially since you added good docs, unit cases, and even a practical use case :-)

ottobackwards:
I will take care of it :)

src/main/java/org/apache/commons/lang3/time/StackWatch.java
```diff
@@ -0,0 +1,329 @@
...
+ * This class is not thread safe, and is meant to track timings across multiple calls on the same
+ * thread
+ * </p>
+ */
+public class StackWatch {
+
+  /**
+   * The default name for the root level timing if not provided
+   */
+  public static final String DEFAULT_ROOT_NAME = "ROOT_TIMING";
```
kinow: Here is needs to be four spaced as well. You can look at another file and compare that (e.g. https://github.com/apache/commons-lang/blob/f50ec5e608286b0c48d6b9b4c792352de8353804/src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java#L58).

kinow:
Sorry, what I meant was that indentation in the lang is done with 4 spaces. Added an example to another class. Hope that makes sense.

ottobackwards:
I am so sorry, I'll take care of it.

kinow:
I was the one who was not clear enough :-) sorry

src/main/java/org/apache/commons/lang3/time/TimingRecordNode.java
```diff
@@ -0,0 +1,222 @@
...
+        return String.format(PATH_FMT, parentTimingPath, timingName);
+    }
+
+    /**
+     * Returns the child nodes of this node.
+     *
+     * @return Iterable of the child nodes.
+     */
+    public Iterable<TimingRecordNode> getChildren() {
+        return children;
```
kinow: Couldn't we return an read-only copy of the children here?

src/main/java/org/apache/commons/lang3/time/TimingRecordNode.java
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * The tags associated with this timing.
+     */
+    private String[] tags;
+
+    /**
+     * The child nodes of this node.
+     */
+    private List<TimingRecordNode> children = new LinkedList<>();
```
kinow: Would it make much difference if we used an `ArrayList` here? We seem to `#add` only in `createChild()`, and not sure if we are using any head/tail operation, nor inserting with indexes. So maybe having an `ArrayList` would give us the same functionality for less memory?

src/main/java/org/apache/commons/lang3/time/TimingRecordNode.java
ottobackwards: done.

src/main/java/org/apache/commons/lang3/time/TimingRecordNode.java
ottobackwards: good point

ottobackwards:
Close and re-open to kick travis.  My travis built fine, not sure why apache failed

ottobackwards:
bump

kinow:
Pull request looks good. Toyed with the code a bit (https://gist.github.com/kinow/c0873daf069c2e64bc69d6305400e610), and I like the current design.

@ottobackwards let's ask in the ML if anybody else would like to review it, and if nobody replies by the weekend, consider it merged. During the merge, I'd like to squash the commits to keep the commit history/tree simpler. Would you mind squashing the commits and updating the pull request to one single commit (or a few if you think it's necessary), please?

Cheers
B

kinow:
E-mail sent. See the latest messages in the Commons Developers mailing list, with the tag [lang]. There should be one from me about this PR and your JIRA ticket, asking for more reviews.

Thanks for being so patient @ottobackwards !!

Bruno

ottobackwards:
would a rebase -i with FIXUP work?  What git command would *you* run?  Do we have to worry about authorship?

kinow:
I would run something like `git rebase -i HEAD~123` where 123 is the number of commits I'm going back. Then squash the commits into a single one, discarding commit messages. Finally, would use a single commit message like "LANG-1373: The description here...". And then `git fetch --all` followed by a `git rebase origin/master` to make sure it's at the most recent change in the remote repository as well. If no merge conflicts, you can simply `push -f` to your branch, and that should do it.

In case you are concerned about losing your changes, just do from your branch a `git checkout -b backup-1373` or some other name, `git checkout -` to go back to your branch, and give it a try (or do it in the other branch, whichever you prefer)

kinow:
Oh, and you have a pending comment in your JIRA ticket from another reviewer :-)

ottobackwards:
yes, i'm interested on your take on the jira stuff from today @kinow

kinow:
@ottobackwards I'm +1 for using generics, as well as the fluent API suggested by Gilles just now. But I'd have to look at the code to see if that's doable. I think you are probably the best to take a look if that'd a) be a good idea, b) be doable, or c) argument why not.

The generics I suspect might be easier to achieve. Fluent API's are normally useful, but easily to be overdone IMO.

ottobackwards:
Just a note, discussion of this issue has been split between here and the jira issue.

kinow:
Well, no, but I quite like your example. Will give it a try to see how it would look like  without looking at the code (mobile right now), my only concern would be using try-with-resource when we don't open a real resource nor throw any exception (can't recall that part).

But I liked the start static method in special. WDYT @ottobackwards?

ottobackwards:
I did think of that at the start.  There were two issues for me:

1. I didn't want to expose the internal node to the caller and complicate the interface
2. In my scenario ( a language engine, that *might* be timing things in the repl or *might not* be timing things at runtime in a storm topology, there may not be a watch at all, so having:

if(context.WatchOptional.isPresent())(
try(xxxxxx){
TIMED_FUNCTION()
}
} else {
TIMED_FUNCTION()
}

didn't seem as clean.

ottobackwards:
That doesn't mean I don't want to re-visit the idea....

ottobackwards:
```java   
 Optional<StackWatch> watchOptional = state.context.getWatch();
    watchOptional.ifPresent((sw) -> sw.startTiming("lambda", "LAMBDA"));
     Object result = apply(localState);
     watchOptional.ifPresent(StackWatch::stopTiming);
     return result;
```
```java
 Optional<StackWatch> watchOptional = state.context.getWatch();
Object result;
if(watchOption.isPresent() {
   try( TimingNode timing = watchOption.get().startTiming("lambda","LAMBDA")) {
       result = apply(localState);
    }
} else {
    result = apply(localState);
}
return result;
```

or something

ottobackwards:
Try on the overall watch would have to behave differently, as it would both close the watch and stop the root timing...

garydgregory:
I hope we do not advocate running code like the fragment you just quoted. This is the same kind of issue we finally resolved in Apache Log4 2 where you no longer need to write code like:

```
if (logger.isDebugEnabled()) {
   logger.debug("This is " + foo + " and " +bar);
}
```

Instead you say:

`logger.debug("This is {} and {}", foo, bar);`

You can also delegate evaluation of expensive calls only if the logger is enabled:

`logger.debug("This is {} and {}", () -> foo(), () -> bar());`

The logger is configured from a file or another code snippet.

IMO, we are already dangerously close to out of bounds of [lang] with the current code, so I am mentioning this for exploration ATM. That said, I think it would be good to see in text at least what a fully featured gadget looks like before we include the current version in [lang]. Once in, it will likely evolve, and then it might end up being moved to another component if it is no longer 'lang-y'...

garydgregory:
WRT "Try on the overall watch would have to behave differently, as it would both close the watch and stop the root timing..." 

I am not saying you would always use a try-with-resources, just that it should be an option.

ottobackwards:
Sorry @garydgregory I understood you to mean it would be optional.

So, the context of the creation of this idea is I did it for an environment where we have optional available 'capabilities', and it is common to check and use if present or fail if required but not present.

There is a difference between the logging example and this.  I am familiar with logging quick fail ( I implemented my own c++ streams based quick failing logger back in the day even ), but the logger is always there, you are checking for configuration or optionally enabled functionality ( debug logging ), where as I wrote this for an environment where we may not have a watch at all.

I am not saying having a 'singleton' watch, with a 'do nothing' set of default interface implementations isn't something to think about.

But things start from where you have a need usually.

ottobackwards:
To do try with resources we would have to have do nothing default returns right?

ottobackwards:
Maybe there should be a commons-examples project with the src for examples?  The java doc examples could come from there

ottobackwards:
the static factory, is pretty straight forward.
The try with resources, less so.  What ends up happening is you cannot stop the watch before visiting the nodes.

For example, this is what you want to do if you are going to time, but visit later:

```java
 @Test
    public void testTryWithResources() {
        final StopWatch stopWatch = new StopWatch();
        
        try(StackWatch<String,String> watch = new StackWatch<>("testStackWatch")) {
            watch.start();
            stopWatch.start();
            stopWatch.stop();
        }
        watch.visit(new StackWatch.TimingVisitor<String,String>() {
            @Override
            public void visitTiming(int level, List<String> path, StackWatch.Timing<String,String> node) {
                assertTrue(node.getStopWatch().getNanoTime() > stopWatch.getNanoTime());
            }
        });
    }
```

but it won't work/compile with try with resources

ottobackwards:
same with the static factory:

```java
 @Test
    public void testTryWithResourcesStaticFactory() {
        final StopWatch stopWatch = new StopWatch();

        try(StackWatch<String,String> watch = StackWatch.startStackWatch("testStackWatch")) {
            stopWatch.start();
            stopWatch.stop();
        }
        watch.visit(new StackWatch.TimingVisitor<String,String>() {
            @Override
            public void visitTiming(int level, List<String> path, StackWatch.Timing<String,String> node) {
                assertTrue(node.getStopWatch().getNanoTime() > stopWatch.getNanoTime());
            }
        });
    }
```

ottobackwards:
@garydgregory @kinow I added the static method.  I did not squash/rebase though. I will if you need me to but I want the review.

garydgregory:
When I see code like
```
watch.visit(new StackWatch.TimingVisitor<String,String>() {
            @Override
            public void visitTiming(int level, List<String> path, StackWatch.Timing<String,String> node) {
                assertTrue(node.getStopWatch().getNanoTime() > stopWatch.getNanoTime());
            }
        });
```
I have my doubts as to this belonging in Commons Lang and not a new module like a Commons Time (but nothing with Joda-Time or java.time overlap) or Commons Perf.
This feels very 'framework-y' to me and beyond the otherwise simple Commons Lang APIs.

Thoughts from anyone else?

ottobackwards:
@garydgregory, since we are getting out of code review / pr and into bigger issues, can we move your comments to jira?  Gilles doesn't have github and the conversation will make more sense if we are all there.

ottobackwards:
giving up:  https://github.com/palindromicity/stackwatch

PR #313, opened 2018-02-10
LANG-1060: NumberUtils.isNumber assumes number starting with Zero

piotrkosyoo:
NumberUtils.isCreatable("002.345"); return false


PascalSchumacher:
Thanks! 👍

piotrkosyoo:
No problem Sir,
Calling Number.to Float also works correctly, and I skip this point.  but maybe
 better is add the tests for this case?

PascalSchumacher:
A pull request with tests for the `float` case would be welcome.

piotrkosyoo:
ok in this same issue number ?

PascalSchumacher:
Sure, that is fine!

PR #314, opened 2018-02-11
LANG-1060 NumberUtils.isNumber assumes number starting with Zero is octal

piotrkosyoo:
Add test case


PascalSchumacher:
Thanks! 👍

PR #315, opened 2018-02-22
Get boolean value if the size of array is between two integer values

marcosstefani:
/**
     * Get boolean value if the size of array is between two integer values
     *
     * @param array  the array that has the number of records to compare
     * @param first  initial value to be compared
     * @param last   final value to be compared
     * @return boolean value that will tell if the result is true or false
     * @author marcos.cantor@gmail.com
 */


marcosstefani:
I dont understend this error.

PR #318, opened 2018-03-01
clean code

JiangYongKang:
Implement the existing functionality with fewer code.


JiangYongKang:
@stokito thinks

kinow:
@JiangYongKang, thanks for your contribution.

Even though the behaviour is the same, in my opinion, the other code is easier to read, and fewer code doesn't necessarily mean better or more clean code.

Checking if a number if even or odd, for example, can be done with bit operation, or with a `if/else`. While the former is shorter, not all developers immediately understand it, nor know how to maintain/change it.

PR #319, opened 2018-03-16
Contribution of Span implementations for LANG-1387

Claudenw:
Implementation of LANG-1387

Added underflow and overflow checks ot NumberUtils.
Added IntSpan and LongSpan classes.


Claudenw:
Undetected errors in build.  Will fix and resubmit

PR #323, opened 2018-04-06
Add comparison methods to ObjectUtils

aaoferreira:
Additional comparison methods to make `greater than`, `greater than or equal to`, `less then` and `less than or equal to` easier to use instead of checking the int outcome of `compare` everywhere and making the code that utilises it more readable

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -591,6 +591,58 @@ public class ObjectUtils {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @aaoferreira,

This PR should be superseded by the new `org.apache.commons.lang3.compare.ComparableUtils`
May you please check this new functionality to make sure it satifies your use cases?

Thank you,
Gary

aaoferreira:
I had forgotten about this one! haha thanks for letting me know @garydgregory , it does satisfy my use cases indeed. I'll close this PR.

PR #325, opened 2018-04-19
LANG-1392 methods for getting first non empty or non blank value

jeffnelson:


garydgregory:
Thank you for your PR. It feels to me like this new API should be built on top of a more generate "getting Nth non empty or non blank value"

jeffnelson:
@garydgregory - if you want me to refactor as such, I can. But I modeled the implementations of these 2 new methods after the existing `ObjectUtils.firstNonNull(Object T ... values)` https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/ObjectUtils.java#L125-L134

What I've submitted so far seemed like an obvious implementation choice since it is consistent with other similar existing code.

garydgregory:
Your proposal is fine, my suggestion currently falls in the YAGNI category.

PascalSchumacher:
Thanks! 👍

PR #327, opened 2018-04-28
(doc) Remove invalid example of `lastIndexOf` from javadoc

fedochet:
`StringUtils.lastIndexOf("aabaabaa", "ba", 2)  = -1` is invalid; moreover, example below is applied to the same arguments, but gets another (correct) result.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1822,7 +1822,6 @@ public class StringUtils {
      * StringUtils.lastIndexOf("aabaabaa", "b", 0)  = -1
      * StringUtils.lastIndexOf("aabaabaa", "b", 1)  = -1
      * StringUtils.lastIndexOf("aabaabaa", "b", 2)  = 2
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Thanks! 👍

printfcoder:
`StringUtils.lastIndexOf("aabaabaa", "b", 1)  = -1`

Could you please explain why the result is `1`,please? I see it shoud be 2, because that `to search starts at 1 position and b is at 2 idx`.

fedochet:
@printfcoder according to [docs](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html), 
```
(lastIndexOf) Returns the index within seq of the last occurrence 
of the specified character, searching backward starting 
at the specified index. 
```
So `b` will be searched in range of positions `[0, 1]`, and there's no `b`s in there

printfcoder:
@fedochet Oh ! base on your explaining, it is ok. But the param `startPos ` means `start`, I’m totally confused, I think it should be renamed as `endPos`, aHa ?.

PR #332, opened 2018-06-02
Add methods allowing masking of Strings

greenman18523:
maskStart and maskEnd allow masking the original str by replacing it's
characters with the specified character-mask.
Common usecase is to hide sensitive information from logs, by using it
in toString of classes or in inputs to log calls.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -9355,7 +9355,7 @@ public class StringUtils {
      *
      * <pre>
      * StringUtils.maskStart(null, *, *, *)       =  null
-     * StringUtils.maskStart("", *, *, *)         =  null
+     * StringUtils.maskStart("", *, *, *)         =  ""
      * StringUtils.maskStart("test", *, -1, 4)    =  "test"
      * StringUtils.maskStart("test", *, 0, 4)     =  "test"
      * StringUtils.maskStart("test", 'X', 1, 4)   =  "Xest"
@@ -9364,6 +9364,8 @@ public class StringUtils {
      * StringUtils.maskStart("test", 'X', 1, 0)   =  "XXXX"
      * </pre>
      *
+     * <p>This is equivalent to calling {@code mask(str, mask, minMasked, 0, maxUnmasked)}</p>
+     *
      * @param str the String to mask it's content
      * @param mask the character used as the mask
      * @param minMasked the minimum number of characters to mask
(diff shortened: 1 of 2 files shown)


stokito:
Hi @greenman18523 thank for your contribution - from my experience this is "must have" functionality for Commons Lang library because in almost all bit projects what I saw was their home grown masking function.

But what I would like to propose is to simplify the api and use one function `mask()` instead of two `maskStart()` and `maskEnd()`. This will simplify code of the functions especially both of them are quite similar. I created a pull request with my alternative implementation #335
Also I created a JIRA ticket https://issues.apache.org/jira/browse/LANG-1400 

Please give your feedback,
Thank you

greenman18523:
Hello @stokito, thanks for the input. The reasoning of creating two method was for a more intuitive API and to not add too many parameters in just one method. And I also didn't want to create a 3rd method that would just do the main logic without it being able to be exposed (i.e. private).

I like the idea of the mask being in the middle of the string, and I was thinking of incorporating it into the code, but was not sure how useful it would be for others.

An other goal of the way the methods are, is to clearly enforce a minimum amount of masking before being able to show any information. e.g. in the case of a credit card, even if the user has erroneously entered 8 out of the 16 characters, a call to ``StringUtils.maskEnd("35660020", '*', 12, 4)`` would not show any sensitive information, but a call to ``StringUtils.mask("35660020", 0, 4, '*')`` would expose. 

I think that both implementations can be useful for a number of use-cases, some common to both, but some are covered by one of the two implementation.

Of course this can also be achieved by an extra parameter in your implementation. What do you think?

greenman18523:
New method added, to cover usacases of https://issues.apache.org/jira/browse/LANG-1400

sebastian-nagel:
> "must have" ... because in almost all bit projects what I saw was their home grown masking function.

Well, and I had to add another one to address [NUTCH-2905](https://issues.apache.org/jira/browse/NUTCH-2905) in apache/nutch#704. We had one additional use case: mask only a part of string matched by a regular expression.

garydgregory:
Masking might belong in Commons Text, not here in Commons Lang IMO.

stokito:
As a basic secure functionality it must be in core Java library or better even as a method of `String`. Security functions are "batteries" that must be included.
Almost all projects that I saw had commons-lang  as a dependency but the commons-text I can't even remember.

garydgregory:
> As a basic secure functionality it must be in core Java library or better even as a method of `String`. Security functions are "batteries" that must be included. Almost all projects that I saw had commons-lang as a dependency but the commons-text I can't even remember.

There is no point arguing here what belongs in what JRE class ;-) 
We have been moving text based code to Commons Text for some time now while deprecating the original code in Commons Lang. Therefore, I don't see us adding more text based utilities to Lang. Masking is definitely a text concept IMO, as opposed to java.lang level processing.

garydgregory:
Closing: No action in 3 years. See reference to Commons Text. Also Java has stock support for replace methods and regular expressions.

PR #336, opened 2018-07-07
LANG-1402: Null/index safe get methods for ArrayUtils

MarkDacek:
@chtompki 
If this is deemed worthy or needs improvement - let me know.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -8676,7 +8676,7 @@ public class ArrayUtils {
     /**
      * Returns whether a given array can safely be accessed at the given index.
      * @param <T> the component type of the array
-     * @param array the array to inspect
+     * @param array the array to inspect, may be null
      * @param index the index of the array to be inspected
      * @return Whether the given index is safely-accessible in the given array
      * @since 3.8
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -8675,7 +8675,7 @@ public class ArrayUtils {
 
     /**
      * Gets an element from the array if the array is non-null and appropriately long, otherwise returns null
-     * @param <T> the component type of the array
+     * @param <T> the component type of the array, may be null
      * @param array   the array holding the desired element
      * @param index  the index of the element in the array
      * @return The element in the array at the index, or null if it is ill-formatted
@@ -8688,10 +8688,10 @@ public class ArrayUtils {
     /**
      * Gets an element from the array if the array is non-null and appropriately long, otherwise returns the specified value
      * @param <T> the component type of the array
-     * @param array   the array holding the desired element
+     * @param array   the array holding the desired element, may be null
      * @param index  the index of the element in the array
      * @param defaultReturn the object to be returned if the array is null or shorter than the index
-     * @return The element in the array at the specified index, or the given argument if it is ill-formatted
+     * @return The element in the array at the specified index, or the given argument if it the array is not sufficiently long for the index. May return null if the array contains null
      * @since 3.8
      */
     public static <T> T get(T[] array, int index, T defaultReturn){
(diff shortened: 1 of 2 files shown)


sebbASF:
User code has to cast the return to make use of it; that is not ideal.

MarkDacek:
I added your suggested method and refactored the other to use generics.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8672,4 +8672,53 @@ public static void shuffle(final double[] array, final Random random) {
...
+     * @return The element in the array at the specified index, or the given argument if it is ill-formatted
+     * @since 3.8
+     */
+    public static <T> T get(T[] array, int index, T defaultReturn){
+        if(getLength(array) == 0 || array.length <= index){
+            return defaultReturn;
+        }
+
+        if(index < 0 ){
+            index = 0;
```
sebbASF: Surely a negative index should return the default?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8672,4 +8672,53 @@ public static void shuffle(final double[] array, final Random random) {
...
+        }
+
+        return array[index];
+    }
+
+    /**
+     * Gets an element from the array if the array is non-null and appropriately long, otherwise returns the specified value
```
sebbASF: Does not agree with code

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8672,4 +8672,53 @@ public static void shuffle(final double[] array, final Random random) {
...
+     * @param <T> the component type of the array
+     * @param array   the array holding the desired element
```
sebbASF: May be null

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8672,4 +8672,53 @@ public static void shuffle(final double[] array, final Random random) {
             swap(array, i - 1, random.nextInt(i), 1);
         }
     }
+
+    /**
+     * Gets an element from the array if the array is non-null and appropriately long, otherwise returns null
+     * @param <T> the component type of the array
+     * @param array   the array holding the desired element
+     * @param index  the index of the element in the array
+     * @return The element in the array at the index, or null if it is ill-formatted
```
sebbASF: What does ill-formatted mean?
Can also return null if the array element is present and null.

I don't think this method is sufficiently useful to be worth adding

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8672,4 +8672,53 @@ public static void shuffle(final double[] array, final Random random) {
...
+        return get(array, index, null);
+    }
+
+    /**
+     * Gets an element from the array if the array is non-null and appropriately long, otherwise returns the specified value
+     * @param <T> the component type of the array
+     * @param array   the array holding the desired element
+     * @param index  the index of the element in the array
+     * @param defaultReturn the object to be returned if the array is null or shorter than the index
+     * @return The element in the array at the specified index, or the given argument if it is ill-formatted
```
sebbASF: ill-formatted?

Again, I don't see the point of this method.
It's a bit more flexible than the original method.
But it's still only useful for the case where you can choose a default value that does not appear in the array, and you still have to check whether the return value is the default or not.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
MarkDacek: Sorry, mistaken copy-paste there. Feel free to use this as a tutorial for how NOT to contribute.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
MarkDacek: Fixed.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
MarkDacek: My thought process was something similar to this https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringUtils.java#L7456-L7458

I would understand if Strings are understood well-enough to merit a true default value and generic arrays aren't.

MarkDacek:
@sebbASF Thanks for your patience and efforts to make this a worthy contribution. I have removed the other methods per your suggestion.

PR #339, opened 2018-08-02
[LANG-1354] Added Methods to FieldUtils to Return Non-Synthetic Fields From a Class

mdbuck:
I created two new methods as I felt that modifying the existing methods would break backward compatibility.


mdbuck:
shrug

PR #340, opened 2018-08-08
[LANG-1406]  StringIndexOutOfBoundsException in StringUtils.replaceIgnoreCase

HiuKwok:
Fix for Lang-1406 to avoid any exception while performing String.UTils.replaceIgnoreCase() against uniCode String object. 

Plz let me if there have any extra things need to be done for this PR since I am the first timer for commons-lang project( add more test case?).

All the best


kinow:
I'm surprised by this bug. Had no idea something like this could happen. Will debug later and see if I can understand why that happens (might have to train my brain to default to always use uppercase instead of lowercase?). Thanks for the pull request, we will review the code and if everything looks OK a committer will merge it.

kinow:
Oh, that does make sense now. So the first visible character we see is the ["Latin Capital Letter I with Dot Above"](https://unicode-table.com/en/#0130) (see also [this other link](https://en.wikipedia.org/wiki/Dotted_and_dotless_I)), and the second an `x`. And doing `toUpperCase()` simply won't change it as it's considered already upper case.

When doing a `toLowerCase`, it gets translated into two visible characters. The second is the normal `x`. While the first contains two codepoints. I tested in Python, and got the lower case `i` (`print(u"\u0069")`) followed by a character invisible by itself (`print (u"\u0307")`).

The special/invisible character, is visible when coming after certain letters.

```python
>>> print(u"\u0307")

>>> print(u"\u0069\u0307")
i̇
>>> print(u"\u0068\u0307")
ḣ
>>> print(u"\u0067\u0307")
ġ
>>> print(u"\u0067\u0307")
```

When we get these invisible characters, as we have one code point more, the length returned is not 2, but 3. Resulting in exception in this issue.

I don't believe the fix here would fix the reverse case, where we had a lower case, single codepoint, unicode; that would be represented by a two code codepoint. The exception could happen again (I haven't investigated whether such case exist, but I'm assuming there could be such case - if not now, maybe a character could still be added in future editions).

What do you think @HiuKwok ? Any suggestions? I'm not sure if there's any easy way to fix this case, except by adding a note to the documentation saying that the method is not intended to be used with unicode strings, as it doesn't handle supplementary characters well. Or maybe we could try to remove the `length()` call around the `StringBuilder`'s near the end of the method...

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -5596,8 +5596,8 @@ private static String replace(final String text, String searchString, final Stri
          }
          String searchText = text;
          if (ignoreCase) {
-             searchText = text.toLowerCase();
-             searchString = searchString.toLowerCase();
+             searchText = text.toUpperCase();
+             searchString = searchString.toUpperCase();
```
kinow: Just leaving a comment here too to have a review here in GitHub. While your example works, as the character is considered already in upper case, the reverse case would still fail after changing from `toLowerCase` to `toUpperCase`. So I think we should find another solution or update the documentation stating how the code works with unicode.

kinow: Oohh, great testing @HiuKwok ! Thanks for sharing here.

>Just a quick question what you mean by remove the length( ) mean? Would you mind to specify more on that?

Sure. I think there could be a possibility to fix the issue by addressing how the length of the lower'ed/upper'ed text is used https://github.com/apache/commons-lang/blob/590f90889bf61a5570bd98b78e73410a07d7410b/src/main/java/org/apache/commons/lang3/StringUtils.java#L5603

So maybe there could be another way to work around the way we use the strings lengths, and avoid the exception.

HiuKwok: @kinow May be try to store a original text.length before performance any upper || lower case operation and make use of it?

haha, it's just a thought. Would try out later

HiuKwok: @kinow It turn out the problem is large than that, because the incorrect length basically affecting the whole method, especially the part which involve parsing inside the while loop.

https://github.com/apache/commons-lang/blob/590f90889bf61a5570bd98b78e73410a07d7410b/src/main/java/org/apache/commons/lang3/StringUtils.java#L5612-L5619

While string like following is entered, then Exception would be thrown on here instead.
https://github.com/apache/commons-lang/blob/590f90889bf61a5570bd98b78e73410a07d7410b/src/main/java/org/apache/commons/lang3/StringUtils.java#L5613

image

kinow: Yeah, a very slippery problem. We still have the option to simply document that the method does not work well with unicode strings. 

But first I would like to spend at least a few hours with paper and pencil (and eraser, because this could take a bit till I give up or find a possible way around it), and perhaps even check in the mailing list if other devs have any idea.

I think you found a very interesting problem (*)! Keep the ideas coming if you have any on how to solve this issue :+1: 

_* if I had more time, I would possibly either go through other methods checking for that or, just try some fuzzifier approach to test the whole project ! Not aware of any static or dynamic analysis tool that does that_

HiuKwok: Hi all, Just a quick update finally I kind of figure the way out of this problem,
while I am trying to make it happen (the implementation).
But the concept is mainly Normalize incoming word at the beginning of the method call by ```Normalizer.normalize()```

HiuKwok:
To whom who interested in this issue, here is some founding that I discovered throughout this month of issue solving. 

Problem:
 - The exception would happened when any String object passed in with unicode character. In order to achieve ignore case replacement, the internal logic would first transform both `text` and `SearchString` to lowerCase( ) for comparaition.   

- However if anyone passion enough to digger deeper into the src logic of `.toLowerCase( )`. Certain unicode character would be denormalized. In this way the result String length would tend to longer than original length().  Example like:  image
So making use of the transformed String, Out bound exception would happen when trying to access the index that doesn't access at all (3 in this case vs 2 in length before lowerCase).

Flow:

 - So the first thought into my mind is, why dun just normalize both `text` and `searchString` before performing ignore case comparation? In this way the String length would always stay consistence no matter `toLowerCase( )` or `toUpperCase( )` 3 -> 3.  However the another problem would emerged, as you may noticed, while the String mentioned above denormalize, it would turn into a UpperCase I and a dot sign. 

- But what happen if the search pattern emerge into searchText in decompose form. In this case let say I am trying to match a upper [I]. Then mismatch would happen and this is certain not the desire behavior of this method I believe. 

BTW I Drafted a simple main method to demonstrate how mismatch would happen in here.

https://github.com/HiuKwok/commons-lang/blob/master/src/main/java/com/hiukwok/test.java#L10-L20

kinow:
Good progress so far @HiuKwok . Understanding the problem well is a good first step to solve it 👍 my next development cycle for Apache will probably be for a release for Apache Commons Imaging, then after that Lang+Text. So will try to help here if you haven't found a way to solve it yet.

kinow:
Oh, actually the issue has been closed in #423! Closing the PR, but thanks for your contribution @HiuKwok  !

PR #342, opened 2018-08-17
LANG-1411: Add empty checks to ObjectUtils

nictas:


kinow:
Hi @nictas , thanks for your contribution. The code looks really good, well commented and formatted with unit tests. :clap: :clap: 

And you seem to have a valid use case for it. I think I could even get used to calling this method in some of my projects.

Leaving my +1 here, but leaving it for others to review. Would be especially good someone with more time to confirm we don't have anything similar, or that could be chained, etc, I think. 

Thank you again
Bruno

PR #344, opened 2018-08-20
Adding method to remove special characters

Callegas:
After having always experienced the need to remove **special characters** in some projects, especially when working with other **languages** that use a lot of **accented words**, I found it interesting that String Utils provided us with this method to avoid creating it in different places and projects.


fredericohubler:
I also experience problems with special characters in my day-to-day work. Supported! +1

PR #345, opened 2018-08-23
add jvmLaunchers

harbby:
When you need to perform some operations through a child process, it is very cumbersome and requires a lot of coding. I have an api below to simplify this process and make it elegant

Like this: The Callable event is executed by the child process (jvm)

```java
        JVMLauncher<Integer> launcher = JVMLaunchers.<Integer>newJvm()
                .setCallable(() -> {
                    System.out.println("************ Compile start ***************");
                    TimeUnit.SECONDS.sleep(1);
                    System.out.println("************ Compile stop ***************");
                    return 1;
                })
                .addUserjars(Collections.emptyList())
                .setConsole((msg) -> System.err.println(msg))
                .build();

        VmFuture<Integer> out = launcher.startAndGet();
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/vm/Consumer.java
+++ b/src/main/java/org/apache/commons/lang3/vm/Consumer.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


sebbASF:
I think this is getting out of scope for LANG, given that there is already the EXEC component.

harbby:
@sebbASF  I think exec focuses on executing the operating system child process (shell start).
And my api, focusing on the task collaboration with the two jvm,
For example, the parent process hands a temporary compilation job to the child process (jvm) to execute, and waits for it to return the compilation result.

sebbASF:
AFAICT the code starts the JVM using an OS process.
It also makes assumptions about the classpath separator character that may not apply to all OSes.

There may be other such assumptions. EXEC should already deal with all of these.

I still think this belongs in EXEC rather than LANG.

It might make a useful enhancement to EXEC?

harbby:
Well, what you said is right, it really should be returned to the exec domain.
This api wants to quickly pass the lambda to the child process. In order to simplify the cues of exec, it should be very useful in many scenarios.

harbby:
Hi all, After a long time, we practiced in many projects, we found pr very interesting, I plan to close this discussion, discuss in commons-exec

PR #349, opened 2018-08-28
LANG-1417: Deprecate ThreadPredicate and ThreadGroupPredicate in favo…

britter:
…r of java.util.function.Predicate

I don't see a way to let ThreadGroupPredicate and ThreadPredicate inherit from java.util.function.Predicate. See my comment in [LANG-1417](https://issues.apache.org/jira/browse/LANG-1417). @salyh as author of the initial implementation, can you have a look please? Am I missing something?

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -46,6 +46,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
 
   <release version="3.9" date="????-??-??" description="??">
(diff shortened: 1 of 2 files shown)


britter:
In this proposal I added the java.util.function.Predicate APIs to the ThreadUtils class. Another option is to deprecate ThreadUtils all together and put a new ThreadUtils class alongside which only hat the new APIs.

britter:
Build is failing on Travis but not locally... 🤔

britter:
We agreed to just add `@FunctionalInterface` to `ThreadPredicate` and `ThreadGroupPredicate`

PR #354, opened 2018-09-05
Convert tests for Validate.isTrue overloads to @Nested test

britter:
Proposal for a better structure of tests using `@Nested`. Each method should have it's own `@Nested` container which is called like the method. Inside that container each overload has it's own `@Nested` container. I did this for `Validate.isTrue` to show the approach.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ValidateTest.java
+++ b/src/test/java/org/apache/commons/lang3/ValidateTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.commons.lang3;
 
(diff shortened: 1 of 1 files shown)


britter:
@kinow @PascalSchumacher @chtompki WDYT?

PascalSchumacher:
+1

britter:
Okay, as there is positive feedback for this change, I'm going to migrate the whole `ValidateTest` to this style and then merge it 👍

britter:
Build failure seems to be unrelated.

PR #355, opened 2018-09-05
Use @ParameterizedTest to iterate over available locales

britter:


britter:
I'm going to reach out to the JUnit maintainers to find out whether we can even inject zone values via parameter source.

PR #357, opened 2018-09-21
Proposal for LANG-1421

sparsick:
- replace test annotation and assertions in ObjectUtilsTest with junit5 
- implementation of `allNull` and `anyNull`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -236,6 +236,31 @@ public class ObjectUtils {
         return firstNonNull(values) != null;
     }
 
+    /**
+     * Checks if any value in the given array is {@code null}.
+     *
+     * <p>
+     * If none of the values are {@code null} or the array is {@code null}
+     * or empty then {@code false} is returned. Otherwise {@code true} is returned.
+     * </p>
+     *
+     * <pre>
+     * ObjectUtils.anyNull(*)                   = false
+     * ObjectUtils.anyNull(*, null)             = true
+     * ObjectUtils.anyNull(null, *)             = true
+     * ObjectUtils.anyNull(null, null, *, *)    = true
+     * ObjectUtils.anyNull(null)                = true
+     * ObjectUtils.anyNull(null, null)          = true
+     * ObjectUtils.anyNull()                    = false
+     * ObjectUtils.anyNull((Object[]) null)     = false
+     * </pre>
+     *
+     * @param values  the values to test, may be {@code null} or empty
+     * @return {@code true} if there is at least one null value in the array,
+     * {@code false} if all values in the array are not {@code null}s.
+     * If the array is {@code null} or empty {@code false} is also returned.
+     * @since 3.9
+     */
     public static boolean anyNull(final Object... values) {
         if (values == null || values.length == 0) {
             return false;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -236,6 +236,39 @@ public static boolean anyNotNull(final Object... values) {
...
+     * ObjectUtils.anyNull(null)                = true
+     * ObjectUtils.anyNull(null, null)          = true
+     * ObjectUtils.anyNull()                    = false
+     * ObjectUtils.anyNull((Object[]) null)     = false
+     * </pre>
+     *
+     * @param values  the values to test, may be {@code null} or empty
+     * @return {@code true} if there is at least one null value in the array,
+     * {@code false} if all values in the array are not {@code null}s.
+     * If the array is {@code null} or empty {@code fatolse} is also returned.
```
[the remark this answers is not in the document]
andyklimczak: depending on how you feel, you could replace the `if the array is {@code null}` info with something about this method being null safe (to keep the wordage the same as other methods that use `null safe`)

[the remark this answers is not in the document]
sparsick: Thank you for reviewing. I don't find a better replacement for the if sentence. I hope it is ok.

src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
```diff
@@ -148,11 +153,27 @@ public void testAnyNotNull() {
...
+     */
+    @Test
+    void testAnyNull() {
+        assertTrue(ObjectUtils.anyNull((Object) null));
+        assertTrue(ObjectUtils.anyNull(null, null, null));
+
+        assertFalse(ObjectUtils.anyNull());
+        assertFalse(ObjectUtils.anyNull((Object[]) null));
+        assertFalse(ObjectUtils.anyNull(FOO));
+        assertFalse(ObjectUtils.anyNull(null, FOO, null));
```
aaabramov: Should this really be `assertFalse`?
`ObjectUtils.anyNull(null, null, null, null, FOO, BAR)` - should return `true`, because it has `null` elements in varargs.

sparsick: Thank you. You are right. I will create a new commit.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -275,6 +315,45 @@ public static boolean allNotNull(final Object... values) {
...
+     * @return {@code false} if there is at least one non {@code null} value in the array, the array is {@code null}
+     * or array contains no elements, {@code true} if all values in the array are {@code null}s .
+     * @since 3.9
+     */
+    public static boolean allNull(final Object... values) {
+        if (values == null || values.length == 0) {
+            return false;
+        }
+
+        for (final Object val : values) {
```
dekelpilli: This can be replaced by `return firstNonNull(values) == null`, although then you're doubling up on the null check.

sparsick:
I closed this PR because the method `anyNull` was already implemented and merged in commit https://github.com/apache/commons-lang/commit/ee6c682ed5fa8c1b33aa1d52a1e2f088b0191b6b

PR #358, opened 2018-09-28
ClassUtils.getBaseClasses(desiredBase, packageName)

zmacomber:
Hello,

I have been using a method in my company's (avadasoftware.com) code base that I created for getting a list of classes that implement an interface. These classes reside in a particular package that I query.  I use this method in my code base in the following manner:

List<FooInterface> fooClasses = ClassUtils.getBaseClasses(FooInterface.class, "com.avada.foo");

I would like to put this in Apache's ClassUtils to move this generic code out of our code base and into what I believe is a more appropriate place.  I'm striving to replace util/generic code in our code base with 3rd party libs as I believe it will be better tested/supported/enhanced/etc... in the OSS world.

I have written several tests that are passing to test exceptional behavior and also normal path behavior.


zmacomber:
Just checked the Travis CI build and it failed because of an import of java.lang.IllegalArgumentException at the top which is not necessary...

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1066,69 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+
+        if (desiredBase == null) {
+            throw new IllegalArgumentException("desiredBase must not be null");
+        }
+
+        if (StringUtils.isBlank(packageName)) {
+            throw new IllegalArgumentException("packageName must not be blank");
+        }
+
+        if (packageName.contains("/")) {
```
sudeendra-nadager: might be better to validate against a regex.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1066,69 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+        List<T> classes = new ArrayList<>();
+
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(classesPath)) {
+            for (Path file: stream) {
+                Path pathFileName = file.getFileName();
+                if (( ! Files.isDirectory(file)) && (pathFileName != null)) {
+                    String fullClassName = packageName + "." +
+                                           pathFileName.toString().replace(".class", "");
+
+                    // Only add classes that can be instantiated via newInstance()
```
sudeendra-nadager: what if the derived class does not provide default ctor?

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1066,69 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+                    try {
+                        Object obj = Class.forName(fullClassName).newInstance();
+                        if (desiredBase.isInstance(obj)) {
```
sudeendra-nadager: IMO - use of instanceof operator is preferred over method call.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -1202,6 +1206,56 @@ public void testGetInnerClass() throws ClassNotFoundException {
         assertEquals( Inner.DeeplyNested.class, ClassUtils.getClass( "org.apache.commons.lang3.ClassUtilsTest$Inner.DeeplyNested" ) );
     }
 
+    @Test
+    public void testGetBaseClassesExtends() throws Exception {
+        List<AnotherParent> classes = ClassUtils.getBaseClasses(AnotherParent.class, "org.apache.commons.lang3.reflect.testbed");
+        assertTrue(classes.size() > 0);
```
sudeendra-nadager: should validate each element of the container for better accuracy.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -1202,6 +1206,56 @@ public void testGetInnerClass() throws ClassNotFoundException {
...
+    }
+
+    @Test
+    public void testGetBaseClassesAbstract() throws Exception {
+        List<AbstractConcurrentInitializerTest> classes = ClassUtils.getBaseClasses(AbstractConcurrentInitializerTest.class, "org.apache.commons.lang3.concurrent");
+        assertTrue(classes.size() > 0);
```
sudeendra-nadager: should validate each element of the container for better accuracy.

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: The JavaDoc explicitly states that only classes that can be instantiated via a default ctor can be used:

"This method only retrieves base classes/interfaces that have children that can be instantiated via a no-args constructor"

If the class can't be instantiated, an Exception is thrown and it won't be added to the list of returned classes:
```
catch (Exception e) {
    // Class was not instantiable via newInstance()
}
```

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: instanceof and isInstance() are the same.  You use instanceof when you know the actual class type which, in this case, is unknown.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1066,69 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+     * @param <T> The desired base class or interface type to retrieve
+     * @return a list of base classes/interfaces that match the supplied type underneath the supplied package
+     * @throws URISyntaxException if the packageName is not found in the class loader
+     * @throws IOException if an I/O error occurs in getting a new directory stream
+     * @throws IllegalArgumentException if the desiredBase or packageName are invalid
+     */
+    public static <T> List<T> getBaseClasses(final Class<T> desiredBase, final String packageName)
+            throws URISyntaxException, IOException, IllegalArgumentException {
+
+        if (desiredBase == null) {
```
grimreaper: Prefer Objects.requireNonNull instead of explicit `if` checks.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
zmacomber: This is a good idea

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
zmacomber: This is a good idea

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: Will do

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: Will do

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: On 2nd thoughts, think I'll just take out "packageName.contains('/')".  The packageName will be checked below that call.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1067,62 @@ public static boolean isInnerClass(final Class<?> cls) {
         return getClass(loader, className, initialize);
     }
 
+    /**
+     * Returns a list of base classes/interfaces underneath the supplied package
```
aaabramov: I believe it would be better with dots in the end of sentence.

aaabramov: I believe it would be better with dots in the end of sentence.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1067,62 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+     * This method only retrieves base classes/interfaces that have children that can be instantiated
+     * via a no-args constructor
```
aaabramov: This line could be improved. 
At least, interfaces has no constructors.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1067,62 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(classesPath)) {
+            for (Path file: stream) {
+                Path pathFileName = file.getFileName();
+                if (( ! Files.isDirectory(file)) && (pathFileName != null)) {
+                    String fullClassName = packageName + "." +
+                                           pathFileName.toString().replace(".class", "");
+
+                    // Only add classes that can be instantiated via newInstance()
+                    try {
+                        Object obj = Class.forName(fullClassName).newInstance();
```
aaabramov: Why do we need to call `newInstance()` on `Class<?>` if we return just `List<Class<?>>`?

aaabramov:
Should we overload `getBaseClasses` with the following:
- `ClassUtils.getBaseClasses(final Class<T> desiredBase, final String packageName, ClassLoader classLoader)`
- `ClassUtils.getBaseClasses(final Class<T> desiredBase, final Package package)`
?

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: To confirm that the class can actually be instantiated via the default constructor.  Notice the Exception comment and the comment above "try {"

zmacomber:
> Should we overload `getBaseClasses` with the following:
> 
> * `ClassUtils.getBaseClasses(final Class<T> desiredBase, final String packageName, ClassLoader classLoader)`
> * `ClassUtils.getBaseClasses(final Class<T> desiredBase, final Package package)`
>   ?

I have implemented this

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: I have committed changes for this

src/main/java/org/apache/commons/lang3/ClassUtils.java
zmacomber: I have committed changes for this

zmacomber: I have committed changes for this

zmacomber:
Have not heard anything on this PR for 2 months.  Should I just close this PR?  This doesn't seem like the way open source should work with such a long time for feedback.  I'm fine with feedback like, "This is not going to fit in our codebase".

src/main/java/org/apache/commons/lang3/ClassUtils.java
ecki: I dont think that is a good idea to instantiate all classes, this makes the utility rather complicated to use.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1059,6 +1067,85 @@ public static boolean isInnerClass(final Class<?> cls) {
...
+        if (StringUtils.isBlank(packageName)) {
+            throw new IllegalArgumentException("packageName must not be blank");
+        }
+
+        Objects.requireNonNull(classLoader, "classLoader must not be null");
+
+        URL url = classLoader.getResource(packageName.replaceAll("[.]", "/"));
+        Objects.requireNonNull(url, "supplied package not found");
+
+        Path classesPath = Paths.get(url.toURI());
```
ecki: Will that work with jmod or jar files or only with class directories?

zmacomber: I'm not sure if it will work with jmod or jar files.  I designed the method to just work with class directories.

PR #363, opened 2018-10-02
Update builder tests to JUnit Jupiter

mureinik:
Upgrade the tests in the `builder` package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

While most of these changes are drop-in replacements with no functional benefit, it is worth mentioning the change to how expected exceptions are tested.
Unlike `org.junit.Test`, `org.junit.jupiter.api.Test` does not have an `expected` argument. Instead, an explicit call to `org.junit.jupiter.api.Assertions.assertThrows` is used. This call allows the test to pinpoint the exact statement that is expected to throw the exception and allows making the tests a bit stricter by preventing
false-positives that could occur if the setup code would throw the expected exception instead of the statement that was supposed to throw it.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/builder/CompareToBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/CompareToBuilderTest.java
@@ -16,12 +16,13 @@
  */
 package org.apache.commons.lang3.builder;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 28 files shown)


PascalSchumacher:
Thanks! 👍 

Merged in https://github.com/apache/commons-lang/commit/729adb624d3e720afb8686093814ab2bcc2d2f13

Not sure why this pull request was not auto-closed even-though I fast-forward merged the branch.

Could you please close the pull request?

Thanks!

mureinik:
> Thanks! 👍
> 
> Merged in [729adb6](https://github.com/apache/commons-lang/commit/729adb624d3e720afb8686093814ab2bcc2d2f13)
> 
> Not sure why this pull request was not auto-closed even-though I fast-forward merged the branch.
> 
> Could you please close the pull request?
> 
> Thanks!

Closing, thanks @PascalSchumacher

PR #366, opened 2018-10-06
Update concurrent tests to JUnit Jupiter

mureinik:
Upgrade the tests in the concurrent package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

Unlike `org.junit.Test`, `org.junit.jupiter.api.Test` does not have an `expected` argument. Instead, an explicit call to `org.junit.jupiter.api.Assertions.assertThrows` is used. This call allows the test to pinpoint the exact statement that is expected to throw the exception and allows making the tests a bit stricter by preventing
false-positives that could occur if the setup code would throw the expected exception instead of the statement that was supposed to throw it.

Another notable change was performed in `MemoizerTest`.
JUnit Jupiter does not support JUnit 4's runners, and EasyMock has no equivalent `@Extension`, so a setup method was added and the mock was explicitly initialized.

It's also worth noting this is a minimal patch for migrating the package's tests to Jupiter. There are several tests that can be made more elegant with Jupiter's new features, but that work is left
for subsequent patches.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializerTest.java
+++ b/src/test/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializerTest.java
@@ -16,12 +16,12 @@
  */
 package org.apache.commons.lang3.concurrent;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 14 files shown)


PascalSchumacher:
Thanks! 👍

PR #367, opened 2018-10-06
Update event tests to JUnit Jupiter

mureinik:
Upgrade the tests in the event package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

While most of these changes are drop-in replacements with no functional benefit, it is worth noting the tests that test thrown exceptions.
Prior to this patch, this package tested for exceptions in two ways, neither of which are supported in JUnit Jupiter:
1. With the `expected` argument of the `org.junit.Test` annotation.
2. With the `org.junit.rules.ExpectedException` `@Rule`

Both of these usages were replaced with calls to `org.junit.jupiter.api.Assertions#assertThrows`.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/event/EventListenerSupportTest.java
+++ b/src/test/java/org/apache/commons/lang3/event/EventListenerSupportTest.java
@@ -17,8 +17,9 @@
 
 package org.apache.commons.lang3.event;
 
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Thanks! 👍

PR #368, opened 2018-10-07
Update exception tests to JUnit Jupiter

mureinik:
Upgrade the tests in the `exception` package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

Another non-obvious change was performed in `ContextedRuntimeExceptionTest`. Unlike JUnit Vintage's `@Before`, JUnit Jupiter's `@BeforeEach` does not apply if a parent's method is annotated with it and the overriding method is not, so an explicit `@BeforeEach` annotation had to be added to `ContexedTuntimeExceptionTest#setUp()`.


PascalSchumacher:
Thanks! 👍

PR #369, opened 2018-10-07
Update reflect tests to JUnit Jupiter

mureinik:
Upgrade the tests in the `reflect` package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

Unlike `org.junit.Assume`, `org.junit.jupiter.api.Assumptions` does not have an `assumeNotNull` method. Instead, `assumeTrue` was used with an explicit condition of a variable being different than `null`.

It's also worth noting this is a minimal patch for migrating the package's tests to Jupiter. There are several tests that can be made more elegant with Jupiter's new features, but that work is left for
subsequent patches.


PascalSchumacher:
Thanks! 👍

PR #371, opened 2018-10-08
Update tuple tests to JUnit Jupiter

mureinik:
Upgrade the tests in the text package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.


PascalSchumacher:
Thanks! 👍

PR #373, opened 2018-10-09
Update mutable tests to JUnit Jupiter

mureinik:
Upgrade the tests in the `mutable` package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

Unlike `org.junit.Assert.assertEquals(double, double, double)`, `org.junit.jupiter.api.Assertions.assertEquals(double, double, double)` does not support deltas of zero, only strictly positive deltas.
This issue will be addressed in JUnit Jupiter 5.4 (see https://github.com/junit-team/junit5/pull/1613 for details). In the meanwhile, `assertTrue(expected==actual)` was used, and `TODO` comments were placed in the code to refactor it to assertEquals once JUnit 5.4 is available.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/mutable/MutableBooleanTest.java
+++ b/src/test/java/org/apache/commons/lang3/mutable/MutableBooleanTest.java
@@ -17,11 +17,12 @@
 
 package org.apache.commons.lang3.mutable;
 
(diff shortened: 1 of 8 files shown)


PascalSchumacher:
Thanks! 👍

PR #374, opened 2018-10-09
Update time tests to JUnit Jupiter

mureinik:
Upgrade the tests in the `time` package to use JUnit Jupiter as part of the effort to remove the dependency on the Vintage Engine.

JUnit Jupiter no longer has a concept of runners. Tests previously run with the `org.junit.runners.Parameterized` runner were rewritten to use `@ParameterizedTest` and a `@MethodSource`.

JUnit Jupiter also no longer has the concept of Rules. Usages of the `SystemDefaultsSwitch` rule and its accompanying annotates were replaced with the `@DefaultLocale` and `@DefaultTimezone` annotations that @britter contributed to [JUnit Pioneer](https://github.com/junit-pioneer/junit-pioneer), the semi-official JUnit extension project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -544,6 +544,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit-pioneer</groupId>
+      <artifactId>junit-pioneer</artifactId>
+      <version>0.2.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-all</artifactId>
       <version>1.3</version>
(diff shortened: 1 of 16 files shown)


mureinik:
@britter I was half way through rewriting the `SystemDefaultsSwitch` as a JUnit Jupiter extension when I remembered your tweet from earlier this month and figured out that you already did this work in JUnit Pioneer. I hope I'm not stepping on any toes here, but if this conflicts with anything you had planned, I'd gladly abandon this PR.

pom.xml
```diff
@@ -543,6 +543,11 @@
       <artifactId>junit-vintage-engine</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit-pioneer</groupId>
+      <artifactId>junit-pioneer</artifactId>
+      <version>0.2.0</version>
```
PascalSchumacher: This dependency should have `test` scope.

mureinik: Good catch, thanks.

mureinik:
Updated the patch to fix the scoping issue in `pom.xml` @PascalSchumacher commented about.

PascalSchumacher:
Thanks! 👍

PR #375, opened 2018-10-11
JUnit Jupiter migration completion

mureinik:
This patch finalizes the upgrade of commons-lang's tests to use JUnit Jupiter and removes the Vintage Engine dependency entirely.

Unlike `org.junit.Assert.assertEquals(double, double, double)`, `org.junit.jupiter.api.Assertions.assertEquals(double, double, double)` does not support deltas of zero, only strictly positive deltas.
This issue will be addressed in JUnit Jupiter 5.4 (see https://github.com/junit-team/junit5/pull/1613 for details). In the meanwhile, `assertTrue(expected==actual)` was used, and `TODO` comments were placed in the code to refactor it to assertEquals once JUnit 5.4 is available.

Unlike `org.junit.Test`, `org.junit.jupiter.api.Test` does not have a `timeout` argument either. Instead, an explicit call to `org.junit.jupiter.api.Assertions.assertTimeoutPreemptively` is used.

JUnit Jupiter also no longer has the concept of Rules. Usages of the `SystemDefaultsSwitch` rule and its accompanying annotates were replaced with the `@DefaultLocale` annotation that @britter contributed to JUnit Pioneer, the semi-official JUnit extension project, or simply removed entirely where they had no use (i.e., from `StringUtilsEqulasIndexOfTest`).
Following the removal of their usages, the `@SystemDefaults` annotation, the `SystemDefaultsSwitch` rule and the `SystemDefaultsSwitchTest` class that tests them had no more use, and they were removed entirely.


PascalSchumacher:
Thank you very much for this massive effort! 👍

PR #376, opened 2018-10-13
Test cleanup

mureinik:
Following up after the migration of the test suite to JUnit Jupiter, some test code could be cleaned up to make it clearer and easier to maintain.
This PR contains several patches around that area:

General fixes:
- In order to test an expected exception, `assertThrows` was used instead of a cumbersome `if`-`fail`-`catch` construct.
- In order to fail a test on an unexpected exception, the exception is thrown outside the method, instead of the cumbersome construct of catching it and calling `fail`
- Redundant `throws` clauses were removed
- `assertEquals`, `assertNotEquals`, `assertSame`, `assertNotSame`, `assertNull` and `assertNotNull` were used instead of reimplementing them with conditions in `if` statements or `assertTrue`s.

Specific improvements:
- `ConverstionTest#assertBinaryEquals` was removed in favor of the built-in `Assertions#assertArraysEquals(boolean[], boolean[])`.
- `LocaleUtilsTest#parseAllLocales` was rewritten as a parameterized test instead of implementing it manually with a loop.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
@@ -24,6 +24,7 @@ import static org.apache.commons.lang3.AnnotationUtilsTest.Stooge.MOE;
 import static org.apache.commons.lang3.AnnotationUtilsTest.Stooge.SHEMP;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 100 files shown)


mureinik:
> Looks great, thank you very much! 👍
> 
> There are just some checkstyle errors related to unused imports left to fix (see failed travis build).

Missed a couple of unused import statements after I cleaned up the usages of `assertTrue` and `assertFalse`. I removed them, and the build should pass now.

PascalSchumacher:
Thanks! 👍

PR #385, opened 2018-11-21
WIP: CheckedFunction utils

asciborek:
This is an example of utils which I would like to create. Purpose of interfaces like this is to simplify usage of java streams


src/test/java/org/apache/commons/lang3/function/CheckedFunctionTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.function.Function;
+
+public class CheckedFunctionTest {
+
+    @DisplayName("unchecked should return a correctly working function")
```
ajs6f: This doesn't seem to test the interesting behavior of the code. There is no exception rethrown as a `RuntimeException`. Shouldn't that be happening?

PR #392, opened 2018-12-08
LANG-1410: eliminate one unnecessary local int variable and add more tests

rosti-il:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -338,15 +338,17 @@ public static boolean isAllEmpty(final CharSequence... css) {
      * @since 3.0 Changed signature from isBlank(String) to isBlank(CharSequence)
      */
     public static boolean isBlank(final CharSequence cs) {
-        int strLen;
```
asciborek: I think that name strLen is more informative

rosti-il: Maybe, however 'strLen' sounds like something unchangeable, since String is immutable in Java. I've chosen this variable name to be the same to name of the similar 'sz' variable in isWhitespace() method. The 'sz' name stands for "size", I think. If we decide to change it we should change names of they both.

MarkDacek: I tend to agree with sz being more appropriate of a name here than strLen. It would appear that no other variable named `strLen` ever deviates from its original assignment (even the non-final ones.) 

Saving a loop variable isn't a great gain in memory usage but I suppose this is a bit more appropriate. Though, thinking ahead for the future - can one imagine having to reference the string length for a future change? That would be my opposition for this.

src/test/java/org/apache/commons/lang3/StringUtilsEmptyBlankTest.java
```diff
@@ -99,6 +110,7 @@ public void testIsNotBlank() {
         assertFalse(StringUtils.isNotBlank(null));
         assertFalse(StringUtils.isNotBlank(""));
         assertFalse(StringUtils.isNotBlank(StringUtilsTest.WHITESPACE));
+        assertTrue(StringUtils.isNotBlank("a"));
```
MarkDacek: I might be missing something, but this appears redundant with the next line. Did this add any coverage?

rosti-il: This is a special case that fails if you accidentally use `index > 0` instead of `index >= 0` in the expression of the down going loop. The next line will not fail in this case.

src/main/java/org/apache/commons/lang3/StringUtils.java
rosti-il: @MarkDacek, this and also the `isWhitespace()` methods are just string scan methods. They don't need the string length except the very beginning. Thus the original value returned by `cs.length()` doesn't need to be memorized.

src/test/java/org/apache/commons/lang3/StringUtilsEmptyBlankTest.java
MarkDacek: Might be worth adding a comment in there for that.

rosti-il: Done in second commit.

rosti-il:
Force-pushed twice from 27b1742 to 9d8ac06 and then back from 9d8ac06 to 27b1742 because of a confusion about the reason of the Travis CI build failure. The reason of this failure is the a410aab6515931489dc2594d4e0f0ecac2c66071 commit of @garydgregory at master branch.

Will the Travis CI build run again automatically after @garydgregory fixed his commit?

src/main/java/org/apache/commons/lang3/StringUtils.java
asciborek: So maybe instead of "sz" the name should be "i" or "index"? It's quite good name for a looping-control variable, isn't?

rosti-il: Or `idx`? But aren't we spend too much time for a discussion of one variable name? :-)

MarkDacek: Hm, I've given another look through and it would appear that other uses of `sz` are final as well. 

I don't object to this, but I would question the need for it as the memory gain isn't noticeable.

eceejcr: Regarding the final keyword for ‘sz’, in my view makes sense to make the code uniform and in general, the final keyword is used in the surrounding affected for variables non-mutating (even for variables inside methods).  Personally, I like it because think it makes the code more readable providing information about the mutability of variable

rosti-il: So what name do you think is better in this case? `index`, `idx` or something else?

eceejcr: idx is fine for me

garydgregory:
Closing: Out of date with current tests.

PR #395, opened 2019-01-04
Replaces the given String, with the String which is nested in between two Strings.

Yash-777:
Replace String with the String which is nested in between two Strings.

I have used this Method in order to generate dynamic HTML reports to send a mail.

```java	    
public class Test {
    public static void main(String[] args) throws Exception {
        String data = new String(Files.readAllBytes(Paths.get("./dynamicReport.html")));
        String rowSaperator = "<!--  [$$ListRow$$] -->";
        String rowDataSaperator = "[~$List$~]";
        
        List<String[]> dynamicRecords = new ArrayList<>();
        dynamicRecords.add( new String[]{ "1", "1" } );
        dynamicRecords.add( new String[]{ "1", "1" } );
        dynamicRecords.add( new String[]{ "2", "1" } );
        dynamicRecords.add( new String[]{ "1", "2" } );
        
        String finalTableData = dynamicRecordData(data, rowSaperator, rowDataSaperator, dynamicRecords);
        
        String targetFilePath = "D:/dynamicReport.txt";
        try (PrintWriter out = new PrintWriter( targetFilePath )) {
            out.println( finalTableData );
        }
    }
    
    public static String dynamicRecordData(String str, String rowSaperator, String rowDataSaperator, List<String[]> dynamicRecords) throws Exception {
        if( str.length() <= ( (rowSaperator.length())*2 + rowDataSaperator.length()) ) {
            return str;
        }
        
        String substringBetween = StringUtils.substringBetween(str, rowSaperator, rowSaperator);
        if( substringBetween == null) {
            return str;
        }
        
        int countMatches = StringUtils.countMatches(substringBetween, rowDataSaperator);
        if( str.length() <= ( (rowSaperator.length())*2 + (rowDataSaperator.length())*countMatches ) ) {
            return str;
        }
        
        StringBuffer dynamicRowData = new StringBuffer();
        
        String[] searchList = null;
        for (String[] dynamicRecord : dynamicRecords) {
            if( dynamicRecord.length == countMatches ) {
                if( searchList == null ) {
                    searchList = new String[countMatches];
                    for (int i = 0; i < countMatches; i++) {
                        searchList[i] = rowDataSaperator;
                    }
                }
                
                String replaceEach = StringUtils.replaceEach(substringBetween, searchList, dynamicRecord);
                
                dynamicRowData.append(replaceEach);
            } else {
                throw new Exception("Records mismatch.");
            }
        }
        
        String finalTableData = StringUtils.replaceSubstringInBetween(str, dynamicRowData.toString(), rowSaperator, rowSaperator);
        return finalTableData;
    }
}
```
Sample HTML Template. Where i my case there are to many dynamic tables with provided list data to replace.
```html
<table>
    <tbody>
        <tr><td>Header1</td><td>Header2</td>
        <!--  [$$ListRow$$] -->
        <tr><td>[~$List$~]</td><td>[~$List$~]</td>
        <!--  [$$ListRow$$] -->
    </tbody>
</table>
```
After code generated report.

```html
<table>
    <tbody>
        <tr><td>Header1</td><td>Header2</td>
        <!--  [$$ListRow$$] -->
        <tr><td>1</td><td>1</td>
        
        <tr><td>1</td><td>1</td>
        
        <tr><td>2</td><td>2</td>
        
        <tr><td>1</td><td>1</td>
        <!--  [$$ListRow$$] -->
    </tbody>
</table>
```
Avoid previous pull requests [`393`, `394`](https://travis-ci.org/apache/commons-lang/pull_requests)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3124,6 +3124,45 @@ public class StringUtils {
         return list.toArray(new String [list.size()]);
     }
 
+    /**
+	 * <p>Replaces the given String, with the String which is nested in between two Strings.</p>
+	 *
+	 * <p>A {@code null} input String returns {@code null}.
+	 * A {@code null} open/close returns {@code null} (no match).
+	 * An empty ("") open and close returns an empty string.</p>
+	 *
+	 * @param str  the String containing the substring, may be null
+	 * @param replace the Sting to be replaced, which is nested in between open and close substrings
+	 * @param open  the String before the substring, may be null
+	 * @param close  the String after the substring, may be null
+	 * @return the substring, {@code null} if no match
+	 */
+	public static String replaceSubstringInBetween(final String str, final String replace, final String open, final String close) {
+		if (str == null || open == null || close == null) {
+			return null;
+		}
+		final int start = str.indexOf(open);
+		if (start != INDEX_NOT_FOUND) {
+			String preceding = "", exceding = "";
+			if( start > 0 ) {
+				preceding = str.substring(0, start);
+			}
+			
+			final int end = str.indexOf(close, start + open.length());
+			if (end != INDEX_NOT_FOUND) {
+				if( (end < str.length()-1) && (end+close.length() < str.length() ) ) {
+					exceding = str.substring(end+close.length(), str.length());
+				}
+				
+				String middleString = str.substring(start + open.length(), end);
+				System.out.println("Middle Str : "+ middleString);
+				
+				return preceding+open+replace+close+exceding;
+			}
+		}
+		return null;
+	}
+    
     // Nested extraction
     //-----------------------------------------------------------------------
 
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3147,7 +3147,6 @@ public class StringUtils {
             if (start > 0) {
                 preceding = str.substring(0, start);
             }
-            
             final int end = str.indexOf(close, start + open.length());
             if (end != INDEX_NOT_FOUND) {
                 String exceding = "";
@@ -3160,7 +3159,6 @@ public class StringUtils {
         }
         return null;
     }
-    
     // Nested extraction
     //-----------------------------------------------------------------------
 
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3125,43 +3125,41 @@ public class StringUtils {
     }
 
     /**
-	 * <p>Replaces the given String, with the String which is nested in between two Strings.</p>
-	 *
-	 * <p>A {@code null} input String returns {@code null}.
-	 * A {@code null} open/close returns {@code null} (no match).
-	 * An empty ("") open and close returns an empty string.</p>
-	 *
-	 * @param str  the String containing the substring, may be null
-	 * @param replace the Sting to be replaced, which is nested in between open and close substrings
-	 * @param open  the String before the substring, may be null
-	 * @param close  the String after the substring, may be null
-	 * @return the substring, {@code null} if no match
-	 */
-	public static String replaceSubstringInBetween(final String str, final String replace, final String open, final String close) {
-		if (str == null || open == null || close == null) {
-			return null;
-		}
-		final int start = str.indexOf(open);
-		if (start != INDEX_NOT_FOUND) {
-			String preceding = "", exceding = "";
-			if( start > 0 ) {
-				preceding = str.substring(0, start);
-			}
-			
-			final int end = str.indexOf(close, start + open.length());
-			if (end != INDEX_NOT_FOUND) {
-				if( (end < str.length()-1) && (end+close.length() < str.length() ) ) {
-					exceding = str.substring(end+close.length(), str.length());
-				}
-				
-				String middleString = str.substring(start + open.length(), end);
-				System.out.println("Middle Str : "+ middleString);
-				
-				return preceding+open+replace+close+exceding;
-			}
-		}
-		return null;
-	}
+     * <p>Replaces the given String, with the String which is nested in between two Strings.</p>
+     *
+     * <p>A {@code null} input String returns {@code null}.
+     * A {@code null} open/close returns {@code null} (no match).
+     * An empty ("") open and close returns an empty string.</p>
+     *
+     * @param str  the String containing the substring, may be null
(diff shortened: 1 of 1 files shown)


Yash-777:
Can any one explain that, the build is fail because of  extra added function **`replaceSubstringInBetween`** in class name `StringUtils`.

>  #1454.2
```log
[INFO] Running org.apache.commons.lang3.RandomStringUtilsTest
[ERROR] Tests run: 18, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.226 s <<< FAILURE! - in org.apache.commons.lang3.RandomStringUtilsTest
[ERROR] testRandomStringUtilsHomog  Time elapsed: 0.019 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: test homogeneity -- will fail about 1 in 1000 times ==> expected: <true> but was: <false>
	at org.apache.commons.lang3.RandomStringUtilsTest.testRandomStringUtilsHomog(RandomStringUtilsTest.java:464)
```
>  #1454.2

```log
[ERROR] Failures: 
[ERROR]   FieldUtilsTest.testRemoveFinalModifier:998 expected: <false> but was: <true>
[ERROR]   FieldUtilsTest.testRemoveFinalModifierWithAccess:1009 expected: <false> but was: <true>
```

ecki:
The test States `will fail about 1 in 1000 times`

ecki:
Can you please merge your commits to a single one, so it is easier to review and apply.

Yash-777:
Already i have closed the pull requests `393, 394`

MarkDacek:
For one, the default return value should be the original `str`, not null. That way, it would be much simpler to use it safely, e.g.:
`list.add(StringUtils.replaceSubstringInBetween(args..))`
instead of having to add in a null check before adding it.

Please add sufficient unit-tests to show that this functions as expected.

I would also suggest that a boolean `repeat` be added, to allow this to be done indefinitely. This might require a slight refactor. 

Other comments to be made inline.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3124,6 +3124,41 @@ public static String substringBetween(final String str, final String open, final
...
 
+    /**
+     * <p>Replaces the given String, with the String which is nested in between two Strings.</p>
+     *
+     * <p>A {@code null} input String returns {@code null}.
+     * A {@code null} open/close returns {@code null} (no match).
+     * An empty ("") open and close returns an empty string.</p>
+     *
+     * @param str  the String containing the substring, may be null
+     * @param replace the Sting to be replaced, which is nested in between open and close substrings
```
MarkDacek: Based on your code, this is "the String to replace the inner substring".

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3124,6 +3124,41 @@ public static String substringBetween(final String str, final String open, final
...
+        final int start = str.indexOf(open);
+        if (start != INDEX_NOT_FOUND) {
+            String preceding = "";
+            if (start > 0) {
+                preceding = str.substring(0, start);
+            }
+            final int end = str.indexOf(close, start + open.length());
+            if (end != INDEX_NOT_FOUND) {
+                String exceding = "";
+                if ((end < str.length() - 1) && (end + close.length() < str.length())) {
```
MarkDacek: I'm lost on why this `if` is needed. I would think you'd just call the `str.substring(end)` and add together `preceeding, open, replace, end`

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3124,6 +3124,41 @@ public static String substringBetween(final String str, final String open, final
...
+        }
+        final int start = str.indexOf(open);
+        if (start != INDEX_NOT_FOUND) {
+            String preceding = "";
+            if (start > 0) {
+                preceding = str.substring(0, start);
+            }
+            final int end = str.indexOf(close, start + open.length());
+            if (end != INDEX_NOT_FOUND) {
+                String exceding = "";
```
MarkDacek: exceeding

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3124,6 +3124,41 @@ public static String substringBetween(final String str, final String open, final
...
+                if ((end < str.length() - 1) && (end + close.length() < str.length())) {
+                    exceding = str.substring(end + close.length(), str.length());
+                }
+                //String middleString = str.substring(start + open.length(), end);
```
MarkDacek: please remove

src/main/java/org/apache/commons/lang3/StringUtils.java
MarkDacek: also, `Sting`

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3124,6 +3124,41 @@ public static String substringBetween(final String str, final String open, final
...
+     */
+    public static String replaceSubstringInBetween(final String str, final String replace, final String open, final String close) {
+        if (str == null || open == null || close == null) {
+            return null;
+        }
+        final int start = str.indexOf(open);
+        if (start != INDEX_NOT_FOUND) {
+            String preceding = "";
+            if (start > 0) {
+                preceding = str.substring(0, start);
```
MarkDacek: Similar to my comment below, I'd think it wise to simplify.
`String preceding = str.substring(0, start + open.length());`

MarkDacek: As is, you would still need to account for negative indices being returned from `str.indexOf(open) `and `str.indexOf(close, start + open.length())`

kinow:
I am not sure if this belongs to commons-lang. It has a practical use, and I can see you put some effort to document it and write some tests (thanks for that). But being a low level library, I feel like this would be a very simplistic template engine.

Don't have time to create a working example, but I suspect I could do that with Apache Commons JEXL, Apache Velocity, or another complete template engine.

This looks more related to Apache Commons Text (plus `StringUtils` is quite bloated too), but I would still keep the argument above before merging it into [text]. The replacement part can be done with `StrSubstitutor` from text. The iteration part, that creates multiple `<td>` entries... well, that part I am not sure if we have in Commons Lang or Text something similar.

Anyway, if others still want to merge this new feature, it might be a good idea to drop an e-mail to the mailing list to see what others think.

Cheers
Bruno

Yash-777:
creating new pull request.

kinow:
@Yash-777 I think it would be easier if you used the same pull request. This makes it a lot harder to review, especially as all the previous comments from other devs will be lost on your new pull request. You can edit commits, or push new commits to your own branch, and that should update the pull request automatically.

PR #398, opened 2019-01-17
Add ComparableUtils

Sam-Kruglov:
I have turned `feeAmount.compareTo(BigDecimal.ZERO) > 0` into `is(feeAmount).greaterThan(BigDecimal.ZERO)` and decided to share.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ComparableUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ComparableUtils.java
@@ -1,0 +1,206 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.lang3;
+
+import java.util.function.Predicate;
+
+/**
+ * <p>Utility library to provide helper methods for translating {@link Comparable#compareTo} result into a boolean.</p>
+ *
+ * <p>Example: {@code boolean x = is(myComparable).lessThanOrEqualTo(otherComparable)}</p>
+ *
+ * <p>#ThreadSafe#</p>
+ *
+ * @since 3.10
+ */
+public class ComparableUtils {
+
+    private ComparableUtils() {}
+
+    /**
+     * Provides access to the available methods
+     *
+     * @param a base object in the further comparison
+     * @param <A> type of the base object
+     * @return a builder object with further methods
+     */
+    public static <A extends Comparable<A>> ComparableCheckBuilder<A> is(A a) {
+        return new ComparableCheckBuilder<>(a);
+    }
+
+    /**
+     * Checks if the tested object is less than {@code b}
+     *
+     * @param b the object to compare to the tested object
+     * @param <A> type of the test object
+     * @return a predicate for true if the value returned by {@link Comparable#compareTo} is less than {@code 0}
+     */
+    public static <A extends Comparable<A>> Predicate<A> lt(A b) {
+        return a -> is(a).lessThan(b);
+    }
+
+    /**
+     * Checks if the tested object is less than or equal to {@code b}
+     *
+     * @param b the object to compare to the tested object
+     * @param <A> type of the test object
+     * @return a predicate for true if the value returned by {@link Comparable#compareTo}
+     * is less than or equal to {@code 0}
+     */
+    public static <A extends Comparable<A>> Predicate<A> le(A b) {
+        return a -> is(a).lessThanOrEqualTo(b);
+    }
+
+    /**
+     * Checks if the tested object is greater than {@code b}
+     *
+     * @param b the object to compare to the tested object
+     * @param <A> type of the test object
+     * @return a predicate for true if the value returned by {@link Comparable#compareTo} is greater than {@code 0}
+     */
+    public static <A extends Comparable<A>> Predicate<A> gt(A b) {
+        return a -> is(a).greaterThan(b);
+    }
+
+    /**
+     * Checks if the tested object is greater than or equal to {@code b}
+     *
+     * @param b the object to compare to the tested object
+     * @param <A> type of the test object
+     * @return a predicate for true if the value returned by {@link Comparable#compareTo}
+     * is greater than or equal to {@code 0}
+     */
+    public static <A extends Comparable<A>> Predicate<A> ge(A b) {
+        return a -> is(a).greaterThanOrEqualTo(b);
+    }
+
+    /**
+     * Checks if {@code [b <= a <= c]} or {@code [b >= a >= c]} where the {@code a} is the tested object.
+     *
+     * @param b the object to compare to the tested object
+     * @param c the object to compare to the tested object
+     * @param <A> type of the test object
+     * @return a predicate for true if the tested object is between b and c
+     */
+    public static <A extends Comparable<A>> Predicate<A> between(A b, A c) {
+        return a -> is(a).between(b, c);
+    }
+
+    /**
+     * Checks if {@code (b < a < c)} or {@code (b > a > c)} where the {@code a} is the tested object.
+     *
+     * @param b the object to compare to the tested object
+     * @param c the object to compare to the tested object
+     * @param <A> type of the test object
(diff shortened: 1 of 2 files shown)


Sam-Kruglov:
Good idea will add. But following that logic, the whole commons is just an overhead then.

GedMarc:
@MarkDacek compareTo() is null safe by default, that's the difference vs compare(T a1,T a2)? 
So adding in null checks into compareTo() seems likes a waste of overhead?

Not sure if i'm misunderstanding this - http://www.javapractices.com/topic/TopicAction.do?Id=10

Sam-Kruglov:
@MarkDacek @GedMarc actually that's correct, compareTo is null safe, here is a piece of its javadoc: `@throws NullPointerException if the specified object is null`

Sam-Kruglov:
CI seems to fail due to some access permissions related to Java 9 module system. Some JavaDocs can't compile. Passes JDK 8. There isn't really much to this PR, just a simple util

PascalSchumacher:
Yes, the ci javadoc failure probably is: https://bugs.openjdk.java.net/browse/JDK-8212233

MarkDacek:
Sorry, was on vacation for a bit. The compareTo method is defined by each class. Not everything will check for nulls, though I would assume most JRE classes would. 

I might think that this idea is better accomplished using static methods and building in null-checks. `new ComparableCheckBuilderl(a).greaterThan(b)` doesn't seem worthwhile when `a.compareTo(b) > 0` is a less-expensive operation. 

Something like `ComparableUtils.greaterThan(a,b)` might be better, though I'm still unsure of its usefulness.

Sam-Kruglov:
@MarkDacek thanks, for the input. This is intended to be a fluent API (like AssertJ) so it reads very well in terms of the English language.
Doing `ComparableUtils.greaterThan(a,b)` is just worthless IMHO.
Doing `new ComparableCheckBuilder(a).greaterThan(b)` is actually exactly the same as I currently suggest, except that `new ComparableCheckBuilder(a)` part is actually returned by static factory method `ComparableUtils.is(a)`.
About null checks, since in some cases the check is already there and by Comparable docs it is supposed to throw NPE, I see no reason at all to include that. Although if I wanted, it is perfectly possible to do so inside the builder.

chtompki:
What's the goal with this class?? is the `Comparable` interface insufficient? I understand the idea behind adding a better naming convention, but still shouldn't we just learn how `Comparable` works?

garydgregory:
-1 Over the top, the same 'legibility' IMO can simply be achieved with:

```
public class ComparableUtils {

    public enum Compared {EQUALS, GREATER, LESSER};
    
    public static Compared compare(Comparable c1, Object o) {
        final int result = c1.compareTo(o);
        return result == 0 ? Compared.EQUALS : result > 0 ? Compared.GREATER : Compared.LESSER;
    }
    
    @Test
    public void test() {
        Assertions.assertEquals(Compared.GREATER, ComparableUtils.compare(1, 0));
        Assertions.assertEquals(Compared.LESSER, ComparableUtils.compare(0, 1));
        Assertions.assertEquals(Compared.EQUALS, ComparableUtils.compare(0, 0));
    }
}
```

Sam-Kruglov:
That's supposed to be used in production code, not in test code. Test libraries already cover it nicely.

I sugfest: `if (is(fee).greaterThanOrEqualTo(discountThreshold) {...}`
You suggest 
```
Compared feeComparedToDiscountThreshold = compare(fee, discountThreshold);
if (feeComparedToDiscountThreshold == EQUALS || feeComparedToDiscountThreshold == GREATER) {...}
```
and it's not immediately clear if fee is greater than the threshold or the other way around, so you may have to look at the javadoc, which is exactly the same thing as with the `Comparable`.

garydgregory:
@Sam-Kruglov That's a nice comparison in favor of your scheme. Thank you for providing it. What do others think?

src/main/java/org/apache/commons/lang3/ComparableUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    public static class ComparableCheckBuilder<A extends Comparable<A>> {
+
+        private final A a;
+
+        private ComparableCheckBuilder(A a) {
+            this.a = a;
+        }
+
+        /**
+         * Checks is the value returned by {@link Comparable#compareTo} is equal to {@code 0}
```
kinow: I think all the javadocs here have the same typo, "Checks is the value..." should be "Check if the value..." ?

garydgregory: It should read "Checks if the value..."

kinow: Ops, good spot @garydgregory

Sam-Kruglov: fixed thanks

Sam-Kruglov:
I am not aware what is `changes.xml`, but looks like there is an automatic Jira already https://issues.apache.org/jira/browse/LANG-1431

Sam-Kruglov:
changed `@since 3.10`

src/test/java/org/apache/commons/lang3/ComparableUtilsTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+
+import static org.apache.commons.lang3.ComparableUtils.is;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ComparableUtilsTest {
```
garydgregory: Hi @Sam-Kruglov ,
Why is this class not `public'?

Sam-Kruglov: Just so I don't have to type the word "public" :D
Should I make it public?

PascalSchumacher: @garydgregory 

>Test classes, test methods, and lifecycle methods are not required to be public, but they must not be private. 

https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

Sam-Kruglov:
I am back with a private constructor for the main class and a `public` modifier for the test!

garydgregory:
This now reminds me of JPA's [`CriteriaBuilder`](https://docs.oracle.com/javaee/7/api/javax/persistence/criteria/CriteriaBuilder.html) and the work I did writing [_Java Persistence with Hibernate, Second Edition_](https://www.manning.com/books/java-persistence-with-hibernate-second-edition) where `CriteriaBuilder` includes shorthand methods like `le` for `lessThanOrEqualTo`, `lt` for `lessThan`, and so on. We should not implement `CriteriaBuilder` of course but rather get inspiration from it.

Thoughts?

Sam-Kruglov:
I like fully written names more. We can take inspiration from AssertJ :D http://joel-costigliola.github.io/assertj/core/api/index.html?org/assertj/core/internal/Comparables.html

garydgregory:
@Sam-Kruglov, 

Did you look at CriteriaBuilder? I am not arguing about not one style over the other, CriteriaBuilder offers both.

Sam-Kruglov:
No, I didn't, but now I did. 
I see, good idea, we could definitely include `between` in there.
I noticed, that the short names are used for returning `Predicate`. It is similar how in tests Hamcrest uses naming for returning `Matchers`. Do you think we should do that or also return `boolean`? I think we should return `Predicate` - can easily be used with `Stream#filter`

Sam-Kruglov:
Got a little time available. So, I added the `between`, `betweenExclusive`, and all the `Predicate` versions. 
I didn't add any tests for the `Predicate` methods because they are just delegates. I could rewrite existing tests to use predicates to get better coverage but it would look less readable.

Sam-Kruglov:
@kinow @garydgregory So, can we merge that?

garydgregory:
> Looks good to me. All tests passing too, and I liked the final API. +1 from me, will leave the final review/merge for @garydgregory
> 
> Thanks for the PR @Sam-Kruglov !

I will review tonight.

ameyjadiye:
Initially I though below might fail, however my testings with diffrent values are working fine.
API looks good to me.
```

        @Test
        void ameyVerification() {
            
            BigDecimal bg1, bg2;

            bg1 = new BigDecimal(0.7);
            bg2 = new BigDecimal(0.70000);
            
            assertTrue(is(bg1).equalTo(bg2));
            assertFalse(is(bg1).lessThan(bg2));
            assertFalse(is(bg1).greaterThan(bg2));

        }
```

Sam-Kruglov:
Merry Christmas :) 🎄 
@garydgregory let's merge this if all ok

mjiderhamn:
I was surprised to stumble upon my old [mjiderhamn/compare-to](https://github.com/mjiderhamn/compare-to), [announced back in 2013](https://java.jiderhamn.se/2013/10/15/better-compareto-syntax/), having been integrated into Commons Lang.

@Sam-Kruglov , did you come up with the exact same syntax on your own, or did you happen to forget to give credit where credit was due...?

Sam-Kruglov:
@mjiderhamn I came up with it on my own, haven’t seen yours. But my utility here differs from whats currently in main, someone changed it without PR I think, you can follow commits

PR #400, opened 2019-01-23
Close/flush the `OutputStream` before calling `toByteArray()` on underlying `ByteArrayOutputStream`

emopers:
When an `OutputStream` instance wraps an underlying `ByteArrayOutputStream` instance,
it is recommended to flush or close the `OutputStream` before invoking the underlying instances' `toByteArray()`. Although in some of these case it is not strictly necessary because the
`writeObject()` method is invoked right before `toByteArray()`, and `writeObject()` internally calls `flush()`/`drain()`. However, it is good practice to call `flush()`/`close()` explicitly as mentioned, for example, [here](http://stackoverflow.com/questions/2984538/how-to-use-bytearrayoutputstream-and-dataoutputstream-simultaneously-java).

This pull request adds a call to `close()` or `flush()` before calls to `toByteArray()`.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/tuple/ImmutablePairTest.java
+++ b/src/test/java/org/apache/commons/lang3/tuple/ImmutablePairTest.java
@@ -131,6 +131,7 @@ public class ImmutablePairTest {
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
         final ObjectOutputStream out = new ObjectOutputStream(baos);
         out.writeObject(origPair);
+        out.close();
         final ImmutablePair<Integer, String> deserializedPair = (ImmutablePair<Integer, String>) new ObjectInputStream(
                 new ByteArrayInputStream(baos.toByteArray())).readObject();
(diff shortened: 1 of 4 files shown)


src/test/java/org/apache/commons/lang3/tuple/ImmutablePairTest.java
```diff
@@ -131,6 +131,7 @@ public void testSerialization() throws Exception {
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
         final ObjectOutputStream out = new ObjectOutputStream(baos);
         out.writeObject(origPair);
+        out.close();
```
HubertWo, line 134: ```ObjectOutputStream ``` implements ```AutoCloseable```. What about putting this to ```try-with-resources```?

garydgregory:
@emopers 
Please rebase on master.

garydgregory:
Closing: No longer applies as tests are different now.

PR #401, opened 2019-01-28
Lang 1424 Checked Functions

asciborek:
This pull request contains wrappers for standard java.util.function

### DIFF
--- a/src/main/java/org/apache/commons/lang3/function/CheckedBiConsumer.java
+++ b/src/main/java/org/apache/commons/lang3/function/CheckedBiConsumer.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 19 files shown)


src/main/java/org/apache/commons/lang3/function/CheckedBiConsumer.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param <E> the exception thrown by an instance of CheckedBiConsumer
+     * @return an instance of BiConsumer
+     * @since 3.9
+     */
+    static <T, U, E extends Exception>  BiConsumer<T, U>  unchecked(CheckedBiConsumer<T, U, E> checkedBiConsumer) {
+        return (arg1, arg2) -> {
+            try {
+                checkedBiConsumer.accept(arg1, arg2);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
jvz: You could create a utility method (also useful for commons-lang) similar to what we did in log4j2 to rethrow without wrapping the exception: https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/util/Throwables.java#L91

garydgregory:
Closing: This feature has been in Commons Lang for a while.

PR #404, opened 2019-02-09
Assertion cleanup

mureinik:
Use JUnit features to clean up the assertions in the test so they are easier to read and maintain, and to produce better failure messages in case an assertion fails:

- `assertArraysEqual(x, y)` was used to replace `assertTrue(Arrays.equals(x, y))`
- `assertEquals(x, y)` was used to replace `assertTrue(x == y)`
- `assertNotEquals(x, y)` was used to replace `assertFalse(x == y)`
- Assertions in `FunctionsTest` were rewritten to use `assertThrows` instead of the boilerplate template of throwing an exception in a `try` block, `fail`ing after that statement to make sure the exception was thrown and asserting the properties of the exception in the `catch` block.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ArrayUtilsAddTest.java
+++ b/src/test/java/org/apache/commons/lang3/ArrayUtilsAddTest.java
@@ -17,14 +17,12 @@
 
 package org.apache.commons.lang3;
 
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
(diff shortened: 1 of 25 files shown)


kinow:
I don't have time to review the pull request now @mureinik , but I did have a brief look, and it looks really good! Kudos for the patience creating this patch. Hopefully another dev will jump in and help reviewing it. :clap:

PascalSchumacher:
Thanks! 👍

PR #405, opened 2019-02-11
ensure java version are supported even when not in our enum

rmannibucau:
Last java 11 version is 11.0.2 which breaks JavaVersion current impl. This fixes that and opens (making public) getJavaVersion static method to let a user convert System.getProperty("user.version") to a JavaVersion instance.


src/test/java/org/apache/commons/lang3/JavaVersionTest.java
```diff
@@ -83,4 +90,10 @@ public void testToString() {
         assertEquals("1.2", JAVA_1_2.toString());
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"11" ,"11.0", "11.0.2"})
```
Stzx:
```suggestion
    @ValueSource(strings = {"11", "11.0", "11.0.2"})
```

garydgregory:
Hi @rmannibucau 
Please revisit or close. TY.

PR #406, opened 2019-02-12
[LANG-1407] DurationUtils - methods for rounding duration 

asciborek:
This class contains methods for rounding duration in a given unit, 
for instance when Duration instance is equal to 5h 31 minutes and you want to get rounded value of hours you can use the method:
`roundHours(Duration duration)`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationUtils.java
@@ -30,6 +30,7 @@ public class DurationUtils {
     private static final long NANOS_PER_HOUR = 3_600_000_000_000L;
     private static final long NANOS_PER_MINUTE = 60_000_000_000L;
     private static final long NANOS_PER_SECOND = 1_000_000_000L;
+    private static final long NANOS_PER_MILLISECOND = 1_000_000L;
 
     /**
      * <p>DurationUtils instances should NOT be constructed in standard programming.</p>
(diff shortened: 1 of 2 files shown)


sebbASF:
Thanks.

However, the class Javadoc does not make it clear that there are two types of rounding involved, nor does do the methods explain what rounding is done, particularly for the roundDays etc methods.

What about supporting Banker's rounding?
This rounds to the nearest even number, e.g. 0.5 => 0 and 1.5 => 2

I think it might be better to have a single round method for each time unit, with a parameter to specify the rounding type: up, half-up, bankers (are there any more rounding methods?)

Or even a single method for all time units.

asciborek:
@sebbASF 
you're right about javadoc, I'm going to improve it.
I created only methods for rounding up and normal rounding because I believe that are the most common cases in (business) applications, for instance when you need to calculate a payment  you probably need to round the value of hours. I will try to create more generics method for rounding duration .

asciborek:
@sebbASF 
I've improved Javadoc, however I don't see if there is good way to create a single method per rounding type, because it would cause "ifology".

chtompki:
I'm not even sure that "rounding" is what's going on here. Specifically:

```java
 //given
Duration duration = Duration.of(2, ChronoUnit.DAYS).plusMillis(1);
//when
long roundedDays = DurationUtils.roundUpDays(duration);
//then
Assertions.assertEquals(3, roundedDays);
```

looks a whole lot like a ceiling calculation. I'm I'm curious if there's literature or existent examples out there in the wild about how to name this sort of functionality. I would think we would want to follow what ever convention in place is already out there.

asciborek:
@chtompki Actually I think you are right about this naming. "Ceiling" suits more to this kind of calculation, e.g. in the spreadsheet programs that functionality it called "Ceiling", therefore I believe I should change this method name.

ericbarnhill:
"Rounding a number means replacing it with a different number that is approximately equal to the original, but has a shorter, simpler, or more explicit representation"

Read more at https://en.wikipedia.org/wiki/Rounding

PR #408, opened 2019-02-24
LANG-1435: Implemented defaultIfNull method that uses a supplier for default value

dekelpilli:
From JIRA ticket: 

> Currently, if the default value in defaultIfNull is computationally slow (for example, it requires a database call, or complex calculations), it's not worth using the util method. This feature is to implement a method that would only perform this computationally slow action if the object value is actually null. This can be done using a Supplier.

> In short, you'd call ObjectUtils.getDefaultIfNull(someNullableValue, DbUtil::doGet) instead of ObjectUtils.defaultIfNull(someNullableValue, DbUtil.doGet())

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -180,11 +180,11 @@ public class ObjectUtils {
      *  <p>Returns a default value if the object passed is {@code null}.</p>
      *
      * <pre>
-     * ObjectUtils.defaultIfNull(null, () -> null) = null
-     * ObjectUtils.defaultIfNull(null, () -> "")   = ""
-     * ObjectUtils.defaultIfNull(null, () -> "zz") = "zz"
-     * ObjectUtils.defaultIfNull("abc", *)         = "abc"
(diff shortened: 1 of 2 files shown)


MarkDacek:
I'm curious whether this is worthwhile given the existence of ObjectUtils.firstNonNull. Is the supplier that much of a necessity?

The only plus I see is that it wouldn't be executed if the default isn't null. Is that a major performance increase for a significant number of use-cases?

PascalSchumacher:
Thanks for the pull request!

Java 11 offers `Objects#requireNonNullElseGet` (see https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#requireNonNullElseGet(T,java.util.function.Supplier) for details), so I don't think we should add this.

chtompki:
Ok we’ll close this I think. Many thanks all.

dekelpilli:
I think it's useful, but I agree, if J11 has it then we shouldn't double up.

mikkomaunu:
Method suggested here serves slightly different use case and cannot always be substituted by `java.lang.Objects#requireNonNullElseGet`. As 'requireNonNull' in name hints, `requireNonNullElseGet` will throw NullPointerException if Supplier that returns null is utilized.

PR #409, opened 2019-03-03
Consolidate the StringUtils.equals and equalsIgnoreCase methods.

aherbert:
Implement the same edge case logic for null. Use the same parameter
names. Use the same Javadoc wording. Change the equals method to use a
step-wise charAt comparison.


kinow:
(Ah! We might need a JIRA for this, to include in changes.xml later too)

aherbert:
This is now documented in [LANG-1436](https://issues.apache.org/jira/browse/LANG-1436).

garydgregory:
Squashing and merging.

PR #410, opened 2019-03-03
Add proposal for Jenkins Pipeline

britter:


britter:
I'm not sure whether we should have fine grained step definitions for jenkins or just call the default goal like we do on Travis CI.

Jenkinsfile
```diff
@@ -43,11 +43,7 @@ pipeline {
         }
         stage('Analyze') {
             steps {
-                sh 'mvn apache-rat:check'
-                sh 'mvn clirr:check'
-                sh 'mvn checkstyle:check'
-                sh 'mvn spotbugs:check'
-                sh 'mvn javadoc:javadoc'
+                sh 'mvn apache-rat:check clirr:check checkstyle:check spotbugs:check javadoc:javadoc'
```
PascalSchumacher: What about just calling `mvn` which does pretty much the same, see: https://github.com/apache/commons-lang/blob/master/pom.xml#L627 ?

kinow: Might be a good idea. I think in Travis we have that in some other components too... that way once updated, we won't have to touch the Jenkinsfile (which I'm less familiar than pom.xml). Or would there be some advantage in maintaining the build maven goals in the Jenkinsfile instead?

britter: Okay, let's do it like we do it for travis. In this case I would also combine the build and analyze stages into one stage.

britter: @kinow please take another look. Can we merge the pipeline script like this?

kinow: +1 LGTM

ParanoidUser:
Ahoy! Just checking to see if any of the maintainers are still actively using Jenkins to build the library? It feels like it has been overtaken by GitHub Actions as the main CI platform for the past few years. It may be reasonable to drop this pipeline to keep the project tidy and clean.

kinow:
Hi. I am not using Jenkins at the moment. I know some projects still use it as the main CI (e.g. Jena), even when they have GitHub Actions due to being able to customize ASF Jenkins jobs more (e.g. if you need another VM, download test data as OpenNLP does for models, etc.).

In the case of Commons Lang specifically, it's strange we have the `Jenkinsfile` here but I couldn't find the job in Jenkins for this repo: https://ci-builds.apache.org/job/Commons/

Maybe others are more up to date with the ASF Jenkins builds…

garydgregory:
I do not look at Jenkins, it fails too often with odd git errors and the UI is a bit sluggish. If someone cares about Jenkins builds, they should step up and maintain them. I am happy using GitHub Actions for builds.

kinow:
So I think it'd make sense to remove the Jenkinsfile from lang, @garydgregory ?

garydgregory:
@kinow 
Yes, I'll drop it.

PR #413, opened 2019-03-12
Fix javadoc in wrapIfMissing

machaval:
wrapIfMissing function was referencing warp function in the javadoc


kinow:
Good spot! Merged, thanks @machaval !

PR #414, opened 2019-03-19
LANG-1369: Formatted and Paramaterized Exception Classes

BELUGABEHR:


jochenw:
Would you please be so kind, and suggest a use case, which your pr is going to address. Cause, I don't see it immediately.

Also a suggestion: Out in the wild, there are quite a number of Pattern format styles, including, for example

    Hi, guys, my name is {}.

    Hi, guys, my name is %s.

    Hi, guys, my name is %1.

(Sure, a bit of digging would reveal others.) So, I would suggest, that

1.  You replace the PatternFormatter with an interface, and a default implementation.
2.  You replace the static ExceptionFactory with instance methods, that use a configurable instance. The static methods might be reimplemented using an instance of the default implementation.

PR #415, opened 2019-04-05
LANG-1443: Add more SystemUtils.IS_JAVA_XX variants

britter:


src/test/java/org/apache/commons/lang3/JavaVersionTest.java
```diff
@@ -20,23 +20,10 @@
 
 import org.junit.jupiter.api.Test;
 
-import static org.apache.commons.lang3.JavaVersion.JAVA_RECENT;
+import static org.apache.commons.lang3.JavaVersion.*;
```
kinow: Checkstyle ain't happy with this .* import I think.

garydgregory: Right, we need specific imports.

chtompki: IntelliJ does that by default at greater than 5 import statements. It rewrites them with the *

britter: Damn you, IntelliJ! 😤Will fix that later today.

britter:
Fixed the Checkstyle problems.

PR #416, opened 2019-04-09
[LANG-1498] Add support of lambda value evaluation for defaulting methods

Lysergid:
https://issues.apache.org/jira/browse/LANG-1498


garydgregory:
I am really not a fan of the "lazy" method name prefix. It does not convey the correct meaning of the method. Passing in a Supplier is a big hint as it is, proper Javadoc will take care of the details.

Lysergid:
Removing prefix will cause "ambiguous method call" when second parameter is null. Which will force clients to cast.
It is minor but still breaks backward compatibility.

garydgregory:
This will not break binary compatibility. We are usually OK with breaking source compatibility, which I am in favor of doing instead of adding APIs with weird (to me) names.

Lysergid:
Not sure if this is better now. Compiler cannot derive type from lambda so I have to explicitly declare type or cast as you can see in tests.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -1491,12 +1491,12 @@ public static int countMatches(final CharSequence str, final CharSequence sub) {
      * <p>Caller responsible for thread-safety and exception handling of default value supplier</p>
      *
      * <pre>
-     * StringUtils.lazyDefaultIfBlank(null, () -&gt; "NULL")   = "NULL"
-     * StringUtils.lazyDefaultIfBlank("", () -&gt; "NULL")     = "NULL"
-     * StringUtils.lazyDefaultIfBlank(" ", () -&gt; "NULL")    = "NULL"
-     * StringUtils.lazyDefaultIfBlank("bat", () -&gt; "NULL")  = "bat"
-     * StringUtils.lazyDefaultIfBlank("", () -&gt; null)       = null
-     * StringUtils.lazyDefaultIfBlank("", null)                = null
+     * StringUtils.defaultIfBlank(null, () -&gt; "NULL")   = "NULL"
```
verhas: You know that you can combine `<pre>` and `{@code` to avoid using the ugly HTML escape sequences. I learned that from the JDK source.

verhas:
>Removing prefix will cause "ambiguous method call" when second parameter is null.

My 2c is `getIfBlank()` if you are keen on keeping source compatibility.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -676,45 +677,45 @@ public void testDefaultIfBlank_StringString() {
...
-    public void testDefaultIfBlank_StringStringSupplier() {
-        assertEquals("NULL", StringUtils.<String>defaultIfBlank(null, () -> "NULL"));
-        assertEquals("NULL", StringUtils.<String>defaultIfBlank("", () -> "NULL"));
-        assertEquals("NULL", StringUtils.<String>defaultIfBlank(" ", () -> "NULL"));
-        assertEquals("abc", StringUtils.<String>defaultIfBlank("abc", () -> "NULL"));
-        assertNull(StringUtils.<String>defaultIfBlank("", () -> null));
-        assertNull(StringUtils.defaultIfBlank("", (Supplier<String>) null));
+    public void testGetIfBlank_StringStringSupplier() {
+        assertEquals("NULL", StringUtils.getIfBlank(null, () -> "NULL"));
+        assertEquals("NULL", StringUtils.getIfBlank("",  () -> "NULL"));
```
verhas: There is an extra space on the line.

src/main/java/org/apache/commons/lang3/StringUtils.java
Lysergid: Thanks for hint

garydgregory:
I applied a this functionality to git master, but not directly from here. Please review and close this PR if it works for you.

garydgregory:
The feature exists through https://issues.apache.org/jira/browse/LANG-1498, closing here.

PR #418, opened 2019-04-13
Changing switch statement to if/else statement for greater clarity

ethanwood17:
Switch statement with only 2 branches seems unnecessary unless future additions are planned. If not, converting to if/else seems clearer.


chtompki:
I’ll pull this in, tomorrow morning.

PR #419, opened 2019-05-03
Lang 1456 either

jangroothuijse:
Stores a single value, either of type `L`, or of type `R`. Fits in between `Stream` and `Optional` in terms of applying transformations conditionally (and safely).
An `Either` implementation is available in the standard library of both haskell and scala.

[https://issues.apache.org/jira/browse/LANG-1456](https://issues.apache.org/jira/browse/LANG-1456)


kinow:
It would be a good fit for Commons Functor if we were close to a release. Not sure what others thing about restarting work there and adding some interesting code like this `Either`.

garydgregory:
Closing: Old PR, won't do, but is something that could be considered for `Pair`.

PR #420, opened 2019-05-08
LANG-1453: using toUpperCase instead of toLowerCase solve the problem

geratorres:
I changed the using of toLowerCase to toUpperCase on replace function on StringUtils. more detailed info is in a comment on Jira

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -2688,8 +2688,8 @@ public class StringUtilsTest {
         // StringUtils.removeIgnoreCase("queued", "zZ") = "queued"
         assertEquals("queued", StringUtils.removeIgnoreCase("queued", "zZ"));
 
-        // StringUtils.removeIgnoreCase("?a", "a") = "?"
-        assertEquals("?", StringUtils.removeIgnoreCase("?a", "a"));
+        // StringUtils.removeIgnoreCase("İa", "a") = "?"
+        assertEquals("İ", StringUtils.removeIgnoreCase("İa", "a"));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


cesartxt:
Please, review the conversation of the pull request #340

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -2687,6 +2687,9 @@ public void testRemoveIgnoreCase_String() {
 
         // StringUtils.removeIgnoreCase("queued", "zZ") = "queued"
         assertEquals("queued", StringUtils.removeIgnoreCase("queued", "zZ"));
+
+        // StringUtils.removeIgnoreCase("İa", "a") = "İ"
+        assertEquals("İ", StringUtils.removeIgnoreCase("İa", "a"));
```
chtompki: We need to minimally replace `"İ"` with `"\u0130"` for this to properly work. For example try running:

```java
System.out.println("İ");
System.out.println("\u0130");
```
only the second will give your desired output. That said, your point is still quite valid.

chtompki:
I agree with @cesartxt - it seems that this resolves the unit tests as written for `StringUtils`, but the question becomes are there similar lower case letters that would yield a similar result when ran through `toUpperCase`.

chtompki:
An example of another character that has this problem from the lowercase alphabet is `ŉ` or `\u0149`. So the question becomes how do we accommodate for these oddities? I'm open to thoughts here

garydgregory:
Some background: https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/

chtompki:
Another non-deprecated Unicode character that has this  `IndexOutOfBoundsException`  as well is `\uFB01` or `ﬁ`.

and 

yes @garydgregory there is a point here as I read an article that proported that there was at least one Turkish loss of life due to a cell phone mistranslating the specific character in question in the proposed tests for LANG-1453

geratorres:
Thanks! for the context with this new info I'll be trying to find a good solution

chtompki:
Yeah. I fiddled with it some today and didn’t see any clever solution that wasn’t brute force. I think we may have to check both situations.

geratorres:
Closing this pull request because doesn't resolve the problem and the issue is duplicated with LANG-1406

PR #421, opened 2019-05-09
revise java-doc error

ScaldingBlood:
Revise the java-doc of splitPreserveAllTokens(final String str, final String separatorChars, final int max)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3622,8 +3622,8 @@ public class StringUtils {
      * <pre>
      * StringUtils.splitPreserveAllTokens(null, *, *)            = null
      * StringUtils.splitPreserveAllTokens("", *, *)              = []
-     * StringUtils.splitPreserveAllTokens("ab de fg", null, 0)   = ["ab", "cd", "ef"]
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3622,8 +3622,8 @@ public static String substringBetween(final String str, final String open, final
      * <pre>
      * StringUtils.splitPreserveAllTokens(null, *, *)            = null
      * StringUtils.splitPreserveAllTokens("", *, *)              = []
-     * StringUtils.splitPreserveAllTokens("ab de fg", null, 0)   = ["ab", "cd", "ef"]
-     * StringUtils.splitPreserveAllTokens("ab   de fg", null, 0) = ["ab", "cd", "ef"]
+     * StringUtils.splitPreserveAllTokens("ab de fg", null, 0)   = ["ab", "de", "fg"]
+     * StringUtils.splitPreserveAllTokens("ab   de fg", null, 0) = ["ab", ,"", "de", "fg"]
```
kinow: The first one matches what I got using the `master` branch. But the second example has a problem with the quotes? Supposed to be `["ab", "", "", "de", "fg"]` I think.

ScaldingBlood: Yep, should be that.

kinow:
Perfect! Thanks for the contribution and for fixing it so quickly! Merged.

PR #381, opened 2018-11-15
LANG-1406 StringIndexOutOfBoundsException in StringUtils.replaceIgnor…

drajakumar:
…eCase

PR #422, opened 2019-05-14
LANG-1406: avoid StringIndexOutOfBounds exceptions for some cases of …

geratorres:
…StringUtils.replaceIgnoreCase.
creates new method indexOf(String, String, int, boolean) and changes some logic of indexOfIgnoreCase

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1471,6 +1471,11 @@ public class StringUtils {
         return CharSequenceUtils.indexOf(seq, searchSeq, startPos);
     }
 
+    private static int indexOf(String str, String searchStr, int fromIndex, boolean ignoreCase) {
+        return ignoreCase ? indexOfIgnoreCase(str, searchStr, fromIndex)
+                : indexOf(str, searchStr, fromIndex);
+    }
+
     /**
      * <p>Finds the n-th index within a CharSequence, handling {@code null}.
      * This method uses {@link String#indexOf(String)} if possible.</p>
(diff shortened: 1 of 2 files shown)


garydgregory:
See error at the end of https://travis-ci.org/apache/commons-lang/jobs/532071438

geratorres:
Thanks!, I ran tests with "mvn clean verify" and didn't get any error, but I see in  https://travis-ci.org/apache/commons-lang/jobs/532071438 is used only "mvn", I fixed now the Bug/Error. I will push it in a moment.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -5624,12 +5638,17 @@ private static String replace(final String text, String searchString, final Stri
...
                  break;
              }
-             end = searchText.indexOf(searchString, start);
+             end = indexOf(text, searchString, start, ignoreCase);
          }
          buf.append(text, start, text.length());
          return buf.toString();
      }
 
+     private static int indexOf(String str, String searchStr, int fromIndex, boolean ignoreCase) {
```
geratorres: I will move this method below indexOf(CharSequence, CharSequence, int) declaration around line 1473

geratorres: java doc suggestions are welcome

chtompki:
We have a variety of pull requests attempting to address [LANG-1406](https://issues.apache.org/jira/browse/LANG-1406), namely: this pull request, #340, #381, and your other one at #420. @geratorres have you looked at #340 and #381 yet? I think that #381 might be closer, but I'm not sure.

geratorres:
> We have a variety of pull requests attempting to address [LANG-1406](https://issues.apache.org/jira/browse/LANG-1406), namely: this pull request, #340, #381, and your other one at #420. @geratorres have you looked at #340 and #381 yet? I think that #381 might be closer, but I'm not sure.

yes, I already checked those pull request, I think that this pull request doesn't look very well  because I also tried to resolve another problem I detected, I have another commit that looks better than this and the other pull requests, this new commit is focused to resolve only the issue 1406, it has minimal changes.  I'm going to close this pull request, delete the fork and create a new pull request, sorry for the bunch of pull requests.

PR #424, opened 2019-05-20
Annotated packages exception, builder, reflect and partially time. 

Kakarot-SSJ4:
Annotated and suppressed warnings to remove errors completely from the following packages:

- Exception 

- Builder

- Reflect

- Time(only the following classes)-
      FastTimeZone.java
      DateUtils.java
      FastDatePrinter.java


garydgregory:
You should discuss why we want this vast a change to the component on the development mailing list. As of now, this is way too many changes for one PR. I am not sure if we will even agree on decorating sources all over the place with annotations. You should start a discussion in the ML...

PR #425, opened 2019-05-26
RandomUtils : comment error

shenqioa:
Modify the error comment


src/main/java/org/apache/commons/lang3/RandomUtils.java
```diff
@@ -166,23 +166,23 @@ public static long nextLong() {
      *
      * @param startInclusive
      *            the smallest value that can be returned, must be non-negative
-     * @param endInclusive
-     *            the upper bound (included)
+     * @param endExclusive
```
kinow: Ah, got it. It appeared to be inclusive, but looking at the `RandomUtils` implementation, it uses only `java.util.Random` (constant, not parametrized). And the methods used here, e.g. `.nextDouble()` is [inclusive, exclusive).

PR #427, opened 2019-06-01
LANG-1460: (doc) (trivial) correct release year for v3.9 in release-history.xml

larrywest42:
For v3.9, it had 2018-04-09, which should be 2019-04-09.

_(I missed the note about not requiring a JIRA ticket for such trivial changes.)_


kinow:
Spent some time looking for any documentation about `release-history.xml`, found the commits that changed this file, and look to be done manually. Not sure if a script is used to generate the file or not, but looks to be a manual process for now.

In any case, looks harmless, let's fix the date :+1: 

Thanks @larrywest42 !

PR #429, opened 2019-06-03
ArrayUtils: merge 8 isSameLength,isEmpty,isNotEmpty into one function each.

XenoAmess:
There are 8 isSameLength,8 isEmpty,8 isNotEmpty in ArrayUtils, but only one getLength function.
That is somehow weird, and breaks compatibility.
And also, if we make such change, we can apply isSameLength upon different kind of arrays.

Otherwise, we shall make functions like getLength be 8 functions to keep compatibility in my opinion.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -1269,111 +1269,7 @@ public class ArrayUtils {
      * @return {@code true} if length of arrays matches, treating
      *  {@code null} as an empty array
      */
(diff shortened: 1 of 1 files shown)


XenoAmess:
> Wouldn't it allow for `ArrayUtils.isSameLength("this is a test", "a test this is")`? Which would actually result in `IllegalArgumentException` in `getLength`, but could be misleading.
> 
> IMHO, the current way is better, or if someone could come up with a simplification using generics, that could work too. But using `Object` changes the semantics of the method - and not exactly in an intuitive way.

Then I wonder, why ArrayUtils.getLength accept an Object as param then?
Shouldn't it become 8 functions just like isSameLength?

PR #431, opened 2019-06-12
adding power support

ghatwala:
enabling power builds for this repository via travis jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 language: java
+os:
+  - linux
+  - linux-ppc64le
 
 jdk:
   - oraclejdk8


garydgregory:
How does testing Linux on Power test anything different for a given component compared to Linux on any other hardware?

ghatwala:
> How does testing Linux on Power test anything different for a given component compared to Linux on any other hardware?

@garydgregory - Power(ppc64le) is a different architecture than the x86(intel) architecure . So it would be good to check for code regression on this . We have observed that many a times even though both are little endian arch's there are build/test case failures observed for any code on power which pass successfully on intel.

kinow:
Conflict in `.travis.yml` now

ghatwala:
@kinow - thanks for the update ..but this is still open PR , merged PR was this one - #435 .
@lysannef  - could you please check this  ?

PR #434, opened 2019-07-05
Fixed a problem when parsing Chinese using the ToStringStyle.JSON_STYLE

ch0ice:
When I write the following code
```java
//org.apache.commons.lang3.builder.ReflectionToStringBuilderTest
public static class TestJsonStyle{
        private String a;
        public void setA(String a){
            this.a = a;
        }
        @Override
        public String toString() {
            return ReflectionToStringBuilder.toString(this, ToStringStyle.JSON_STYLE);
        }
    }

    @Test
    public void testJsonStyle() {
        TestJsonStyle jsonStyle = new TestJsonStyle();
        jsonStyle.setA("测试");
        Assertions.assertEquals("{\"a\":\"测试\"}",jsonStyle.toString());
    }
```
You get the following result，I don't think this is correct
```java
{"a":"\u6D4B\u8BD5"}
```

When I change the following code, the result works fine
```java
//org.apache.commons.lang3.builder.ToStringStyle
private void appendValueAsString(final StringBuffer buffer, final String value) {
//            buffer.append('"').append(StringEscapeUtils.escapeJson(value)).append('"');
            buffer.append('"').append(value).append('"');
}
//{"a":"测试"}
```

But when I did the unit test, I found out
An error occurred on this unit test
When I looked at the test case, I thought there was something wrong with the test case itself
```java
//org.apache.commons.lang3.builder.JsonToStringStyleTest
@Test
    public void testLANG1395() {

//        assertEquals("{\"name\":\"value\"}", new ToStringBuilder(base).append("name", "value").toString());
//        assertEquals("{\"name\":\"\"}", new ToStringBuilder(base).append("name", "").toString());
//        assertEquals("{\"name\":\"\\\"\"}", new ToStringBuilder(base).append("name", '"').toString());
//        assertEquals("{\"name\":\"\\\\\"}", new ToStringBuilder(base).append("name", '\\').toString());
//        assertEquals("{\"name\":\"Let's \\\"quote\\\" this\"}", new ToStringBuilder(base).append("name", "Let's \"quote\" this").toString());

        assertEquals("{\"name\":\"value\"}", new ToStringBuilder(base).append("name", "value").toString());
        assertEquals("{\"name\":\"\"}", new ToStringBuilder(base).append("name", "").toString());
        assertEquals("{\"name\":\"\"\"}", new ToStringBuilder(base).append("name", '"').toString());
        assertEquals("{\"name\":\"\\\"}", new ToStringBuilder(base).append("name", '\\').toString());
        assertEquals("{\"name\":\"Let's \"quote\" this\"}", new ToStringBuilder(base).append("name", "Let's \"quote\" this").toString());

    }
```

Or when I write it this way
```java
System.out.println(new ToStringBuilder(base).append("name", '"').toString());
//r1(Before fix)      {"name":"\""}
//r2(After fix)      {"name":"""} 
```
I think the right result is r2 instead of r1

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
@@ -2608,7 +2608,8 @@ public abstract class ToStringStyle implements Serializable {
          * @param value the value to append.
          */
         private void appendValueAsString(final StringBuffer buffer, final String value) {
-            buffer.append('"').append(StringEscapeUtils.escapeJson(value)).append('"');
+//            buffer.append('"').append(StringEscapeUtils.escapeJson(value)).append('"');
+            buffer.append('"').append(value).append('"');
         }
 
         @Override
--- a/src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/JsonToStringStyleTest.java
@@ -450,11 +450,19 @@ public class JsonToStringStyleTest {
 
     @Test
     public void testLANG1395() {
+
+//        assertEquals("{\"name\":\"value\"}", new ToStringBuilder(base).append("name", "value").toString());
+//        assertEquals("{\"name\":\"\"}", new ToStringBuilder(base).append("name", "").toString());
+//        assertEquals("{\"name\":\"\\\"\"}", new ToStringBuilder(base).append("name", '"').toString());
+//        assertEquals("{\"name\":\"\\\\\"}", new ToStringBuilder(base).append("name", '\\').toString());
+//        assertEquals("{\"name\":\"Let's \\\"quote\\\" this\"}", new ToStringBuilder(base).append("name", "Let's \"quote\" this").toString());
+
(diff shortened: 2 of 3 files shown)


kinow:
I don't feel confident enough to vote +1 on this change. It seems to fix the bug described here, and the test is passing on Travis. But I am concerned about the change of removing the code that seems to escape the JSON value. It probably had a reason for being used I guess.

But will take a better look later when I have more time to try the code and read code and commit history. Thanks for the pull request!

Bruno

src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -2608,7 +2608,7 @@ private boolean isJsonObject(final String valueAsString) {
          * @param value the value to append.
          */
         private void appendValueAsString(final StringBuffer buffer, final String value) {
-            buffer.append('"').append(StringEscapeUtils.escapeJson(value)).append('"');
+            buffer.append('"').append(value).append('"');
```
Stzx: Directly writing characters causes JSON special characters to not be properly escaped.

ch0ice: so how to solve this problem.

PR #439, opened 2019-08-17
Improvement JavaVersion get method

Stzx:
When nom is null, return directly without unnecessary judgment.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/JavaVersion.java
+++ b/src/main/java/org/apache/commons/lang3/JavaVersion.java
@@ -194,7 +194,9 @@ public enum JavaVersion {
      * version is unknown
      */
     static JavaVersion get(final String nom) {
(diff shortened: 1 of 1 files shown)


kinow:
And I think this is such a small change that it may not require a JIRA/changelog?

Stzx:
I also think this is a small change, so it was not created in JIRA.

Stzx:
@kinow I want to know what conditions are required to merge PR?

kinow:
Hi @Stzx, no other I believe. But as there are other commons developers, it is common to wait a few hours/days so others can have a look too. If there are no objections here I will merge it Sunday/Monday (NZ time) when I should have time to triage some issues and look at the PR's again :+1:

Stzx:
@kinow Thank you for your answer. 👍

PR #440, opened 2019-08-19
Fix static code issues

ham1:
A few changes based upon static analysis issues.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -107,7 +107,7 @@ Changes in this version include:
 
 New features:
 o LANG-1352: EnumUtils.getEnumIgnoreCase and isValidEnumIgnoreCase methods added Thanks to Ruslan Sibgatullin.
-o LANG-1372: Add ToStringSummary annotation Thanks to Sérgio Ozaki.
+o LANG-1372: Add ToStringSummary annotation Thanks to SÃ©rgio Ozaki.
 o LANG-1356: Add bypass option for classes to recursive and reflective EqualsBuilder Thanks to Yathos UG.
 o LANG-1391: Improve Javadoc for StringUtils.isAnyEmpty(null) Thanks to Sauro Matulli, Oleg Chubaryov.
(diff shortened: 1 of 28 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -6378,8 +6378,8 @@ private static String replace(final String text, String searchString, final Stri
          }
          final int replLength = searchString.length();
          int increase = replacement.length() - replLength;
-         increase = increase < 0 ? 0 : increase;
-         increase *= max < 0 ? 16 : max > 64 ? 64 : max;
+         increase = Math.max(increase, 0);
+         increase *= Math.min(max, 0);
```
garydgregory: How can this be correct when the bounds above are 16 and 64?

ham1: That is a very good spot, not sure what happened here. I've fixed it now. Interesting that no tests picked it up.

RELEASE-NOTES.txt
```diff
@@ -983,7 +983,7 @@ o LANG-462:  FastDateFormat supports parse methods.
 BUG FIXES
 ===========
 
-o LANG-932:  Spelling fixes. Thanks to Ville Skytt�.
+o LANG-932:  Spelling fixes. Thanks to Ville Skyttä.
```
kinow: Interesting issue with these special characters. But I think this file is generated automatically by a script during release, so it may be overridden later I guess? Someone else more familiar with the release process of lang can comment here I think

Stzx: I checked it and found that it should be because the project default encoding is ISO-8859-1. Github seems to use UTF-8 encoding by default to read the content and cause garbled characters.

jochenw:
I'd be ready to accept these one by one, but not 15 files in one go.

ham1:
Could you explain your rational and what should I do differently next time to make the review easier?

This PR is 86 lines of changes broken down into differet commits for each type of change. Should I create a PR for each file or each commit?

PR #441, opened 2019-08-20
[LANG-1475] Fix unwrap StringIndexOutOfBoundsException

Stzx:
When the string length is shorter than two, it should be returned directly without operation.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -54,6 +54,7 @@ The <action> type attribute can be add,update,fix,remove.
     <action issue="LANG-1467" type="add" dev="ggregory">Add zero arg constructor for org.apache.commons.lang3.NotImplementedException.</action>
     <action issue="LANG-1470" type="add" dev="ggregory">Add ArrayUtils.addFirst() methods.</action>
     <action issue="LANG-1437" type="update" dev="ggregory" due-to="Andrei Troie">Remove redundant if statements in join methods #411.</action>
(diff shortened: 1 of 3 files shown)


Stzx:
It seems that the file I did not modify caused the build to fail. 😂

```
[ERROR] src/main/java/org/apache/commons/lang3/Range.java:[29] (regexp) RegexpSingleline: Line has trailing spaces.
[ERROR] src/main/java/org/apache/commons/lang3/Range.java:[30] (regexp) RegexpSingleline: Line has trailing spaces.
```

src/changes/changes.xml
```diff
@@ -61,6 +61,7 @@ The <action> type attribute can be add,update,fix,remove.
     <action                   type="update" dev="ggregory">checkstyle.version 8.18 -> 8.23.</action>
     <action                   type="update" dev="ggregory">junit-jupiter 5.5.0 -> 5.5.1.</action>
     <action issue="LANG-1477" type="add" dev="jochen">Added Functions.as*, and tests thereof, as suggested by Peter Verhas</action>
+    <action issue="LANG-1475" type="fix" dev="stzx">StringUtils.unwrap incorrect throw StringIndexOutOfBoundsException.</action>
```
kinow: Normally the `dev` is a committer, and `due-to` is the PR contributor name/alias. But I will amend this manually in my working copy while merging :+1:

Stzx: Ok, thank you for your work for this.👍

PR #442, opened 2019-08-20
LANG-1474 documentation correction

verhas:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3600,7 +3600,7 @@ public static boolean isNoneBlank(final CharSequence... css) {
      * <p>Checks if none of the CharSequences are empty ("") or null.</p>
      *
      * <pre>
-     * StringUtils.isNoneEmpty((String) null)    = false
+     * StringUtils.isNoneEmpty((String) null)    = true
```
kinow: Same as this one, `false` not `true`.

Stzx:
+1

kinow:
@Stzx do you think the documentation changes are correct, or did you get the same as I did testing it? (sorry, couldn't tell if +1 was for my comment or for the PR change :)

Stzx:
@kinow Sorry.
Make you feel confused.
The conclusion of my test is the same as yours.

kinow:
Thanks for testing, it was my bad too for not understanding it. Let's wait @verhas to comment now :+1:

kinow:
@verhas tested again today, and same behavior in master. Closing the PR, but in case you can confirm it's indeed an issue, feel free to re-open and reply with some step to confirm/reproduce it. Thanks!

PR #443, opened 2019-08-21
Optimize string split methods: 1. Use ThreadLocal to make reuse of th…

YuyuZha0:
####Optimize string split methods:

-  In current impl of split methods, a string array list is created every time the method called, this should increase gc burden， also, if a string contains too many segments, it should contains several array resizing during the method call. Use ThreadLocal to make reuse of ArrayList and avoid resizing, this trick can also be found on JDK: see java.math.BigDecimal#toString();

```

// Private class to build a string representation for BigDecimal object.
    // "StringBuilderHelper" is constructed as a thread local variable so it is
    // thread safe. The StringBuilder field acts as a buffer to hold the temporary
    // representation of BigDecimal. The cmpCharArray holds all the characters for
    // the compact representation of BigDecimal (except for '-' sign' if it is
    // negative) if its intCompact field is not INFLATED. It is shared by all
    // calls to toString() and its variants in that particular thread.
    static class StringBuilderHelper {
        final StringBuilder sb;    // Placeholder for BigDecimal string
        final char[] cmpCharArray; // character array to place the intCompact

        StringBuilderHelper() {
            sb = new StringBuilder();
            // All non negative longs can be made to fit into 19 character array.
            cmpCharArray = new char[19];
        }

        // Accessors.
        StringBuilder getStringBuilder() {
            sb.setLength(0);
            return sb;
        }
       ...

```

-  From JetBrains Intellij Idea inspection:

> There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
> 
> In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
> 
> This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

See more:[https://shipilev.net/blog/2016/arrays-wisdom-ancients/](https://shipilev.net/blog/2016/arrays-wisdom-ancients/)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -19,7 +19,12 @@ package org.apache.commons.lang3;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.text.Normalizer;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 /**
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -7981,12 +7981,16 @@ public class StringUtils {
         }
     };
 
-    // Private class act as a buffer while splitting.
-    // "SplitBufferThreadLocalHelper" is constructed as a thread local variable so it is
-    // thread safe. The "splitBuffer" field acts as a buffer to hold the temporary
-    // representation of string split segments. It is shared by all
-    // calls to splitByCharacterType() or splitByWholeSeparatorWorker()
-    // or splitWorker(String) or splitWorker(char) and its variants in that particular thread.
+
+
+    /**
+     * Private class act as a buffer while splitting.
+     * "SplitBufferThreadLocalHelper" is constructed as a thread local variable so it is
+     * thread safe. The "splitBuffer" field acts as a buffer to hold the temporary
+     * representation of string split segments. It is shared by all
+     * calls to {@link #splitByCharacterType(String, boolean)} or {@link #splitByWholeSeparatorWorker(String, String, int, boolean)}
+     * or {@link #splitWorker(String, char, boolean)} or {@link #split(String, String, int)} and its variants in that particular thread.
+     */
     private static final class SplitBufferThreadLocalHelper {
 
         private final SplitBuffer splitBuffer = new SplitBuffer();
@@ -7997,7 +8001,18 @@ public class StringUtils {
         }
     }
 
-    //buffer class
+    /**
+     * A buffer class to hold split segments,
+     * this class is hold by the {@link ThreadLocal} so it is thread-safe.<br>
+     * While splitting, the segments will be add to the tail of the array with
+     * {@link SplitBuffer#add(String)} method(If the capacity of the array is not enough, it will enable auto-expanding).<br>
+     * {@link SplitBuffer#toArray()} method will copy current segments to a single String array.<br>
+     * {@link SplitBuffer#reset()} method will set the length of the array to 0, however, the "elementData" array will be reused next time.<br>
+     * This class is designed to replace previous {@link ArrayList}, in previous version, an ArrayList
+     * is created every time the split method is called, It means array allocation every time,if the segments
+     * is large, it may also contains several resizing desiged by ArrayList.<br>
+     * The mainly purpose of this class is to improve performance.
+     */
     private static final class SplitBuffer {
 
         /**
@@ -8073,7 +8088,7 @@ public class StringUtils {
          * the given minimum capacity is greater than MAX_ARRAY_SIZE.
          *
          * @param minCapacity the desired minimum capacity
-         * @throws OutOfMemoryError if minCapacity is less than zero
+         * @throws RuntimeException if minCapacity is less than zero
          */
         private int newCapacity(int minCapacity) {
             // overflow-conscious code
@@ -8084,7 +8099,7 @@ public class StringUtils {
                     return Math.max(DEFAULT_CAPACITY, minCapacity);
                 }
                 if (minCapacity < 0) { // overflow
-                    throw new OutOfMemoryError();
+                    throw new RuntimeException("capacity overflow!");
                 }
                 return minCapacity;
             }
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8010,7 +8010,7 @@ public class StringUtils {
      * {@link SplitBuffer#reset()} method will set the length of the array to 0, however, the "elementData" array will be reused next time.<br>
      * This class is designed to replace previous {@link ArrayList}, in previous version, an ArrayList
      * is created every time the split method is called, It means array allocation every time,if the segments
-     * is large, it may also contains several resizing desiged by ArrayList.<br>
+     * is large, it may also contains several resizing designed by ArrayList.<br>
      * The mainly purpose of this class is to improve performance.
      */
     private static final class SplitBuffer {
@@ -8040,7 +8040,7 @@ public class StringUtils {
 
         private static int hugeCapacity(int minCapacity) {
             if (minCapacity < 0) {// overflow
-                throw new OutOfMemoryError();
+                throw new RuntimeException("capacity overflow!");
             }
             return (minCapacity > MAX_ARRAY_SIZE)
                     ? Integer.MAX_VALUE
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -19,6 +19,7 @@ package org.apache.commons.lang3;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.text.Normalizer;
+import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
@@ -7434,7 +7435,7 @@ public class StringUtils {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
         final char[] c = str.toCharArray();
-        final List<String> list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
+        final StringSplitHelperList list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
         int tokenStart = 0;
         int currentType = Character.getType(c[tokenStart]);
         for (int pos = tokenStart + 1; pos < c.length; pos++) {
@@ -7455,7 +7456,7 @@ public class StringUtils {
             currentType = type;
         }
         list.add(new String(c, tokenStart, c.length - tokenStart));
-        return list.toArray(new String[0]);
+        return list.toArray();
     }
 
     /**
@@ -7639,7 +7640,7 @@ public class StringUtils {
 
         final int separatorLength = separator.length();
 
-        final ArrayList<String> substrings = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
+        final StringSplitHelperList substrings = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
         int numberOfSubstrings = 0;
         int beg = 0;
         int end = 0;
@@ -7683,7 +7684,7 @@ public class StringUtils {
             }
         }
 
-        return substrings.toArray(new String[0]);
+        return substrings.toArray();
     }
 
     // -----------------------------------------------------------------------
@@ -7850,7 +7851,7 @@ public class StringUtils {
         if (len == 0) {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
-        final List<String> list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
+        final StringSplitHelperList list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
         int i = 0, start = 0;
         boolean match = false;
         boolean lastMatch = false;
@@ -7871,7 +7872,7 @@ public class StringUtils {
         if (match || preserveAllTokens && lastMatch) {
             list.add(str.substring(start, i));
         }
-        return list.toArray(new String[0]);
+        return list.toArray();
     }
 
     /**
@@ -7900,7 +7901,7 @@ public class StringUtils {
         if (len == 0) {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
-        final List<String> list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
+        final StringSplitHelperList list = SPLIT_BUFFER_THREAD_LOCAL.get().getList();
         int sizePlus1 = 1;
         int i = 0, start = 0;
         boolean match = false;
@@ -7970,7 +7971,7 @@ public class StringUtils {
         if (match || preserveAllTokens && lastMatch) {
             list.add(str.substring(start, i));
         }
-        return list.toArray(new String[0]);
+        return list.toArray();
     }
 
     private static final ThreadLocal<SplitBufferThreadLocalHelper> SPLIT_BUFFER_THREAD_LOCAL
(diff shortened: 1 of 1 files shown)


kinow:
Thanks for the pull request @YuyuZha0 and for the explanation & links about the fix. I don't have time for a thorough review now, and found nothing out of ordinary after a cursory review.

While we try to write code that performs well, there is a balance between maintainability and performance. So it will be important to confirm that while this improves performance, it won't be too hard to maintain the code (e.g. a comment on the newly added code explaining why those variables are being created could be helpful to other developers) :+1:

Stzx (commented):
looks great.

The benefits are obvious, no need to redistribute List, may not need to resize the internal size.

I am worried that some problems will still be caused by using the clear method. such as

> The clear() method has to assign null to each (used) slot in the ArrayLists backing array.
>
> The clear() does not resize the backing array to release memory. So if you repeatedly fill and clear a list, it will end up (permanently) using enough memory to represent the largest list that it encounters. In other word, you have increased the memory footprint. You can combat that by calling trimToSize(), but that creates a garbage object, etcetera1.
>
> There are locality and cross-generational issues that could affect performance. When you repeatedly recycle an ArrayList, the object and its backing array are likely to be tenured. That means that:
> > The list objects and the objects representing list elements are likely to be in different areas of the heap, potentially increasing TLB misses and page traffic, especially at GC time.
>
> > Assignment of (young generation) references into the (tenured) list's backing array are likely to incur write barrier overheads ... depending on the GC implementation.
>
> https://stackoverflow.com/a/18371813

@kinow What do you think?

kinow:
Thanks for reviewing it @Stzx ! I will have to sit down with the IDE and read the code with calm to do a proper review, unless another dev beats me to it (no hard feelings :). But I will re-read your comments and take your comment into consideration (really helpful, thanks).

YuyuZha0:
Anyway, thanks for your effort! @kinow @Stzx

YuyuZha0:
On my Mac(Intel Core i5, 2.3GHz, openjdk version "11.0.2"), It brings about 17% of performance improvation:

```
Benchmark                                   Mode  Cnt    Score    Error  Units
StringSplitBenchmark.testCommonsLang3Split  avgt   25  928.749 ± 11.299  ns/op
StringSplitBenchmark.testFastSplitUtils     avgt   25  771.481 ± 19.010  ns/op
```

Here is the test cases:

```
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@BenchmarkMode(Mode.AverageTime)
@Warmup(iterations = 5, time = 5)
@Measurement(iterations = 5, time = 5)
public class StringSplitBenchmark {

    private static final char separator = '@';

    @Benchmark
    public String[] testCommonsLang3Split(StringSupplier stringSupplier) {
        return StringUtils.splitPreserveAllTokens(stringSupplier.get(), separator);
    }

    @Benchmark
    public String[] testFastSplitUtils(StringSupplier stringSupplier) {
        return FastSplitUtils.splitPreserveAllTokens(stringSupplier.get(), separator);
    }


    @State(Scope.Thread)
    public static class StringSupplier implements Supplier<String> {

        private final String[] array;
        private int index = 0;

        public StringSupplier() {

            List<String> list = new ArrayList<>(1000);
            ThreadLocalRandom random = ThreadLocalRandom.current();
            for (int i = 0; i < 1000; i++) {
                String s = StringUtils.join(
                        random.ints(25).toArray(),
                        separator
                );
                list.add(s);
            }
            this.array = list.toArray(new String[0]);
        }

        @Override
        public String get() {
            if (index >= array.length)
                index = 0;
            return array[index++];
        }
    }
}
```

I think this may be helpful to you @kinow

jochenw:
@kinow Strongly opposed to accepting this PR: I don't have studied this in detail, but in my oppinion we'd loose thread safety here. Split methods, and similar simple stuff should always be thread safe, unless there is a **very** good reason.

YuyuZha0:
I know your opinion, but i was just thinking about where we lose the thread safety..., you see, there is a ThreadLocal. @jochenw What I'm warring about is the memory impact...

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7968,9 +7968,145 @@ public static String rotate(final String str, final int shift) {
...
+        private int newCapacity(int minCapacity) {
+            // overflow-conscious code
+            int oldCapacity = elementData.length;
+            int newCapacity = oldCapacity + (oldCapacity >> 1);
+            if (newCapacity - minCapacity <= 0) {
+                if (elementData == DEFAULTCAPACITY_EMPTY_ELEMENTDATA) {
+                    return Math.max(DEFAULT_CAPACITY, minCapacity);
+                }
+                if (minCapacity < 0) { // overflow
+                    throw new OutOfMemoryError();
```
kinow: We shouldn't use this error type, this is expected to be thrown by the JVM

>Thrown when the Java Virtual Machine cannot allocate an object because it is out of memory

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7968,9 +7968,145 @@ public static String rotate(final String str, final int shift) {
...
+        private final SplitBuffer splitBuffer = new SplitBuffer();
+
+        SplitBuffer getBuffer(){
+            splitBuffer.reset();
+            return splitBuffer;
+        }
+    }
+
+    //buffer class
+    private static final class SplitBuffer {
```
kinow: `StringUtils` is a long class, with high maintenance cost already as-is. Adding an extra class increases the maintenance. I would prefer to avoid it if possible. If we really need to add it, we **must** provide good documentation. Reading the name of the class, and the single comment `buffer class` gives too much room for interpretation. IOW, I would have to read the code to understand what this class does, how it works, and whether I should use it or not.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7968,9 +7968,145 @@ public static String rotate(final String str, final int shift) {
...
+            return new SplitBufferThreadLocalHelper();
+        }
+    };
+
+    // Private class act as a buffer while splitting.
+    // "SplitBufferThreadLocalHelper" is constructed as a thread local variable so it is
+    // thread safe. The "splitBuffer" field acts as a buffer to hold the temporary
+    // representation of string split segments. It is shared by all
+    // calls to splitByCharacterType() or splitByWholeSeparatorWorker()
+    // or splitWorker(String) or splitWorker(char) and its variants in that particular thread.
```
kinow: Comments like this are ignored by Javadoc tools and IDE's, and only available when a developer reads the source code.

kinow:
@YuyuZha0 

>On my Mac(Intel Core i5, 2.3GHz, openjdk version "11.0.2"), It brings about 17% of performance improvation:

I believe there are other methods in Commons Lang that could be optimized by removing branches ([e.g. of what I mean](https://stackoverflow.com/a/16369225)), or by using more specialized data structures.

While some users could welcome this optimization, this is not the main goal of Commons Lang (i.e. it is not to offer what is missing in the Java language with the best performance, but yes to provide what is missing in the Java language).

And adding it would increase the code readability and maintenance. We should be able to accommodate performance improvements, but I think this one would make a class already complicated to maintain, a bit harder to maintain.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7434,7 +7434,7 @@ public static String rotate(final String str, final int shift) {
             return ArrayUtils.EMPTY_STRING_ARRAY;
         }
         final char[] c = str.toCharArray();
-        final List<String> list = new ArrayList<>();
+        final SplitBuffer buffer = SPLIT_BUFFER_THREAD_LOCAL.get().getBuffer();
```
kinow: Leaving the variable name `list` would have reduced the changes to review I think?

kinow:
@Stzx 

>I am worried that some problems will still be caused by using the clear method. such as

I think this does not apply here. The instance that was a `List` would now be a `SplitBuffer`, i.e.

```
final SplitBuffer buffer = SPLIT_BUFFER_THREAD_LOCAL.get().getBuffer();
```

https://github.com/apache/commons-lang/pull/443/files#diff-3f4c835875ddc8a211c0272e8be51394R7902-R7903

So `buffer` doesn't need to respect the `List`/`ArrayList` contracts.

kinow:
@jochenw 

>@kinow Strongly opposed to accepting this PR: I don't have studied this in detail, but in my oppinion we'd loose thread safety here. Split methods, and similar simple stuff should always be thread safe, unless there is a very good reason.

It appears to be that the solution proposed by @YuyuZha0 would be thread-safe, though I haven't really tried much of the code to confirm.

From what I understand, the suggestion here is to use a single data structure per thread (see previous comments about a `ThreadLocal`), following the approach used by the JVM's `BigDecimal`.

While it seems like it would indeed improve performance, I also feel like opposing the change but from a maintainability point of view. `StringUtils` has well over 5K lines, and the more we keep adding to it, the harder it gets to maintain.

And there is the risk of missing some other use case and having issues in multi-threaded environments as you suggested - even though we have the ThreadLocal, I haven't checked other variables used, or if there is any other possibility of other issues.

kinow (changes requested):
@YuyuZha0 thanks a lot for your contribution. I spent some time reviewing it, and thinking very hard whether we should include it or not.

But `StringUtils` performs quite well for me, though I don't have a case where I need it to perform under ms/ns for multiple users.

So for maintenance, I would prefer to keep the current approach with `ArrayList`. Said that, there are two changes in this PR. One for the per-thread collection. The other for the `toArray()`, where you included some docs from IntelliJ about this optimization.

I think this second improvement could become a separate issue, much smaller, and simpler to be reviewed. If it has indeed the same behavior as `toArray(new String[0])` **or even better**, then I think we should be able to include an improvement like this in the next release.

Stzx:
> @Stzx
> 
> > I am worried that some problems will still be caused by using the clear method. such as
> 
> I think this does not apply here. The instance that was a `List` would now be a `SplitBuffer`, i.e.
> 
> ```
> final SplitBuffer buffer = SPLIT_BUFFER_THREAD_LOCAL.get().getBuffer();
> ```
> 
> https://github.com/apache/commons-lang/pull/443/files#diff-3f4c835875ddc8a211c0272e8be51394R7902-R7903
> 
> So `buffer` doesn't need to respect the `List`/`ArrayList` contracts.

@kinow 

At the time, the code used in the list, and the subsequent commits were improved, so I was concerned at the time. It seems that many problems have been solved now.

kinow:
>At the time, the code used in the list, and the subsequent commits were improved, so I was concerned at the time. It seems that many problems have been solved now.

Ah, got it! Sorry, I lost track of the changes since my first comment. Thanks for clarifying.

src/main/java/org/apache/commons/lang3/StringUtils.java
YuyuZha0: I will edit this later.

src/main/java/org/apache/commons/lang3/StringUtils.java
YuyuZha0: The SplitBuffer instance is not a List instance anymore, leave it with a name "list" would be misleading I think.

YuyuZha0:
@kinow Thanks for the carefully reviewing ! Nice weekend, isn't it? I will edit the code later follow your advice. Currently I was on the performance, I've tried more cases, here is the result:
```
Benchmark                                   (arrayLen)  Mode  Cnt     Score    Error  Units
StringSplitBenchmark.testCommonsLang3Split          10  avgt   25   499.547 ± 10.716  ns/op
StringSplitBenchmark.testCommonsLang3Split          30  avgt   25  1502.510 ± 16.956  ns/op
StringSplitBenchmark.testCommonsLang3Split          50  avgt   25  2467.303 ± 18.970  ns/op
StringSplitBenchmark.testFastSplitUtils             10  avgt   25   396.252 ±  4.653  ns/op
StringSplitBenchmark.testFastSplitUtils             30  avgt   25  1145.600 ±  5.604  ns/op
StringSplitBenchmark.testFastSplitUtils             50  avgt   25  1885.414 ±  4.121  ns/op
StringSplitBenchmark.testGuavaSplit                 10  avgt   25   565.904 ±  5.483  ns/op
StringSplitBenchmark.testGuavaSplit                 30  avgt   25  1665.049 ± 81.051  ns/op
StringSplitBenchmark.testGuavaSplit                 50  avgt   25  2758.394 ±  7.684  ns/op
``` 
Cases is shown bellow:
```
import com.google.common.base.Splitter;
import org.apache.commons.lang3.StringUtils;
import org.openjdk.jmh.annotations.*;

import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.ThreadLocalRandom;
import java.util.concurrent.TimeUnit;
import java.util.function.Supplier;

/**
*
 * @author zhaoyuyu
 * @since 2019-08-21
 **/
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@BenchmarkMode(Mode.AverageTime)
@Warmup(iterations = 5, time = 5)
@Measurement(iterations = 5, time = 5)
public class StringSplitBenchmark {

    private static final char separator = '@';
    private static final Splitter splitter = Splitter.on(separator);


    @Benchmark
    public String[] testCommonsLang3Split(StringSupplier stringSupplier) {
        return StringUtils.splitPreserveAllTokens(stringSupplier.get(), separator);
    }

    @Benchmark
    public String[] testFastSplitUtils(StringSupplier stringSupplier) {
        return FastSplitUtils.splitPreserveAllTokens(stringSupplier.get(), separator);
    }

    @Benchmark
    public String[] testGuavaSplit(StringSupplier supplier) {
        return splitter.splitToList(supplier.get()).toArray(new String[0]);
    }


    @State(Scope.Thread)
    public static class StringSupplier implements Supplier<String> {

        @Param({"10", "30", "50"})
        private int arrayLen;

        private String[] array;
        private int index = 0;

        @Setup
        public void setup() {

            List<String> list = new ArrayList<>(1000);
            ThreadLocalRandom random = ThreadLocalRandom.current();
            for (int i = 0; i < 1000; i++) {
                String s = StringUtils.join(
                        random.ints(arrayLen).toArray(),
                        separator
                );
                list.add(s);
            }
            this.array = list.toArray(new String[0]);
        }

        @Override
        public String get() {
            if (index >= array.length)
                index = 0;
            return array[index++];
        }
    }
}

```
The reason why I propose this optimization is that sometimes these methods are really under heavily usage(In my case, I use splitPreserveAllTokens for log processing, the method would be called **billions** of times every day). So for me, the performance is really important.

The StringUtils is widely used, any edition must be cautiously, so I fully understand you warring. In computer science, everything would be a trade off, it's really a hard choice.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7968,9 +7968,160 @@ public static String rotate(final String str, final int shift) {
...
+        protected SplitBufferThreadLocalHelper initialValue() {
+            return new SplitBufferThreadLocalHelper();
+        }
+    };
+
+
+
+    /**
+     * Private class act as a buffer while splitting.
+     * "SplitBufferThreadLocalHelper" is constructed as a thread local variable so it is
```
verhas: Yes, it is thread-safe, but as soon as project loom is released thread-safety will not be our concerns. It will not be fiber safe. Using thread-local variables instead of objects is a hack and will bite back as hard as static variables bit our assets when we started to create multi-thread programs.

YuyuZha0: I think not only `ThreadLocal`, but also the behavior of any *mutable object* would be unpredictable if the execution of a method call could be break(i.e. another method takes over the thread during the method call).I guess there might be some "safe-point" where another method could have access to the execution in project loom?

YuyuZha0: > Another relatively major design decision concerns thread locals. Currently, thread-local data is represented by the (Inheritable)ThreadLocal class(es). How do we treat thread-locals in fibers? Crucially, ThreadLocals have two very different uses. One is associating data with a thread context. Fibers will probably need this capability, too. Another is to reduce contention in concurrent data structures with striping. That use abuses ThreadLocal as an approximation of a processor-local (more precisely, a CPU-core-local) construct. With fibers, the two different uses would need to be clearly separated, as now a thread-local over possibly millions of threads (fibers) is not a good approximation of processor-local data at all. This requirement for a more explicit treatment of thread-as-context vs. thread-as-an-approximation-of-processor is not limited to the actual ThreadLocal class, but to any class that maps Thread instances to data for the purpose of striping. If fibers are represented by Threads, then some changes would need to be made to such striped data structures. In any event, it is expected that the addition of fibers would necessitate adding an explicit API for accessing processor identity, whether precisely or approximately.

See:[https://cr.openjdk.java.net/~rpressler/loom/Loom-Proposal.html](https://cr.openjdk.java.net/~rpressler/loom/Loom-Proposal.html)

garydgregory:
Closing: Very old PR.

PR #444, opened 2019-08-21
LANG-1478

verhas:
Fix ClassUtils class name abbreviation


verhas:
branched off from a brach... ops...

PR #445, opened 2019-08-22
LAN-1478 class name abbreviation fix

verhas:
Fix ClassUtils class name abbreviation


kinow:
@Stzx 

>I think I should add javaDoc at the same time, add a new example.

+1, added a review comment for @verhas . And @verhas , thanks a lot for your contributions, we should be able to merge them soon (unless we need to wait for a major release, then I will update JIRA issues accordingly)

kinow:
@verhas I think this and #446 are good to be included in the next release. Just need you to address the review feedback and update both branches. If you could rebase it onto `master` as well, please. I've just fixed the Checkstyle issues, and Travis build should pass for both :+1: 

Thanks
Bruno

PR #446, opened 2019-08-22
LANG-1480 getAbbreviatedName refactored to create appropriate length …

verhas:
…short class names

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -16,8 +16,6 @@
  */
 package org.apache.commons.lang3;
 
-import org.apache.commons.lang3.mutable.MutableObject;
-
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
@@ -29,6 +27,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import org.apache.commons.lang3.mutable.MutableObject;
 
 /**
  * <p>Operates on classes without using reflection.</p>
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -1424,82 +1424,73 @@ public class ClassUtils {
      * @since 3.2
      */
     public static Iterable<Class<?>> hierarchy(final Class<?> type, final Interfaces interfacesBehavior) {
-        final Iterable<Class<?>> classes = new Iterable<Class<?>>() {
+        final Iterable<Class<?>> classes = () -> {
+            final MutableObject<Class<?>> next = new MutableObject<>(type);
+            return new Iterator<Class<?>>() {
 
-            @Override
-            public Iterator<Class<?>> iterator() {
-                final MutableObject<Class<?>> next = new MutableObject<>(type);
-                return new Iterator<Class<?>>() {
+                @Override
+                public boolean hasNext() {
+                    return next.getValue() != null;
+                }
 
-                    @Override
-                    public boolean hasNext() {
-                        return next.getValue() != null;
-                    }
+                @Override
+                public Class<?> next() {
+                    final Class<?> result = next.getValue();
+                    next.setValue(result.getSuperclass());
+                    return result;
+                }
 
-                    @Override
-                    public Class<?> next() {
-                        final Class<?> result = next.getValue();
-                        next.setValue(result.getSuperclass());
-                        return result;
-                    }
+                @Override
+                public void remove() {
+                    throw new UnsupportedOperationException();
+                }
 
-                    @Override
-                    public void remove() {
-                        throw new UnsupportedOperationException();
-                    }
-
-                };
-            }
-
+            };
         };
         if (interfacesBehavior != Interfaces.INCLUDE) {
             return classes;
         }
-        return new Iterable<Class<?>>() {
+        return () -> {
+            final Set<Class<?>> seenInterfaces = new HashSet<>();
+            final Iterator<Class<?>> wrapped = classes.iterator();
 
-            @Override
-            public Iterator<Class<?>> iterator() {
-                final Set<Class<?>> seenInterfaces = new HashSet<>();
-                final Iterator<Class<?>> wrapped = classes.iterator();
+            return new Iterator<Class<?>>() {
+                Iterator<Class<?>> interfaces = Collections.<Class<?>>emptySet().iterator();
 
-                return new Iterator<Class<?>>() {
-                    Iterator<Class<?>> interfaces = Collections.<Class<?>>emptySet().iterator();
+                @Override
+                public boolean hasNext() {
+                    return interfaces.hasNext() || wrapped.hasNext();
+                }
 
-                    @Override
-                    public boolean hasNext() {
-                        return interfaces.hasNext() || wrapped.hasNext();
+                @Override
+                public Class<?> next() {
+                    if (interfaces.hasNext()) {
+                        final Class<?> nextInterface = interfaces.next();
+                        seenInterfaces.add(nextInterface);
+                        return nextInterface;
                     }
+                    final Class<?> nextSuperclass = wrapped.next();
+                    final Set<Class<?>> currentInterfaces = new LinkedHashSet<>();
+                    walkInterfaces(currentInterfaces, nextSuperclass);
+                    interfaces = currentInterfaces.iterator();
+                    return nextSuperclass;
+                }
 
-                    @Override
-                    public Class<?> next() {
-                        if (interfaces.hasNext()) {
-                            final Class<?> nextInterface = interfaces.next();
-                            seenInterfaces.add(nextInterface);
-                            return nextInterface;
+                private void walkInterfaces(final Set<Class<?>> addTo, final Class<?> c) {
+                    for (final Class<?> iface : c.getInterfaces()) {
+                        if (!seenInterfaces.contains(iface)) {
+                            addTo.add(iface);
                         }
-                        final Class<?> nextSuperclass = wrapped.next();
-                        final Set<Class<?>> currentInterfaces = new LinkedHashSet<>();
-                        walkInterfaces(currentInterfaces, nextSuperclass);
-                        interfaces = currentInterfaces.iterator();
-                        return nextSuperclass;
+                        walkInterfaces(addTo, iface);
                     }
+                }
 
-                    private void walkInterfaces(final Set<Class<?>> addTo, final Class<?> c) {
-                        for (final Class<?> iface : c.getInterfaces()) {
-                            if (!seenInterfaces.contains(iface)) {
-                                addTo.add(iface);
-                            }
-                            walkInterfaces(addTo, iface);
-                        }
-                    }
+                @Override
+                public void remove() {
+                    throw new UnsupportedOperationException();
+                }
 
-                    @Override
-                    public void remove() {
-                        throw new UnsupportedOperationException();
-                    }
-
-                };
-            }
+            };
         };
     }
 
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -156,7 +156,7 @@ public class ClassUtils {
      * instance to operate.</p>
      */
     public ClassUtils() {
-         super();
+      super();
     }
 
     // Short class name
@@ -581,8 +581,8 @@ public class ClassUtils {
             }
 
             cls = cls.getSuperclass();
-        }
-    }
+         }
+     }
 
     // Convert list
     // ----------------------------------------------------------------------
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -424,7 +424,14 @@ public class ClassUtils {
      * significant loss of meaning.</p>
      * <p>The abbreviated class name will always include the complete package hierarchy.
      * If enough space is available, rightmost sub-packages will be displayed in full
-     * length.</p>
+     * length. The abbreviated package names will be shortened to a single character.</p>
+     * <p>Only package names are shortened, the class simple name remains untouched. (See examples.)</p>
+     * <p>The result will be longer than the desired length only if all the package names
+     * shortened to a single character plus the class simple name with the separating dots
+     * together are longer than the desired length. In other words, when the class name
+     * cannot be shortened to the desired length without significant loss of meaning.</p>
+     * <p>If the class name can be shortened without significant loss of meaning then
+     * the final length will be at most {@code len} characters.</p>
      *
      * <table>
      * <caption>Examples</caption>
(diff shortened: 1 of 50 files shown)


src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -16,19 +16,11 @@
...
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import org.apache.commons.lang3.mutable.MutableObject;
+import java.util.*;
```
kinow: This won't pass checkstyle I think. You will need to replace the `*` by explicit imports.

See Travis builds failing, this part is due to this change: 

>[ERROR] src/main/java/org/apache/commons/lang3/ClassUtils.java:[23] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - java.util.*.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -178,8 +179,22 @@ public void test_getAbbreviatedName_Class_ZeroLen() {
...
+        assertEquals(".", ClassUtils.getAbbreviatedName(".", 18));
+        assertEquals(".", ClassUtils.getAbbreviatedName(".", 1));
+        assertEquals("..", ClassUtils.getAbbreviatedName("..", 1));
+        assertEquals("...", ClassUtils.getAbbreviatedName("...", 2));
+        assertEquals("...", ClassUtils.getAbbreviatedName("...", 3));
+        assertEquals("java.lang.String", ClassUtils.getAbbreviatedName("java.lang.String", Integer.MAX_VALUE));
+        assertEquals("j.lang.String", ClassUtils.getAbbreviatedName("java.lang.String", "j.lang.String".length()));
+        assertEquals("j.l.String", ClassUtils.getAbbreviatedName("java.lang.String", "j.lang.String".length() - 1));
+        assertEquals("j.l.String", ClassUtils.getAbbreviatedName("java.lang.String", "j.l.String".length()));
+        assertEquals("j.l.String", ClassUtils.getAbbreviatedName("java.lang.String", "j.l.String".length() - 1));
```
kinow: I think you are not testing the bug you found? There is no test for a call to `getAbbreviatedName` with class name as a `String`, and length as the exact length of that `String`.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -433,11 +432,13 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
      * <tr><td>"java.lang.String"</td><td>30</td><td>"java.lang.String"</td></tr>
+     * <tr><td>"org.apache.commons.lang3.ClassUtils"</td><td>18</td><td>"o.a.c.l.ClassUtils"</td></tr>
      * </table>
      * @param className  the className to get the abbreviated name for, may be {@code null}
      * @param len  the desired length of the abbreviated name
-     * @return the abbreviated name or an empty string
-     * @throws IllegalArgumentException if len &lt;= 0
+     * @return the abbreviated name or an empty string if the specified
+     * class name is {@code null} or empty string
+     * @throws IllegalArgumentException if {@code len <= 0}
```
kinow: Thanks for updating the docs!

kinow:
I think you need to include tests for this change. See the bot comment about coverage decreasing, and a comment I've added about the missing test for the bug you described here.

src/main/java/org/apache/commons/lang3/ClassUtils.java
verhas: This was the default IntelliJ setting to use * after more than 4 imports from the same package.

I see no reason why not to use '*' in imports these days with such advanced editors, but in case that is the project standard, I comply. Thanks.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -987,33 +1005,35 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
...
-     * @param className  the class name
+     * @param classLoader the class loader to use to load the class
+     * @param className   the class name
      * @param initialize  whether the class must be initialized
      * @return the class represented by {@code className} using the {@code classLoader}
      * @throws ClassNotFoundException if the class is not found
      */
     public static Class<?> getClass(
-            final ClassLoader classLoader, final String className, final boolean initialize) throws ClassNotFoundException {
+        final ClassLoader classLoader, final String className, final boolean initialize) throws ClassNotFoundException {
```
kinow: I think now the PR has a lot more changes that are not related to LANG-1480. Are you able to undo formatting changes, and leave only the changes for the issue, please?

verhas: I am doing it. Actually, these formattings are clumsy in the original code, and IntelliJ just reformatted automatically. I understand that it is no excuse for creating a messy PR. Sorry.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -433,48 +440,82 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
+     * <tr><td>"org.apache.commons.lang3.ClassUtils"</td><td>18</td><td>"o.a.c.l.ClassUtils"</td></tr>
      * </table>
-     * @param className  the className to get the abbreviated name for, may be {@code null}
-     * @param len  the desired length of the abbreviated name
-     * @return the abbreviated name or an empty string
-     * @throws IllegalArgumentException if len &lt;= 0
+     *
+     * @param className the className to get the abbreviated name for, may be {@code null}
+     * @param len       the desired length of the abbreviated name
+     * @return the abbreviated name or an empty string if the specified
```
garydgregory: Please specify if the result string is always <= len.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -424,7 +424,14 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
      * <p>The abbreviated class name will always include the complete package hierarchy.
      * If enough space is available, rightmost sub-packages will be displayed in full
-     * length.</p>
+     * length. The abbreviated package names will be shortened to a single character.</p>
+     * <p>Only package names are shortened, the class simple name remains untouched. (See examples.)</p>
+     * <p>The result will be longer than the desired length only if all the package names
+     * shortened to a single character plus the class simple name with the separating dots
+     * together are longer than the desired length. In other words, when the class name
+     * cannot be shortened to the desired length without significant loss of meaning.</p>
+     * <p>If the class name can be shortened without significant loss of meaning then
```
garydgregory: I do not think we should talk about "loss of meaning" as to me, abbreviating implies "loss of meaning" by design ;-) Just describe what it does. IMO a list would be better to describe the different use cases.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -433,48 +440,82 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
+     *                       character copied when the current part was
+     *                       copied
+     * @param originalLength the original length of the class full name,
+     *                       which is abbreviated
+     * @param desiredLength  the desired length of the abbreviated class
+     *                       name
+     * @return {@code true} if it can be kept in its original length
+     * {@code false} if the current part has to be abbreviated and
+     */
+    private static boolean useFull(final int runAheadTarget,
```
garydgregory: Is this only used once? If so, just in-line it with a `//` comment.

verhas: Is this a strong requirement or something you can live with if I leave it as it is?

As you can see from the length of the comment this is a complex expression and the JavaDoc explains a lot about the algorithm used. Such an extensive commend would degrade the readability of the caller method.

As per the performance: the JIT will inline this method, it should not be a problem.

src/main/java/org/apache/commons/lang3/ClassUtils.java
verhas: It is not and it is described in details in the above text.

src/main/java/org/apache/commons/lang3/ClassUtils.java
garydgregory: I'm ok to leave as it since the description is nice especially using `@param` tags which would end up in some non-standard format if documented with `//`s.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -47,7 +47,7 @@
  * Unit tests {@link org.apache.commons.lang3.ClassUtils}.
  */
 @SuppressWarnings("boxing") // JUnit4 does not support primitive equality testing apart from long
-public class ClassUtilsTest  {
+class ClassUtilsTest  {
```
garydgregory: Let's not change the visibility of test classes please.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -433,48 +440,83 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
-        } else {
-          if (availableSpace > 0) {
-            output[level] = part;
-          } else {
-            // if no space is left still the first char is used
-            output[level] = part.substring(0, 1);
-          }
+        if (className == null) {
+            return StringUtils.EMPTY;
         }
```
garydgregory: I was expecting a short-circuit here when the length hint is greater or equal than the input string, then return the input string to avoid creating a new String.

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -433,48 +440,83 @@ public static String getAbbreviatedName(final Class<?> cls, final int len) {
...
+     *
+     * @param className the className to get the abbreviated name for, may be {@code null}
+     * @param len       the desired length of the abbreviated name
+     * @return the abbreviated name or an empty string if the specified
+     * class name is {@code null} or empty string. The abbreviated name may be
+     * longer than the desired length if it cannot be abbreviated to the desired length.
+     * @throws IllegalArgumentException if {@code len <= 0}
      * @since 3.4
      */
     public static String getAbbreviatedName(final String className, final int len) {
```
garydgregory: I am starting to wonder if `len` would be better named `lengthHint` to make it clearer that this length is not a guarantee in any shape or form of the length of the returned string.

verhas: I can live with that. Makes sense.

src/main/java/org/apache/commons/lang3/ClassUtils.java
verhas: Makes sense.

src/main/java/org/apache/commons/lang3/ClassUtils.java
garydgregory: @verhas Then please go ahead and change it to `lengthHint`.

src/main/java/org/apache/commons/lang3/ClassUtils.java
garydgregory: @verhas Then please go ahead and update the code and add a test.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
garydgregory: @verhas ping.

src/main/java/org/apache/commons/lang3/ClassUtils.java
verhas: There is no point to add a test. This is not a feature change. This is a performance optimization. I do not think the method should guarantee as an API contract that it will return the same String object in this case.

In other cases I totally agree: if there is a change, which is functionality change then there should be a test also changed.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
garydgregory: @verhas ping.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -160,26 +162,59 @@ public void test_getAbbreviatedName_Class() {
...
         assertEquals("java.lang.String", ClassUtils.getAbbreviatedName(String.class, 20));
     }
 
+    /**
+     * Test that in case the required length is larger than the name and thus there is no need for any shortening
+     * then the returned string object is the same as the one passed as argument. Note, however, that this is
+     * tested as an internal implementation detail, but it is not a guaranteed feature of the implementation.
+     */
     @Test
-    public void test_getAbbreviatedName_Class_NegativeLen() {
```
garydgregory: Let's not change the visibility of test classes and methods please.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -160,26 +162,59 @@ public void test_getAbbreviatedName_Class() {
...
+    }
+
+    @Test
+    @DisplayName("When the desired length is negative then exception is thrown")
+    void test_getAbbreviatedName_Class_NegativeLen() {
         assertThrows(IllegalArgumentException.class, () -> ClassUtils.getAbbreviatedName(String.class, -10));
     }
 
     @Test
-    public void test_getAbbreviatedName_Class_ZeroLen() {
```
garydgregory: Let's not change the visibility of test classes and methods please.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -160,26 +162,59 @@ public void test_getAbbreviatedName_Class() {
...
+    @DisplayName("When the desired length is zero then exception is thrown")
+    void test_getAbbreviatedName_Class_ZeroLen() {
         assertThrows(IllegalArgumentException.class, () -> ClassUtils.getAbbreviatedName(String.class, 0));
     }
 
     @Test
-    public void test_getAbbreviatedName_String() {
```
garydgregory: Let's not change the visibility of test classes and methods please.

verhas: Why?

Since the project uses JUnit 5 the required visibility is package private. Any access modifier is noise.

garydgregory:
Because all other tests are public. IOW, consistency.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
verhas: Sorry, but wrong addressee. This is not me, who has changed the visibility, but JUnit 5.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
verhas: Sorry, but wrong addressee. This is not me, who has changed the visibility, but JUnit 5.

garydgregory:
> Please merge it instead of nitpicking.

That's not the best attitude IMO.

verhas:
>>Please merge it instead of nitpicking.
>That's not the best attitude IMO.

I said please. Btw: I opened a ticket https://issues.apache.org/jira/browse/LANG-1491 to fix this issue.

garydgregory:
> > > Please merge it instead of nitpicking.
> > > That's not the best attitude IMO.
> 
> I said please. Btw: I opened a ticket https://issues.apache.org/jira/browse/LANG-1491 to fix this issue.

It's rather simple IMO: this patch should be consistent with the current code base as it is now. IOW, tests are `public`.

If someone wants to do the busy work of changing all of Commons Lang, more power to them, then the community can decide what to do, but this PR is not the place to do it.

verhas:
I changed the test class and methods to be public.

garydgregory:
@verhas 
GitHub seems to think there conflicts here. May you rebase?

src/main/java/org/apache/commons/lang3/ClassUtils.java
verhas: Thanks

garydgregory:
@verhas 
In git master; please verify and close.

PR #448, opened 2019-08-26
Lang 1482

verhas:
mainly code clean up
extra tests that were missing
rethrow now corrects stack trace showing the location where the rethrow was called
(nobody cares that an exception was thrown from a utility function that does nothing else but throws the exception)
documentation was extended


src/test/java/org/apache/commons/lang3/FunctionsTest.java
```diff
@@ -434,4 +492,9 @@ public void testTryWithResources() {
         Functions.tryWithResources(() -> consumer.accept(null), co::close);
         assertTrue(co.isClosed());
     }
+
+    @Test
+    void testRethrowNull(){
```
verhas: this was missing and that one line in rethrow was not covered by test

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -549,10 +584,26 @@ public static RuntimeException rethrow(Throwable pThrowable) {
...
             } else if (pThrowable instanceof IOException) {
-                throw new UncheckedIOException((IOException) pThrowable);
+                throw popStackTrace(new UncheckedIOException((IOException) pThrowable));
             } else {
-                throw new UndeclaredThrowableException(pThrowable);
+                throw popStackTrace(new UndeclaredThrowableException(pThrowable));
             }
         }
     }
+
```
verhas: this may even could be a public utility method in some other class

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -143,164 +151,141 @@
...
      */
     public static <O> Callable<O> asCallable(FailableCallable<O, ?> pCallable) {
-        return () -> {
-            try {
-                return pCallable.call();
-            } catch (Throwable t) {
-                throw rethrow(t);
-            }
-        };
+        return () -> call(pCallable);
```
verhas: The functionality as it was implemented the first hand is already there.

src/test/java/org/apache/commons/lang3/FunctionsTest.java
```diff
@@ -414,8 +412,8 @@ public void testAsPredicate() {
         assertNotNull(cause);
         assertTrue(cause instanceof SomeException);
         assertEquals("Odd Invocation: 1", cause.getMessage());
+        // assert that it does not throw exception
         final boolean instance = predicate.test(null);
```
verhas: `assertNotNull()` makes no sense on primitive

verhas:
And now all the cleanup, extra JavaDoc is just gone...

PR #450, opened 2019-09-04
Add and use IllegalArgumentExceptions

garydgregory:
Add and use `IllegalArgumentExceptions` to combine creating `IllegalArgumentException` instances with formatted messages using `String#format()`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -28,6 +28,7 @@ import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
+import org.apache.commons.lang3.exception.IllegalArgumentExceptions;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.commons.lang3.mutable.MutableInt;
 
@@ -267,15 +268,13 @@ public class ArrayUtils {
             } else if (object instanceof Object[]) {
                 final Object[] entry = (Object[]) object;
                 if (entry.length < 2) {
-                    throw new IllegalArgumentException("Array element " + i + ", '"
-                        + object
-                        + "', has a length less than 2");
+                    throw IllegalArgumentExceptions.format("Array element %,d, '%s', has a length less than 2", i,
+                            object);
                 }
                 map.put(entry[0], entry[1]);
             } else {
-                throw new IllegalArgumentException("Array element " + i + ", '"
-                        + object
-                        + "', is neither of type Map.Entry nor an Array");
+                throw IllegalArgumentExceptions
+                        .format("Array element %,d, '%s', is neither of type Map.Entry nor an Array", i, object);
             }
         }
         return map;
@@ -5098,8 +5097,8 @@ public class ArrayUtils {
              */
             final Class<?> type2 = array2.getClass().getComponentType();
             if (!type1.isAssignableFrom(type2)) {
-                throw new IllegalArgumentException("Cannot store " + type2.getName() + " in an array of "
-                        + type1.getName(), ase);
+                throw IllegalArgumentExceptions.format(ase, "Cannot store %s in an array of %s", type2.getName(),
+                        type1.getName(), ase);
             }
             throw ase; // No, so rethrow original
         }
--- a/src/main/java/org/apache/commons/lang3/CharUtils.java
+++ b/src/main/java/org/apache/commons/lang3/CharUtils.java
@@ -16,6 +16,8 @@
  */
 package org.apache.commons.lang3;
 
+import org.apache.commons.lang3.exception.IllegalArgumentExceptions;
+
 /**
  * <p>Operations on char primitives and Character objects.</p>
  *
@@ -219,7 +221,7 @@ public class CharUtils {
      */
     public static int toIntValue(final char ch) {
         if (!isAsciiNumeric(ch)) {
-            throw new IllegalArgumentException("The character " + ch + " is not in the range '0' - '9'");
+            throw IllegalArgumentExceptions.format("The character %s is not in the range '0' - '9'", ch);
         }
         return ch - 48;
     }
--- a/src/main/java/org/apache/commons/lang3/Conversion.java
+++ b/src/main/java/org/apache/commons/lang3/Conversion.java
@@ -18,6 +18,8 @@ package org.apache.commons.lang3;
 
 import java.util.UUID;
 
+import org.apache.commons.lang3.exception.IllegalArgumentExceptions;
+
 
 /**
  * <p>
@@ -95,7 +97,7 @@ public class Conversion {
     public static int hexDigitToInt(final char hexDigit) {
         final int digit = Character.digit(hexDigit, 16);
         if (digit < 0) {
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
         }
         return digit;
     }
@@ -153,7 +155,7 @@ public class Conversion {
         case 'F':
             return 0xF;
         default:
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
         }
     }
 
@@ -211,7 +213,7 @@ public class Conversion {
         case 'F':
             return TTTT.clone();
         default:
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
         }
     }
 
@@ -269,7 +271,7 @@ public class Conversion {
         case 'F':
             return TTTT.clone();
         default:
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
         }
     }
 
@@ -372,10 +374,11 @@ public class Conversion {
      */
     public static char binaryToHexDigitMsb0_4bits(final boolean[] src, final int srcPos) {
         if (src.length > 8) {
-            throw new IllegalArgumentException("src.length>8: src.length=" + src.length);
+            throw IllegalArgumentExceptions.format("src.length>8: src.length=%,d");
         }
         if (src.length - srcPos < 4) {
-            throw new IllegalArgumentException("src.length-srcPos<4: src.length=" + src.length + ", srcPos=" + srcPos);
+            throw IllegalArgumentExceptions.format("src.length-srcPos<4: src.length=%,d, srcPos=%,d", src.length,
+                    srcPos);
         }
         if (src[srcPos + 3]) {
             if (src[srcPos + 2]) {
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -18,6 +18,8 @@ package org.apache.commons.lang3;
 
 import java.util.Random;
 
+import org.apache.commons.lang3.exception.IllegalArgumentExceptions;
+
 /**
  * <p>Operations for random {@code String}s.</p>
  * <p>Currently <em>private high surrogate</em> characters are ignored.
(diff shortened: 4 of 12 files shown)


garydgregory:
On Tue, Sep 3, 2019 at 8:59 PM Bruno P. Kinoshita <notifications@github.com>
wrote:

> *@kinow* approved this pull request.
>
> No objections from me, assuming Travis build passes and others are OK with
> this. When I read your proposal this morning via e-mail I didn't understand
> what it would look like, but after seeing this PR I guess it makes sense.
> Thanks!
>

@Bruno P. Kinoshita <kinow@apache.org>
TY & YW, I just fixed a Checkstyle issue, rebuilding...

Gary

michael-o:
Why not use our Validate class?

garydgregory:
These are factory methods, they don't validate anything.

XenoAmess:
If there should have an IllegalArgumentExceptions , then there should have a RuntimeExceptions.
then there should have ArithmeticExceptions.
and should have IndexOutOfBoundsExceptions, FileSystemNotFoundExceptions ......

I don't think IllegalArgumentException is special among them in this way (of having a String constructor).
Shouldn't we be fair to all of the throwable classes who have a String constructor?

garydgregory:
In the case of `IllegalArgumentException`, based on the usage within Commons Lang, and in the code bases I've worked on and currently see, `IllegalArgumentException` is used quite often. As you can see from this PR, there are quite a number of files that benefit from this short cut.

RuntimeException feels like a code smell to me, one should only rarely consider throwing such a generic exception, I'd even say it's an anti-pattern.

As a provider of a fairly general purpose library we have to find the balance between being general but also following the YAGNI principle. IMO, RuntimeException is not only YAGNI but also an anti-pattern. Extending outlook to other exceptions is fair and I would consider doing the same for other fairly general purpose exceptions like `NullPointerException` and `IllegalStateException`. This is just a small step.

michael-o:
> 
> 
> These are factory methods, they don't validate anything.

There are several spots you have changed which are subject to conditions and can use `Validate`.

src/main/java/org/apache/commons/lang3/CharUtils.java
```diff
@@ -219,7 +221,7 @@ public static char toChar(final String str, final char defaultValue) {
      */
     public static int toIntValue(final char ch) {
         if (!isAsciiNumeric(ch)) {
-            throw new IllegalArgumentException("The character " + ch + " is not in the range '0' - '9'");
+            throw IllegalArgumentExceptions.format("The character %s is not in the range '0' - '9'", ch);
```
robtimus: Shouldn't %s be %c instead? The same applies to some other exceptions using char.

garydgregory: I could be indeed but IIRC it does not matter since the result will be the same. Unless something interesting would happen related to Unicode.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -267,15 +268,13 @@ public static boolean isEquals(final Object array1, final Object array2) {
             } else if (object instanceof Object[]) {
                 final Object[] entry = (Object[]) object;
                 if (entry.length < 2) {
-                    throw new IllegalArgumentException("Array element " + i + ", '"
-                        + object
-                        + "', has a length less than 2");
+                    throw IllegalArgumentExceptions.format("Array element %,d, '%s', has a length less than 2", i,
+                            object);
```
sebbASF: %,d => %d

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -267,15 +268,13 @@ public static boolean isEquals(final Object array1, final Object array2) {
...
                 }
                 map.put(entry[0], entry[1]);
             } else {
-                throw new IllegalArgumentException("Array element " + i + ", '"
-                        + object
-                        + "', is neither of type Map.Entry nor an Array");
+                throw IllegalArgumentExceptions
+                        .format("Array element %,d, '%s', is neither of type Map.Entry nor an Array", i, object);
```
sebbASF: %,d => %d

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -95,7 +97,7 @@
     public static int hexDigitToInt(final char hexDigit) {
         final int digit = Character.digit(hexDigit, 16);
         if (digit < 0) {
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
```
sebbASF: That should be '%d'

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -153,7 +155,7 @@ public static int hexDigitMsb0ToInt(final char hexDigit) {
         case 'F':
             return 0xF;
         default:
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
```
sebbASF: %d

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -211,7 +213,7 @@ public static int hexDigitMsb0ToInt(final char hexDigit) {
         case 'F':
             return TTTT.clone();
         default:
-            throw new IllegalArgumentException("Cannot interpret '" + hexDigit + "' as a hexadecimal digit");
+            throw IllegalArgumentExceptions.format("Cannot interpret '%s' as a hexadecimal digit", hexDigit);
```
sebbASF: %d

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -269,7 +271,7 @@ public static int hexDigitMsb0ToInt(final char hexDigit) {
...
```
sebbASF: %d

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -372,10 +374,11 @@ public static char binaryToHexDigitMsb0_4bits(final boolean[] src) {
      */
     public static char binaryToHexDigitMsb0_4bits(final boolean[] src, final int srcPos) {
         if (src.length > 8) {
-            throw new IllegalArgumentException("src.length>8: src.length=" + src.length);
+            throw IllegalArgumentExceptions.format("src.length>8: src.length=%,d");
```
sebbASF: %,d => %d

sebbASF: The magic number 8 should be a constant

src/main/java/org/apache/commons/lang3/Conversion.java
```diff
@@ -372,10 +374,11 @@ public static char binaryToHexDigitMsb0_4bits(final boolean[] src) {
...
         }
         if (src.length - srcPos < 4) {
-            throw new IllegalArgumentException("src.length-srcPos<4: src.length=" + src.length + ", srcPos=" + srcPos);
+            throw IllegalArgumentExceptions.format("src.length-srcPos<4: src.length=%,d, srcPos=%,d", src.length,
+                    srcPos);
```
sebbASF: Magic number 4

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -359,7 +361,7 @@ public static String random(int count, int start, int end, final boolean letters
         if (count == 0) {
             return StringUtils.EMPTY;
         } else if (count < 0) {
-            throw new IllegalArgumentException("Requested random string length " + count + " is less than 0.");
+            throw IllegalArgumentExceptions.format("Requested random string length %,d is less than 0.", count);
```
sebbASF: %,d => %d

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -6654,10 +6656,8 @@ private static String replaceEach(
 
         // make sure lengths are ok, these need to be equal
         if (searchLength != replacementLength) {
-            throw new IllegalArgumentException("Search and Replace array lengths don't match: "
-                + searchLength
-                + " vs "
-                + replacementLength);
+            throw IllegalArgumentExceptions.format("Search and Replace array lengths don't match: %s vs %s",
+                    searchLength, replacementLength);
```
sebbASF: %s => %d ?

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -106,7 +108,7 @@ public Validate() {
      */
     public static void isTrue(final boolean expression, final String message, final long value) {
         if (!expression) {
-            throw new IllegalArgumentException(String.format(message, Long.valueOf(value)));
+            throw IllegalArgumentExceptions.format(message, Long.valueOf(value));
```
sebbASF: I think the above change shows why the class is unnecessary.
I don't think it makes sense to create a new class just to drop the text 'new ' and 'String.'

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -536,23 +545,22 @@ public static void tryWithResources(FailableRunnable<? extends Throwable> pActio
...
-     * Rethrow a {@link Throwable} as an unchecked exception.
+     * Rethrows a {@link Throwable} as an unchecked exception.
      * @param pThrowable The throwable to rethrow
      * @return Never returns anything, this method never terminates normally
      */
     public static RuntimeException rethrow(Throwable pThrowable) {
         if (pThrowable == null) {
             throw new NullPointerException("The Throwable must not be null.");
+        }
+        if (pThrowable instanceof RuntimeException) {
```
sebbASF: This is a completely different change; please don't mix changes in a pull request.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -5098,8 +5097,8 @@ public static boolean isNotEmpty(final boolean[] array) {
              */
             final Class<?> type2 = array2.getClass().getComponentType();
             if (!type1.isAssignableFrom(type2)) {
-                throw new IllegalArgumentException("Cannot store " + type2.getName() + " in an array of "
-                        + type1.getName(), ase);
+                throw IllegalArgumentExceptions.format(ase, "Cannot store %s in an array of %s", type2.getName(),
+                        type1.getName(), ase);
```
sebbASF: Why is the ase parameter listed twice? Looks wrong

src/main/java/org/apache/commons/lang3/ArrayUtils.java
garydgregory: Hi @sebbASF ,
One of the reasons for the new API is to use rich formatting, so why loose the comma-separated format since it is much easier for humans to read?

src/main/java/org/apache/commons/lang3/Conversion.java
garydgregory: OK

src/main/java/org/apache/commons/lang3/Conversion.java
garydgregory: Will do.

src/main/java/org/apache/commons/lang3/Conversion.java
garydgregory: Will do.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
garydgregory: See previous comment.

src/main/java/org/apache/commons/lang3/StringUtils.java
garydgregory: I will use `"Search and Replace array lengths don't match: %,d vs %,d"`

src/main/java/org/apache/commons/lang3/Validate.java
garydgregory: For this call site, I am 50/50, I implemented here for consistency. Picking this call site to question the existence of the whole PR does not take into account more complex use cases, IMO ;-)

src/main/java/org/apache/commons/lang3/Functions.java
garydgregory: Will commit to master instead. A PR for this is over the top IMO.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
garydgregory: I will remove the last `ase`. The rest is correct, it is not the same, one param is from `type2` and the next for `type1`.

src/main/java/org/apache/commons/lang3/Validate.java
sebbASF: The same reasoning applies to all the other places once you have changed the string concatenation to use String.format.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
sebbASF: OK, sorry, I did not realise that the comma was a valid numeric flag for the formatter.
It looked like a mistake.

PR #451, opened 2019-09-05
LANG-1485 : Add getters for lhs and rhs objects in DiffResult

nicolasbd:
As explained in the [JIRA ticket](https://issues.apache.org/jira/browse/LANG-1485) I add a getter for the `lhs` and `rhs` objects of the `DiffResult`
Tests included

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/DiffResult.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffResult.java
@@ -89,6 +89,22 @@ public class DiffResult implements Iterable<Diff<?>> {
     }
 
     /**
+     * <p>Return the object the right object has been compared to</p>
+     * @return the left object of the diff
+     */
+    public Object getLhs() {
+        return this.lhs;
+    }
+
+    /**
+     * <p>Return the object the left object has been compared to</p>
+     * @return the right object of the diff
+     */
+    public Object getRhs() {
+        return this.rhs;
+    }
+
+    /**
      * <p>
      * Returns an unmodifiable list of {@code Diff}s. The list may be empty if
      * there were no differences between the objects.
--- a/src/main/java/org/apache/commons/lang3/builder/DiffResult.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffResult.java
@@ -89,18 +89,20 @@ public class DiffResult implements Iterable<Diff<?>> {
     }
 
     /**
-     * <p>Return the object the right object has been compared to</p>
+     * <p>Gets the object the right object has been compared to.</p>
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/builder/DiffResult.java
```diff
@@ -88,6 +88,22 @@
         }
     }
 
+    /**
+     * <p>Return the object the right object has been compared to</p>
```
garydgregory: - You do not need to use p tags for the first sentence of a Javadoc comment IIRC. 
- A sentence should end in a period.
- Use "Gets ..." instead of "Return ..."

nicolasbd: thanks @garydgregory 
which version should I put in the `@since` tag ?

nicolasbd: Also maybe I could call the getters `getLeft` and `getRight` ... seems more readable, what do you think ?

garydgregory: @nicolasbd
Yes, please use `getLeft` and `getRight`, much better!

nicolasbd: @garydgregory ok cool, but what about the `@since` tag ? which number should I set it to ?

garydgregory: 3.10

nicolasbd: Ok @garydgregory , made the changes except for "Returns" and p tags for consistency with other javadocs comments

src/main/java/org/apache/commons/lang3/builder/DiffResult.java
```diff
@@ -88,6 +88,24 @@
...
+     * @since 3.10
+     * @return the left object of the diff
+     */
+    public Object getLeft() {
+        return this.lhs;
+    }
+
+    /**
+     * <p>Gets the object the left object has been compared to.</p>
+     * @ 3.10
```
garydgregory: Nope :-( Use `@since` not `@`. Also, `@since` should be the _last_ tag.

nicolasbd: damn' sorry for the typo, ok for putting it in last
also got back to "Returns" for consistency with other Javadocs comments

XenoAmess:
Just wondered...
Should we make DiffResult extends Pair as well?

PR #452, opened 2019-09-05
[LANG-1486] Generify builder classes Diffable, DiffBuilder, and

garydgregory:
[LANG-1486] Generify builder classes Diffable, DiffBuilder, and DiffResult.


garydgregory:
> 
> 
> Well, after thinking about it, I'm not sure how we should deal with custom comparisons between two objects of different Type, or even add arbitrary differences to a diff, even if it's not related to one of the fields of the base object

Yes, that's a concern but what we started with assumes that the type is the same. Feel free to continue the discussion on the dev mailing list and propose a PR :-)

PR #454, opened 2019-09-07
Make diff result extends pair

XenoAmess:
for compatibility to Diff who extends Pair as well.


garydgregory:
My only concern here is what JApiCmp reports:
```
mvn install -P japicmp japicmp:cmp
```
Yields:
```
MODIFIED (Serializable incompatible(!): superclass modified) public class org.apache.commons.lang3.builder.DiffResult
```
Which I think means it breaks binary compatibility. Can someone confirm either way?

Stzx:
@garydgregory 

+1

`DiffResult` extend `Pair` and does break the serialization function of `Pair`.

XenoAmess:
> @garydgregory
> 
> +1
> 
> `DiffResult` extend `Pair` and does break the serialization function of `Pair`.

My bad. I forgot to add serialVersionUID for it.

garydgregory:
I created https://issues.apache.org/jira/browse/LANG-1503 to bring in the changes to the three Pair classes. Note that there are only new factory methods. The additional constructors are not needed. I think it is also better design to have the conversion work done in factory methods and the results funneled to the constructors.

The DiffResult change of superclass is not brought in as it would break binary compatibility.
You can check this by running:
```
mvn clean package site -V -P jacoco -P japicmp
```
and checking the generated report in the site at `target\site\index.html` where you will see:
```
MODIFIED (Serializable incompatible(!): superclass modified) public class org.apache.commons.lang3.builder.DiffResult
```

XenoAmess:
close as https://github.com/scope-demo/commons-lang/commit/aa3d92498e223ea051f3965d80b9428b4f5d9e98

PR #457, opened 2019-09-13
[LANG-1490] - Creates DiffResultView and DiffView classes

nicolasbd:
Currently, a DiffResult and a Diff can only handle the differences of two objects of the same type A, and we cannot set custom differences for a DiffResult<A> if encapsulated objects of type B are different.

I suggest to implement a DiffView class as a simple `String field`, `String leftValue` and `String rightValue` triplet. 

I also build a DiffResultView holding two objects of type T and a list of DiffView objects.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/DiffResultView.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffResultView.java
@@ -1,0 +1,157 @@
+package org.apache.commons.lang3.builder;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * <p>
+ * A {@code DiffResultView} encapsulates two objects of the same type and list their differences
+ * as a list of {@link DiffView} objects, each holding a field name and the two different values found for this field
+ * in the Left and Right object.
+ * </p>
+ *
+ * <p>
+ * It can be built with or without a pre-defined list of {@link DiffView} objects, or from an existing
+ * {@link DiffResult}.
+ * </p>
+ *
+ * @param <T> type of the left and right object.
+ *
+ * @since 3.10
+ */
+public class DiffResultView<T> {
+
+	/**
+	 * The list of {@link DiffView} objects for the two compared objects.
+	 */
+	private List<DiffView> diffs;
+
+	/**
+	 * Holds the left object of the comparison.
+	 */
+	private T left;
+	/**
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @nicolasbd ,

Both files are missing the Apache license header.

Gary

nicolasbd:
@garydgregory ok, but what do you think about the idea itself ? 

I'll add proper headers and tests once we decide if this should be added to the lib

garydgregory:
Let me take a closer look this weekend. In the meantime, I would also like to work on a patch that defines as `Diffable`:

```
public interface Diffable<T extends Diffable<T>>
```

This will allow diffs on a class and one of its subclasses for example.

src/main/java/org/apache/commons/lang3/builder/DiffView.java
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * <p>Constructs a {@code DiffView} from 3 string params: field, leftValue and rightValue</p>
+	 *
+	 * @param field designates the field identified as different.
+	 * @param leftValue holds the string representation of the value in the left side of the comparison.
+	 * @param rightValue holds the string representation of the value in the right side of the comparison.
+	 */
+	public DiffView(String field, String leftValue, String rightValue) {
+		this.field = field;
```
kinow, line 22: I think the files were formatted with tabs, and this may cause issues in Travis when it builds using Checkstyle.

src/main/java/org/apache/commons/lang3/builder/DiffView.java
```diff
@@ -0,0 +1,64 @@
+package org.apache.commons.lang3.builder;
```
kinow, line 1: I think this file is missing the copyright license header.

src/main/java/org/apache/commons/lang3/builder/DiffResultView.java
```diff
@@ -0,0 +1,157 @@
...
+	 *                      will be built and added to the list.
+	 */
+	public void addDiffs(DiffResult<?> diffResult) {
+		diffs.addAll(
+				diffResult
+						.getDiffs()
+						.stream()
+						.map(DiffView::new)
+						.collect(Collectors.toList())
+		);
```
kinow, line 126: Probably will fail checkstyle due to tabs here, instead of spaces.

PR #459, opened 2019-09-16
(doc): Document public RandomStringUtils exploit

JLLeitschuh:
Document the public exploit that exists for `RandomStringUtils`.

This documentation update stems from my finding this vulnerability in the very popular code generator JHipster.

https://nvd.nist.gov/vuln/detail/CVE-2019-16303

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -37,7 +37,7 @@ import java.util.Random;
  * class relies, are <b>not cryptographically secure</b>.
  * Do not use this classes' default implementation of {@link Random} in security sensitive locations,
  * for example password reset key generation, as all future values can be computed as proven by
- * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd">
+ * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd?source=friends_link&sk=3db1c41cc81a58f70ed05a7315191385">
  * this proof of concept.</a></p>
  *
  * <p>Please note that the Apache Commons project provides a component
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -34,7 +34,11 @@ import java.util.Random;
  * RandomStringGenerator</a> instead.</p>
  *
  * <p>Caveat: Instances of {@link Random}, upon which the implementation of this
- * class relies, are not cryptographically secure.</p>
+ * class relies, are <b>not cryptographically secure</b>.
+ * Do not use this classes' default implementation of {@link Random} in security sensitive locations,
+ * for example password reset key generation, as all future values can be computed as proven by
+ * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd">
+ * this proof of concept.</a></p>
  *
  * <p>Please note that the Apache Commons project provides a component
  * dedicated to pseudo-random number generation, namely


src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -34,7 +34,11 @@
  * RandomStringGenerator</a> instead.</p>
  *
  * <p>Caveat: Instances of {@link Random}, upon which the implementation of this
- * class relies, are not cryptographically secure.</p>
+ * class relies, are <b>not cryptographically secure</b>.
+ * Do not use this classes' default implementation of {@link Random} in security sensitive locations,
+ * for example password reset key generation, as all future values can be computed as proven by
+ * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd">
```
garydgregory: Do we really want to link to a site that only let's you read a limited number of articles per month? Isn't there a freer source for this information?

JLLeitschuh: As an author that regularly publishes on medium, you can choose to monizie or not monizie. It doesn't look like this one is monizied. I can ask the author for a "friend" link that doesn't have that pop-up when you load it though.

JLLeitschuh:
```suggestion
 * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd?source=friends_link&sk=3db1c41cc81a58f70ed05a7315191385">
```

garydgregory: I say we KISS it and put in the comment without a link.

JLLeitschuh: I mean, it's not that hard to find if you know what to Google. But this makes a software engineer/PEN Tester/security researchers life easier by fully explaining the potential impacts of their choices here. I figure the more documentation, the better.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -34,7 +34,11 @@
  * RandomStringGenerator</a> instead.</p>
  *
  * <p>Caveat: Instances of {@link Random}, upon which the implementation of this
- * class relies, are not cryptographically secure.</p>
+ * class relies, are <b>not cryptographically secure</b>.
+ * Do not use this classes' default implementation of {@link Random} in security sensitive locations,
```
kinow: I think it should be 

this class' default..

Or

These classes'

Or simply

Avoid using the default implementation in this classs if you need a cryptographically secure...

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -34,7 +34,11 @@
...
+ * for example password reset key generation, as all future values can be computed as proven by
+ * <a href="https://medium.com/@alex91ar/the-java-soothsayer-a-practical-application-for-insecure-randomness-c67b0cd148cd?source=friends_link&sk=3db1c41cc81a58f70ed05a7315191385">
```
kinow: I agree with others on the issue with a link to medium, but I don't care much about monetization.

If we have a link to CVE, a paper published in some journal, a standard documentation from some site like ietf, or a wikipedia page, it would be preferrable IMHO.

If this is the only place with an explanation, or the best link to understand the issue, then we need to use a web.archive.org link to prevent it from disappearing after some years.

chtompki:
Did anyone note that in 3.9 we have the following at the top of the JavaDoc for `RandomStringUtils`

> Caveat: Instances of Random, upon which the implementation of this class relies, are not cryptographically secure.

> Please note that the Apache Commons project provides a component dedicated to pseudo-random number generation, namely Commons RNG, that may be a better choice for applications with more stringent requirements (performance and/or correctness).

which, both provides the point we're trying to make as well as a place to look for a solution to the insecurity of the randomness.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
JLLeitschuh: @alex91ar Thoughts?

JLLeitschuh:
> Please note that the Apache Commons project provides a component dedicated to pseudo-random number generation, namely Commons RNG, that may be a better choice for applications with more stringent requirements (performance and/or correctness).

When I read this, I don't think this refers to 'security' at all. 'Correctness' & 'Performance' do not equate to providing security benefit when I read it.

If I had more time, I'd actually propose formally deprecating every single method on this class in favor of an API like this:

```java
// Uses new Random()
RandomStringUtils.insecureAndPredictable().random(32);

// Uses new SecureRandom()
RandomStringUtils.secure().random(32);

RandomStringUtils.random(SecureRandom.getInstanceStrong()).random(32);
```

That way it's made very clear to the developer that they are making an explicit choice for security here instead of defaulting to insecure randomness.

I'm arguing this same exact point to the Kotlin team here:
https://github.com/Kotlin/KEEP/issues/184

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
alex91ar: If you'd like I can port the article to somewhere else, just say where and I will!

chtompki:
I'm still mildly confused how:

https://github.com/apache/commons-lang/blob/commons-lang-3.9/src/main/java/org/apache/commons/lang3/RandomStringUtils.java#L36-L37

> Caveat: Instances of [Random](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html), upon which the implementation of this class relies, are not cryptographically secure.

doesn't lead the reader to conclude the class is not to be used for the purpose of security and in turn minimally use `SecureRandom` (as stipulated in the `java.util.Random` javadoc, to which we've linked).

JLLeitschuh:
@chtompki Because many people don't read the documentation. Especially on the top of classes.

I've found this class of vulnerability in other places because of similar issues around not reading the documentation:

- https://nvd.nist.gov/vuln/detail/CVE-2019-11808?cpeVersion=2.2

I've got 3 outstanding undisclosed vulnerabilities I've reported due to insecure RNG caused by this class.

The problem is, in security, defaults really matter. Unfortunately, by defaulting to insecure RNG, this class is exposing a lot of projects to this vulnerability.

Want some examples? Just GitHub search for "RandomStringUtils token" or "RandomStringUtils key" on github. You'll find tens of thousands of examples.

https://github.com/search?q=RandomStringUtils+token&type=Code

garydgregory:
> @chtompki Because many people don't read the documentation. Especially on the top of classes.
uh? That's where this kind of information belongs IMO. "Because many people don't" also implies that many people do. So it's not saying much IMO ;-) Don't assume other folks' brain work like yours or or colleagues'.

My POV here is that this is Javadoc for a util class, we don't need to link to articles on a "proof" on reasons to not use it; if we want to discourage use cases in certain scenarios, we just say so and we're done. If there is a CVE to deal with, let's link to that CVE.

JLLeitschuh:
Any thoughts on my proposal here around moving the use of insecure randomness to be a more explicit decision as I detailed above:

https://github.com/apache/commons-lang/pull/459#issuecomment-532330032

I've got two more public more CVE's I've had issued as a result of `RandomStringUtils`:

 - [CVE-2019-10754](https://nvd.nist.gov/vuln/detail/CVE-2019-10754)
 - [CVE-2019-10755](https://nvd.nist.gov/vuln/detail/CVE-2019-10755)

There are a couple more that aren't public yet.

garydgregory:
Please see git master for a Javadoc update to this class which I believe emphasizes the main points brought up by this discussion. If someone would like to improve further upon that, then feel free to rebase this PR on master.

fche:
> > Caveat: Instances of [Random](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html), upon which the implementation of this class relies, are not cryptographically secure.
> 
> doesn't lead the reader to conclude the class is not to be used for the purpose of security and in turn minimally use `SecureRandom` (as stipulated in the `java.util.Random` javadoc, to which we've linked).

Any chance of coming around to "let's be secure" as a default position?

alex91ar:
So, would it be possible to create a Java CVE over this vulnerability? Thanks.

XenoAmess:
Why not change to use java.security.SecureRandom here?

Is there be any reason for not using it instead?

iProdigy:
No longer a concern given #1235

garydgregory:
We can close this I think unless someone thinks the git master docs can be further improved.

garydgregory:
Closing, no further comments, see current code in git master.

alex91ar:
Shouldn't I get a bounty for this? I mean it's a big vulnerability.

raboof:
As a volunteer-based open source organization, the ASF doesn't have a bug bounty program at this time. There are some 3rd-party bounty programs (https://hackerone.com/ibb, https://yeswehack.com/, https://huntr.com/) that have some ASF projects in scope, but I'm not aware of any of those covering commons-lang.

AFAIK all bug bounty programs have the prerequisite that the issue is responsibly reported to the project in private - AFAICT that didn't happen here, so it wouldn't qualify for that reason either.

alex91ar:
Well if anyone is feeling generous: bc1qrjv23a92uyqrptlkc9mc77w83344k4v37pem7f

garydgregory:
I agree with @raboof

PR #461, opened 2019-09-27
Lang 1491

verhas:
Removing all public modifiers that are not needed anymore when we are using JUnit 5 and not JUnit 4.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
@@ -45,7 +45,7 @@ import org.junit.jupiter.api.BeforeEach;
 
 /**
  */
-public class AnnotationUtilsTest {
+class AnnotationUtilsTest {
     @TestAnnotation(
             booleanValue = false,
             booleanValues = { false },
@@ -401,7 +401,7 @@ public class AnnotationUtilsTest {
     private Field field4;
 
     @BeforeEach
-    public void setup() throws Exception {
+    void setup() throws Exception {
         field1 = getClass().getDeclaredField("dummy1");
         field2 = getClass().getDeclaredField("dummy2");
         field3 = getClass().getDeclaredField("dummy3");
@@ -509,7 +509,7 @@ public class AnnotationUtilsTest {
 
     @Test
     @TestMethodAnnotation(timeout = 666000)
-    public void testToString() {
+    void testToString() {
         assertTimeoutPreemptively(Duration.ofSeconds(666L), () -> {
             final TestMethodAnnotation testAnnotation =
                     getClass().getDeclaredMethod("testToString").getAnnotation(TestMethodAnnotation.class);
--- a/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/AnnotationUtilsTest.java
@@ -409,41 +409,41 @@ public class AnnotationUtilsTest {
     }
 
     @Test
-    public void testEquivalence() {
+    void testEquivalence() {
         assertTrue(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), field2.getAnnotation(TestAnnotation.class)));
         assertTrue(AnnotationUtils.equals(field2.getAnnotation(TestAnnotation.class), field1.getAnnotation(TestAnnotation.class)));
     }
 
     @Test
-    public void testSameInstance() {
+    void testSameInstance() {
         assertTrue(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), field1.getAnnotation(TestAnnotation.class)));
     }
 
     @Test
-    public void testNonEquivalentAnnotationsOfSameType() {
+    void testNonEquivalentAnnotationsOfSameType() {
         assertFalse(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), field3.getAnnotation(TestAnnotation.class)));
         assertFalse(AnnotationUtils.equals(field3.getAnnotation(TestAnnotation.class), field1.getAnnotation(TestAnnotation.class)));
     }
 
     @Test
-    public void testAnnotationsOfDifferingTypes() {
+    void testAnnotationsOfDifferingTypes() {
         assertFalse(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), field4.getAnnotation(NestAnnotation.class)));
         assertFalse(AnnotationUtils.equals(field4.getAnnotation(NestAnnotation.class), field1.getAnnotation(TestAnnotation.class)));
     }
 
     @Test
-    public void testOneArgNull() {
+    void testOneArgNull() {
         assertFalse(AnnotationUtils.equals(field1.getAnnotation(TestAnnotation.class), null));
         assertFalse(AnnotationUtils.equals(null, field1.getAnnotation(TestAnnotation.class)));
     }
 
     @Test
-    public void testBothArgsNull() {
+    void testBothArgsNull() {
         assertTrue(AnnotationUtils.equals(null, null));
     }
 
     @Test
-    public void testIsValidAnnotationMemberType() {
+    void testIsValidAnnotationMemberType() {
         for (final Class<?> type : new Class[] { byte.class, short.class, int.class, char.class,
                 long.class, float.class, double.class, boolean.class, String.class, Class.class,
                 NestAnnotation.class, TestAnnotation.class, Stooge.class, ElementType.class }) {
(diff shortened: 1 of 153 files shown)


kinow:
Hi @verhas 

Thanks for the pull request. It changes a lot of files, it was probably hard to refactor everything.

I'm still more comfortable with JUnit 4. But from what I understood, JUnit 5 now supports multiple modifiers instead of requiring public. Do you have a link to the docs about why it needs to be default now, instead of public?

Cheers
Bruno

verhas:
Why doesn't the project has a unique and generally followed style-guide that governs questions like this.

You say that "if future PR's have tests that omit or use `private` modifier, I'm all good for that. Just don't think we need to change every test we have had."

My other pull request was refused because I omitted the `public` keyword from the `@Test` method. I was told that this was not consistent.

So you (plural), core team members are not on the same page. Whenever I create a pull request should I bend the style to fit the needs of the specific reviewer?

I made this pull request to make it consistent and also to eliminate the `public` keywords, which in my view are essentially noise in this case.

verhas:
"I'm still more comfortable with JUnit 4." 

If I were you I would not disclose that information in public. JUnit 5 was released more than two years ago.

"But from what I understood, JUnit 5 now supports multiple modifiers instead of requiring public."

No, it does not. It requires that the tests are at least package-private. There is no library, except some security enforcing or style check tools that warn or fail you if an protection of a member is looser than it is needed. After all, you may want to call your test methods from some code that needs it to be public. JUnit 5 does not require it to be public, so the test methods should be declared only public if there is some extra use of them other than the JUnit 5 framework that needs them to be public.

"Do you have a link to the docs about why it needs to be default now, instead of public?"

Common sense when programming in Java. Every class and member should be declared with the least permissive modifier. Why don't we just declare everything public? In the case of JUnit 5 tests that is package-private.

I also feel some inconsistency not liking this pull request because some future pull-requests may still come up using the public modifier in tests and the same time saying that if some other pull requests (without this one merged) come up without the public modifier they are ok.

kinow:
> Why doesn't the project has a unique and generally followed style-guide that governs questions like this.
> 
> You say that "if future PR's have tests that omit or use `private` modifier, I'm all good for that. Just don't think we need to change every test we have had."
> 
> My other pull request was refused because I omitted the `public` keyword from the `@Test` method. I was told that this was not consistent.
> 
> So you (plural), core team members are not on the same page. Whenever I create a pull request should I bend the style to fit the needs of the specific reviewer?
> 
> I made this pull request to make it consistent and also to eliminate the `public` keywords, which in my view are essentially noise in this case.

Sorry about your other pull request. It's strange it was closed only for that, unless the whole PR was just to change a method modifier.

kinow:
> "I'm still more comfortable with JUnit 4."
> 
> If I were you I would not disclose that information in public. JUnit 5 was released more than two years ago.
> 
> "But from what I understood, JUnit 5 now supports multiple modifiers instead of requiring public."
> 
> No, it does not. It requires that the tests are at least package-private. There is no library, except some security enforcing or style check tools that warn or fail you if an protection of a member is looser than it is needed. After all, you may want to call your test methods from some code that needs it to be public. JUnit 5 does not require it to be public, so the test methods should be declared only public if there is some extra use of them other than the JUnit 5 framework that needs them to be public.
> 
> "Do you have a link to the docs about why it needs to be default now, instead of public?"
> 
> Common sense when programming in Java. Every class and member should be declared with the least permissive modifier. Why don't we just declare everything public? In the case of JUnit 5 tests that is package-private.
> 
> I also feel some inconsistency not liking this pull request because some future pull-requests may still come up using the public modifier in tests and the same time saying that if some other pull requests (without this one merged) come up without the public modifier they are ok.

Thanks for your reply @verhas. Do you have a link to any JUnit 5 documentation supporting what you just said, please?

I just had a look at one example of their docs (https://junit.org/junit5/docs/current/user-guide/), and simplified it removing the `Calculator`.

```java
import static org.junit.jupiter.api.Assertions.assertEquals;

public class TestTests {

    @Test
    void addition() {
        assertEquals(2, 2);
    }
}
```

Executed with Maven and Eclipse, both worked fine.

```
[INFO] Running br.eti.kinoshita.TestTests
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.06 s - in br.eti.kinoshita.TestTests
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time: 01:10 min
[INFO] Finished at: 2019-10-06T09:15:18+13:00
```

Changing the modifier to `public`, I got the exact same output on both, with no issues/warnings, etc.

verhas:
>Do you have a link to any JUnit 5 documentation supporting what you just said, please?

https://junit.org/junit5/docs/current/user-guide/ says: "Test classes, test methods, and lifecycle methods are not required to be public, but they must not be private."

Also, note that not a single `@Test` sample method in the documentation is `public`. All are package private.

>Changing the modifier to `public`, I got the exact same output on both, with no issues/warnings, etc.

What a surprise. It works as documented.

The documentation does not say they have to be public. It only says, as the documentation also says: they can be public. It is not the library or the language that enforces them not to be public. It is just the programming style. Leaving the test method to be `public` after migrating from JUnit 4 to JUnit 5 is laziness and technical debt.

We could do many other things as well. For example, we could declare them all as `final`. It would be just as senseless, superfluous, useless and misleading as declaring them to be `public`. Maybe, the `final` declaration is even less misleading as it is really a fact that nobody is overriding these methods while declaring them `public` suggests that somebody is calling them from outside the package, which nobody does. So my suggestion is that if we want to declare all these methods to be `public` we also have to declare them to be `final`, which will still be mad, but even though it makes more sense.

In my reading JUnit 5 does not support public tests. It just accepts them and works with them. It lets the programmer declare them to be public, because from the framework point of view why not. From the programming style point of view, however ...

kinow:
So I think it doesn't require us to remove them (i.e. everything will keep working), so I keep that IMO we don't need to change. But again, if other committers prefer to change, all good for me :+1:

verhas:
>So I think it doesn't require us to remove them (i.e. everything will keep working),

Somehow I have the strange feeling that you did not get the notion of the whole pull request. 

It is not a new feature. This is not a bug fix.

* This is code cleanup.

Code cleanup means to delete/modify outdated or originally wrong structures in the code so that the code becomes more readable and thus more maintainable. By the nature of code cleanup (also called refactoring), the functionality does not change, without it, everything will keep working.

Do you really suggest that the project does not ever need any code cleanup? All your reasonings can be used against any code cleanup. Or, perhaps, you have some specific reason against this actual clean-up that you have not put forth so far?

garydgregory:
> It would be just as senseless, superfluous, useless and misleading as declaring them to be public.

All of that really? It's time to tone down the rhetoric IMO,

garydgregory:
> > So I think it doesn't require us to remove them (i.e. everything will keep working),
> 
> Somehow I have the strange feeling that you did not get the notion of the whole pull request.
> 
> It is not a new feature. This is not a bug fix.
> 
> ## This is code cleanup.
> Code cleanup means to delete/modify outdated or originally wrong structures in the code so that the code becomes more readable and thus more maintainable. By the nature of code cleanup (also called refactoring), the functionality does not change, without it, everything will keep working.
> 
> Do you really suggest that the project does not ever need any code cleanup? All your reasonings can be used against any code cleanup. Or, perhaps, you have some specific reason against this actual clean-up that you have not put forth so far?

Sigh, do you really intend to give a lecture in all of your PRs?

The best way to cooperate in this community IMO is respect and cooperation. You submit something and people say "fine with A but let's not change B", my advice is to go along with it. Then as you participate on the mailing lists and in other PRs, you'll get more of a feel for the community and a chance to get your points across. THIS is not the way to do it (for my taste.)

verhas:
I do not see where you see in my words the lack of respect.

My change makes the code more compliant with Item 15 of Joshua Bloch Effective Java, Third Edition (Minimize the accessibility of classes and members).

Please focus on the code and do not speak about rhetoric and who is giving lectures to whom.

N.B.: You are absolutely right when you feel some emotion from my words. I really feel pissed off when I invested a lot of effort to improve the code and what I get back is refusal with reasons that are against documented industry best practice.

verhas:
I am sorry.

I had a discussion with my son, and he read this thread and gave me feedback, that I accepted. I also edited my past comments to remove the parts that I see now, were offending. I did that as a gesture and not to hide the traces of my bad communication. The original versions are there, visible and also included as a quote in replies word by word, which is OK.

My son told me that the original misunderstanding was that I communicated the aim of the change ambiguous. Because of that ambiguity, Mr. Kinoshita was asking me about the reference where JUnit requires that the tests are not public. There is no such thing. JUnit allows the tests to be public. They just do not demand the tests to be public anymore as we change from JUnit 4 to JUnit 5. This misunderstanding is my fault. I can also see that my response to this was not-professional in the details I deleted.

Having all said I still believe that this pull request and also the next one following this is valuable for the project helping to keep the code clean and up to date and the PR should not be neglected.

garydgregory:
Hi Peter,

Thank you sharing your insights with us.

The best path for the PR IMO is to minimize the change set by keeping
elements that are public unchanged.

We have had other PRs in other Commons components recently with updates
from JUnit 4 to 5, all of these were merged and did not change visibility
down from public.

Gary

PR #462, opened 2019-09-27
General Code Cleanup

verhas:


src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -833,30 +833,28 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
...
-                    || Double.TYPE.equals(toClass);
+                    || isSomeNumberType(toClass);
             }
             // should never get here
             return false;
         }
         return toClass.isAssignableFrom(cls);
     }
 
+    private static boolean isSomeNumberType(Class<?> toClass) {
```
verhas: This is not an automatic fix. This code was copy-paste.

src/test/java/org/apache/commons/lang3/RandomUtilsTest.java
```diff
@@ -93,15 +92,6 @@ public void testNextFloatLowerGreaterUpper() {
         assertThrows(IllegalArgumentException.class, () -> RandomUtils.nextFloat(2, 1));
     }
 
-    /**
-     * Tests next boolean
-     */
-    @Test
-    public void testBoolean() {
```
verhas: This test does nothing. `RandomUtils.nextBoolean()` returns a `boolean` that can either be `true` or `false`. This is guaranteed by the language, there is no need to test it. It is kind of evident.

src/test/java/org/apache/commons/lang3/RandomUtilsTest.java
```diff
@@ -40,7 +39,7 @@
 
     @Test
     public void testConstructor() {
-        assertNotNull(new RandomUtils());
```
verhas: The `new` operator never returns `null`. `new RandomUtils()` cannot be `null` in Java.

src/test/java/org/apache/commons/lang3/builder/CompareToBuilderTest.java
```diff
@@ -884,9 +884,9 @@ public void testMultiBooleanArray() {
 
     @Test
     public void testRaggedArray() {
-        final long array1[][] = new long[2][];
```
verhas: This array declaration is called "C-type array declaration" and it is not recommended.

src/test/java/org/apache/commons/lang3/builder/DiffResultTest.java
```diff
@@ -78,11 +78,10 @@ public void testIterator() {
         final Iterator<Diff<?>> expectedIterator = diffs.iterator();
 
         final DiffResult list = new DiffResult(lhs, rhs, diffs, SHORT_STYLE);
-        final Iterator<Diff<?>> iterator = list.iterator();
 
-        while (iterator.hasNext()) {
+        for (Diff<?> diff : (Iterable<Diff<?>>) list) {
```
verhas: This transformation was automatically done by IntelliJ.

src/test/java/org/apache/commons/lang3/RandomUtilsTest.java
Stzx: If you clear the assertion, it seems useless to keep an expression for the new object.

If not used, it should be removed.

verhas: Done. Also in the other cases and all the default constructors were removed in these cases that do not need to be there explicit. I moved the relevant part of the constructor javadoc to the class documentation.

Stzx: Oh, sorry.

I don't think I'm clear, I mean delete is to delete the useless expression like `new RandomUtils()`.

verhas: That was absolutely clear, but I was picking up the thread when I checked why the test instantiates the class at all and then I could see that

1.) there was no reason
2.) the constructors were calling super() which is done by default
3.) the constructors were the default constructors that do not need to be written
4.) some tests just tested that the class with only the default constructor exists in a different way

so I deleted the constructors, where they are created anyway and I changed the tests to follow the same pattern.

verhas:
"Coverage decreased (-0.2%) to 94.965% when pulling 5e197f3 on verhas:CLEANUP into 4f3d3b4 on apache:master."

This simply shows that the tested code became shorter. Yes, I also deleted a test that actually did nothing, it was asserting that a returned boolean value is either true or false (what else?).

So if anyone takes the effort to review this pull request then do not blindly toss it aside because of the coverage in terms of percentage (which is not a good measure really for the very reason demonstrated here) decreased.

src/main/java/org/apache/commons/lang3/CharSet.java
```diff
@@ -225,7 +225,7 @@ protected void add(final String str) {
 // NOTE: This is no longer public as CharRange is no longer a public class.
 //       It may be replaced when CharSet moves to Range.
     /*public*/ CharRange[] getCharRanges() {
-        return set.toArray(new CharRange[set.size()]);
+        return set.toArray(new CharRange[0]);
```
Michael1993: Why not just CharRange[]::new?

Michael1993: The same goes for every other toArray changed this way in the pull request.

Michael1993: My bad, that came with Java 11, I was thinking of `stream().toArray()`.

PR #466, opened 2019-10-14
Functions Javadoc

verhas:
JavaDoc explains why the method that never returns declares a return type and what pattern to use it.


src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -491,9 +491,27 @@ public static void tryWithResources(FailableRunnable<? extends Throwable> pActio
...
 
     /**
-     * Rethrows a {@link Throwable} as an unchecked exception.
-     * @param pThrowable The throwable to rethrow
-     * @return Never returns anything, this method never terminates normally
+     * <p>Rethrow a {@link Throwable} as an unchecked exception. If the argument is
+     * already unchecked, namely a {@code RuntimeException} or {@code Error} then
+     * the argument will be rethrown without modification. If the exception is
+     * {@code IOException} then it will be enveloped into a {@code UncheckedIOException}.
+     * In every other cases the exception will be enveloped into a {@code
```
garydgregory: "enveloped" -> The term "wrapped" is a more standard term IMO.

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -491,9 +491,27 @@ public static void tryWithResources(FailableRunnable<? extends Throwable> pActio
...
+     * UndeclaredThrowableException}</p>
+     *
+     * <p>Note that there is a declared return type for this method, even though it
+     * never returns. The reason for that is to support the usual pattern:</p>
+     *
+     * <pre>{@code
+     *      throw rethrow(myUncheckedException);
+     * }</pre>
+     *
+     * <p>instead of just calling the method. This pattern helps the Java compiler to
```
garydgregory: "help" is not unlikely to be guaranteed, it might be safer to say "may help".

verhas: I do not mind changing the wording and I have done.

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -491,9 +491,27 @@ public static void tryWithResources(FailableRunnable<? extends Throwable> pActio
     }
 
     /**
-     * Rethrows a {@link Throwable} as an unchecked exception.
-     * @param pThrowable The throwable to rethrow
-     * @return Never returns anything, this method never terminates normally
+     * <p>Rethrow a {@link Throwable} as an unchecked exception. If the argument is
```
garydgregory: In general, Javadocs always start with "Gets", not "Get", "Returns", not "Return", and so on; so here let's keep "Rethrows", and not change to "Rethrow" SVP.

src/main/java/org/apache/commons/lang3/Functions.java
```diff
@@ -491,9 +491,27 @@ public static void tryWithResources(FailableRunnable<? extends Throwable> pActio
...
+     * already unchecked, namely a {@code RuntimeException} or {@code Error} then
+     * the argument will be rethrown without modification. If the exception is
+     * {@code IOException} then it will be wrapped into a {@code UncheckedIOException}.
+     * In every other cases the exception will be wrapped into a {@code
+     * UndeclaredThrowableException}</p>
+     *
+     * <p>Note that there is a declared return type for this method, even though it
+     * never returns. The reason for that is to support the usual pattern:</p>
+     *
+     * <pre>{@code
```
garydgregory: The pre tag is enough, there is no need to use the {code ...} within it.

verhas: It is just a habit and a bit safer for later modification. But I am not insisting it to keep.

The difference is that you can use < and > characters instead of their escape sequence if there is the `{@code ...}` also. I have learned this from the JDK source where it is used many times. In this case there is no need. I will remove it.

garydgregory:
Please do not change "Rethrows" to "Rethrow". Javadoc comments should starts with "Gets", "Returns", "Sets" and so on.

verhas:
Is there anything else I can do to get this simple change merged?

PR #468, opened 2019-10-15
Boolean utils little cleanup

verhas:
Some of the boxing and unboxing operations remained in the code since the pre Java 1.4 era.
The documentation was extended where I felt that a non-native English speaker may need more words. Also, some more examples were added to the JavaDoc documentation. Mainly the behavior of boolean to Integer/String etc. conversions were not defined when `trueValue` and `falseValue` and/or other parameters are the same.


garydgregory:
I would say to split this PR into two: Javadoc vs. code.

When I work in performance sensitive code, I have compiler warnings set up to flag boxing and unboxing; but in the end, I always have these warnings enabled. For now, I'd rather keep the code as is.

verhas:
I will split it into two PRs, I am absolutely in-line with your opinion.

As for the boxing and unboxing: I absolutely understand your concerns and I accept that such change needs more consideration. Only FYI: I tried to modify only those boxings and unboxings where afaik Java 8 and later compilers generate exactly the same JVM code after the modification. But I have not checked that.

PR #470, opened 2019-10-16
Boolean utils limited cleanup

verhas:
A limited cleanup where only the explicit unboxings are removed. This version and the `master`  version this branch was created produces the same class file. 

```
diff target/classes/org/apache/commons/lang3/BooleanUtils.class ../tmp/master/BooleanUtils.class
```

returns w/o message.


src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -4803,7 +4803,7 @@ public static Object toPrimitive(final Object array) {
         }
         final boolean[] result = new boolean[array.length];
         for (int i = 0; i < array.length; i++) {
-            result[i] = array[i].booleanValue();
+            result[i] = array[i];
```
verhas: This is included here because it is directly invoked from `BooleanUtils`. There are other similar redundant unboxings in this class in addition to this.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -4609,17 +4609,8 @@ public static int lastIndexOf(final short[] array, final short valueToFind, int
      * (index &lt; 0 || index &gt;= array.length), or if the array is {@code null}.
      * @since 2.1
      */
-    public static boolean[] toPrimitive(final Boolean[] array) {
```
garydgregory: You cannot delete public APIs, that breaks binary compatibility within a major release.

verhas: Oh. Fuck me!

Was an accident. I will fix.

verhas: I think you were looking at some outdated version of the request.

This method starts on the line 8886 in the current pull request and also on the current master.

The other possibility that I am totally missing something. One thing is for sure: I would never intentionally suggest the deletion or change of an API method.

Could you have a look at it again?

verhas:
I think this is still waiting to be merged. I do not see what to change.

garydgregory:
I am -1 to this kind of change in a low-level library like Commons Lang. We should make it obvious where conversions take place such that users and maintainers are aware of the trade-offs/costs.

verhas:
> I am -1 to this kind of change in a low-level library like Commons Lang. We should make it obvious where conversions take place such that users and maintainers are aware of the trade-offs/costs.

I truly believe in Java. That also means that I also believe that it is expressive enough so that developers who really know the language will and should understand where and how autoboxing happens.

But I can live with your implication that there are many users, even possibly maintainers whose knowledge related to Java is sub-par. So let it be as it is.

PR #471, opened 2019-10-18
[LANG-1177] Added indexesOf methods and simplified removeAllOccurences

lielfr:
Based of the opened issue. Tests for indexesOf were also added.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -4319,7 +4319,7 @@ public class ArrayUtils {
 
     /**
      * Finds the indices of the given value in the array starting at the given index.
-     * 
+     *
      * <p>
      * This method will return the indices of the values which fall between the region
      * defined by valueToFind - tolerance and valueToFind + tolerance, between the nearest integers.
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -4261,7 +4261,7 @@ public class ArrayUtils {
 
     /**
      * Finds the indices of the given value within a given tolerance in the array.
-     * 
+     *
      * <p>
      * This method will return all the indices of the value which fall between the region
      * defined by valueToFind - tolerance and valueToFind + tolerance, each time between the nearest integers.
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -3255,6 +3255,57 @@ public class ArrayUtils {
     }
 
     /**
+     * <p>Finds the indices of the given object in the array.
+     *
+     * <p>This method returns an empty BitSet for a {@code null} input array.
+     *
+     * @param array  the array to search through for the object, may be {@code null}
+     * @param objectToFind  the object to find, may be {@code null}
+     * @return a BitSet of all the indices of the object within the array,
+     *  an empty BitSet if not found or {@code null} array input
+     */
+    public static BitSet indexesOf(final Object[] array, final Object objectToFind) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -3254,6 +3254,57 @@ public static int indexOf(final Object[] array, final Object objectToFind, int s
...
 
+    /**
+     * <p>Finds the indices of the given object in the array.
+     *
+     * <p>This method returns an empty BitSet for a {@code null} input array.
+     *
+     * @param array  the array to search through for the object, may be {@code null}
+     * @param objectToFind  the object to find, may be {@code null}
+     * @return a BitSet of all the indices of the object within the array,
+     *  an empty BitSet if not found or {@code null} array input
```
Stzx: `@since` tag should be provided for newly provided methods.

lielfr: Okay, which version? 4.0?

Stzx: The next version is `3.10`.

lielfr: Done

garydgregory:
@lielfr 
Please rebase on master.

lielfr:
@garydgregory Should be okay now

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8202,20 +8654,8 @@ public static boolean isSorted(final boolean[] array) {
...
-        final int[] indices = new int[array.length - index];
-        indices[0] = index;
-        int count = 1;
-
-        while ((index = indexOf(array, element, indices[count - 1] + 1)) != INDEX_NOT_FOUND) {
-            indices[count++] = index;
-        }
-
-        return removeAll(array, Arrays.copyOf(indices, count));
+        BitSet foundBits = indexesOf(array, element);
```
garydgregory: How about inlining all of the 'foundBits'?

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -3254,6 +3254,59 @@ public static int indexOf(final Object[] array, final Object objectToFind, int s
...
+     *
+     * @param array  the array to search through for the object, may be {@code null}
+     * @param objectToFind  the object to find, may be {@code null}
+     * @param startIndex  the index to start searching at
+     * @return a BitSet of all the indices of the object within the array starting at the index,
+     *  an empty BitSet if not found or {@code null} array input
+     * @since 3.10
+     */
+    public static BitSet indexesOf(final Object[] array, final Object objectToFind, int startIndex) {
+        BitSet ret = new BitSet();
```
garydgregory: It seems to me "ret" is not a great name, how about "bitSet"?

lielfr: Done

src/main/java/org/apache/commons/lang3/ArrayUtils.java
lielfr: Done

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -3254,6 +3254,59 @@ public static int indexOf(final Object[] array, final Object objectToFind, int s
         return INDEX_NOT_FOUND;
     }
 
+    /**
+     * <p>Finds the indices of the given object in the array.
```
garydgregory: Oops, I did not notice before that you did not close any of these HTML tags. You need to close paragraphs.

lielfr: @garydgregory Fixed. Apparently this is a problem with other functions too

PR #472, opened 2019-10-24
Add more test inputs for ArrayUtils.toMap() method

apirom9:
Add more test inputs for ArrayUtils.toMap() method as follow:
- zero-length array
- all null values array
- duplicate keys in array

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
@@ -33,7 +33,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 
--- a/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
@@ -30,8 +30,10 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
```diff
@@ -264,6 +266,31 @@ public int hashCode() {
             }
         }});
         assertEquals("bar", map.get("foo"));
+
+        // Return empty map when got input array with length = 0
+        assertEquals(Collections.emptyMap(), ArrayUtils.toMap(new Object[0]));
+
+        // Test all null values
+        map = ArrayUtils.toMap(new Object[][] { {null, null}, {null, null} });
+        assertEquals(new HashMap<Object, Object>() {
```
garydgregory: Your expected map creation is too obtuse IMO, use the what Java provides instead:
```
Collections.singletonMap(null, null);
```

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
```diff
@@ -264,6 +266,31 @@ public int hashCode() {
...
+            {
+                put(null, null);
+            }
+        }, map);
+
+        // Test duplicate keys
+        map = ArrayUtils.toMap(new Object[][] { {"key", "value2"}, {"key", "value1"} });
+        assertEquals(new HashMap<Object, Object>() {
+
+            private static final long serialVersionUID = 1L;
```
garydgregory: See my previous comment.

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
apirom9: Fixed. Thanks

src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
apirom9: Fixed. Thanks

PR #473, opened 2019-10-24
[LANG-1466] excludeNullValues param in reflectionToString method

alasun:
excludeNullValues additional param in ToStringBuilder.reflectionToString


garydgregory:
I am -1 to adding a new method that just hard codes just the one argument.

PR #475, opened 2019-10-28
LANG-1495 Update EnumUtils.java

vlcheong:
Add Overloading Methods getEnum and getEnumIgnoreCase To EnumUtils

### DIFF
--- a/src/main/java/org/apache/commons/lang3/EnumUtils.java
+++ b/src/main/java/org/apache/commons/lang3/EnumUtils.java
@@ -133,6 +133,7 @@ public class EnumUtils {
      * @param enumName    the enum name, null returns default enum
      * @param defaultEnum the default enum
      * @return the enum, default enum if not found
+     * @since 3.10
      */
     public static <E extends Enum<E>> E getEnum(final Class<E> enumClass, final String enumName, final E defaultEnum) {
         if (enumName == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -119,13 +119,29 @@ public EnumUtils() {
...
+     * <p>Gets the enum for the class, returning {@code defaultEnum} if not found.</p>
+     *
+     * <p>This method differs from {@link Enum#valueOf} in that it does not throw an exception
+     * for an invalid enum name.</p>
+     *
+     * @param <E> the type of the enumeration
+     * @param enumClass   the class of the enum to query, not null
+     * @param enumName    the enum name, null returns default enum
+     * @param defaultEnum the default enum
+     * @return the enum, default enum if not found
```
Stzx: Provide a `@since` tag for the new method.

vlcheong: @Stzx, may I know what version number should I put ?

Stzx: `3.10` :)

vlcheong: @Stzx, Thanks for your guidance.

PR #477, opened 2019-11-02
Lang 1463: StringUtils abbreviate returns String of length greater than maxWidth

bbeckercscc:
added handling of abbrevMarker being an empty string

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -292,9 +292,7 @@ public class StringUtils {
      */
     public static String abbreviate(final String str, final String abbrevMarker, final int maxWidth) {
         return abbreviate(str, abbrevMarker, 0, maxWidth);
-        // TODO hackathon String Method
     }
-
     /**
      * <p>Abbreviates a String using a given replacement marker. This will turn
      * "Now is the time for all good men" into "...is the time for..." if "..." was defined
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -332,10 +331,13 @@ public static String abbreviate(final String str, final String abbrevMarker, fin
      * @since 3.6
      */
     public static String abbreviate(final String str, final String abbrevMarker, int offset, final int maxWidth) {
-        if (isEmpty(str) || isEmpty(abbrevMarker)) {
+        if (isEmpty(str) && isEmpty(abbrevMarker)) {
+            return str;
+        } else if (isNotEmpty(str) && "".equals(abbrevMarker) && maxWidth>0) {
```
garydgregory: No magic strings, this class defines `EMPTY` for this purpose. You are missing spaces around `>`

bbeckercscc: I added the spaces. What are you implying about the magic strings? Forgive me, I'm not as experienced of a programmer.

garydgregory: A "magic" string or number is a value that is hard-coded, as opposed to reusing a constant, in this case, `EMPTY` instead of `""`.

bbeckercscc: Thanks for the explanation, I removed the magic string

PR #478, opened 2019-11-02
StringUtils.abbreviate handles empty abbrevMarker

jordanjenkins:
StringUtils.abbreviate would not abbreviate properly when an empty
String was passed to abbrevMarker or abbrevMarker is null
Expected behaviour:
abbreviate("much too long text", "", 13).equals("much too long")
Actual behaviour:
abbreviate("much too long text", "", 13).equals("much too long text")

Behaves the same when abbrevMarker is null.

Added if statements to handle empty and null abbrevMarker
Passed all tests

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -332,8 +332,15 @@ public class StringUtils {
      * @since 3.6
      */
     public static String abbreviate(final String str, final String abbrevMarker, int offset, final int maxWidth) {
-        if (isEmpty(str) || isEmpty(abbrevMarker)) {
+        if (isEmpty(str)) {
             return str;
+        }
+
+        if (abbrevMarker == null){
+            return str.substring(0,maxWidth + 1);
+        }
+        if (isEmpty(abbrevMarker)){
+            return str.substring(0, maxWidth);
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -332,10 +332,17 @@ public static String abbreviate(final String str, final String abbrevMarker, fin
      * @since 3.6
      */
     public static String abbreviate(final String str, final String abbrevMarker, int offset, final int maxWidth) {
-        if (isEmpty(str) || isEmpty(abbrevMarker)) {
+        if (isEmpty(str)) {
             return str;
         }
 
+        if (abbrevMarker == null){
+            return str.substring(0, maxWidth + 1);
```
Stzx:
```java
StringUtils.abbreviate("a", null, 0, 10);
```

May throw a `StringIndexOutOfBoundsException`.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -332,10 +332,17 @@ public static String abbreviate(final String str, final String abbrevMarker, fin
...
+        }
+        if (isEmpty(abbrevMarker)){
+            return str.substring(0, maxWidth);
```
Stzx: May throw a `StringIndexOutOfBoundsException`.

PR #481, opened 2019-11-23
[LANG-1501] Sort fields in ReflectionToStringBuilder for deterministic order

contextshuffling:
This fixes [LANG-1501](https://issues.apache.org/jira/projects/LANG/issues/LANG-1501?filter=allissues).

In this PR, it proposes:
* Sort the fields with their names for a deterministic order (alphabetical)
* Modify some test assertions to match the behavior


TranNgocKhoa:
Could I get a JSON object with its field sorted with this order?
 - declare fields order in the class.
 - declare fields order in parents class, from direct parent to other parents.

PR #482, opened 2019-12-29
LANG-1504 - Adding steps feature to StopWatch

topera:


garydgregory:
FYI, the conventional name for what you call "step" is "split", sometimes also "lap" (less modern terminology.)

topera:
> FYI, the conventional name for what you call "step" is "split", sometimes also "lap" (less modern terminology.)

I didn`t know that, thanks for sharing :)

I take a look now and the StopWatch already has the **split** method, but works in a different way.

Some differences:
1. With **split** we need to call **getSplitTime** while we are in the middle of split (I mean, before calling **unsplit**). In the "steps" solution we can get all times in a separated report, at the end of our performance test, for example.

2. With **split** we can have just one split each time. With **steps**, many

3. With **steps** we can label each step with a name

4. Split counts the time since the StopWatch started, not the time between splits.
Example

       ```
 final StopWatch sw = StopWatch.createStarted();

        Thread.sleep(100);

        sw.split();

        Thread.sleep(200);

        sw.split();

        Thread.sleep(400);

        System.out.println("splitTime: " + sw.getSplitTime()); // returns 300: the time between start and latest split
        System.out.println("time: " + sw.getTime()); // returns 700

```

So, my proposal is to use a new name no change the idea of the old split.
Possibilities: **lap** (as you said), **step**, **stage**

**Comment by garydgregory** [comment]
Since we already have split() we should keep the name. Another idea, is
that split times should be kept in history list. But then, I'd want a list
of pairs (message-split time) and it starts to feel like feature creep ;-)

I'm not sure how handy this would as I do not have a use case for it.

I'd like to cut a RC later this week. There is already plenty for a release
ATM but let's keep the discussion going :-)

Gary

**Comment by topera** [comment]
Hello Mr. Gregory,

Thank you for your answer.
I'm also glad that part of my suggestion was already accepted and done on LANG-1506 (putting labels on the timer)

About your points:
* I agree to use **split** name.
* About **feature creep**, IMHO this is a good feature: with small lines of code we can make the StopWatch much more useful, mainly to detect performance bottlenecks on the code.
* Use case: the developer wants to know which operation is slower. He/she can simply add many split() calls and label then as follow:

```
        final StopWatch watch = StopWatch.createStarted();

        watch.split("Baking cookies");
        ...
        watch.split("Serving cookies");
        ...
        watch.split("Eating cookies");
        ...
        watch.getSplitReport();

```
The output would be (with or without the percentage indicator)

```
        Baking cookies  00:07:00.000 - 70%
        Serving         00:01:00.000 - 10%
        Eating          00:02:00.000 - 20%
```

If you and other members of the team agrees, I'll proceed with the implementation.
Note: this is my first try on open source projects, so please sorry if I'm skipping some important step on the process.

Regards

**Comment by kinow** [comment]
Haven't had time to review this change @topera. But

>Note: this is my first try on open source projects, so please sorry if I'm skipping some important step on the process.

you are doing great, being patient, and trying to express your point :+1: thanks for that.

It would be great if you fixed the conflicting files in this PR. There are two files in conflict, see screenshot:

image

**Comment by garydgregory** [comment]
Let's have this PR's conflicts resolved, rename "steps" to "splits" to match the current terminology and then we can have another look.

**Comment by topera** [comment]
PR updated according to suggestions

File: src/main/java/org/apache/commons/lang3/time/StopWatch.java [hunk 1]
```diff
@@ -523,10 +586,17 @@ public void stop() {
         }
         if (this.runningState == State.RUNNING) {
             this.stopTime = System.nanoTime();
+            split("");
```
garydgregory: Please reuse StringUtils.EMPTY.

garydgregory:
Thanks for the update. I will review later this week but consider java.time classes for more precision.

garydgregory:
As you can see here "All checks have failed", please fix those up. 
https://travis-ci.org/apache/commons-lang/builds/654709829?utm_source=github_status&utm_medium=notification

src/main/java/org/apache/commons/lang3/time/StopWatch.java
topera: Done

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -606,4 +681,175 @@ public void unsplit() {
...
+        return getProcessedSplits(true);
+    }
+
+    /**
+     * Stops the watch and returns the list of splits with duration on each split
+     *
+     * @param useNanos if must use nanoseconds precision
+     * @return list of splits
+     */
+    public List<Split> getProcessedSplits(boolean useNanos) {
```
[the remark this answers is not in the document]
topera: Good idea, since it makes the code more clear. I'll change here. Thank you!

edelgadoh:
Hi everyone, I also wanted to use this split feature to simplify the code when using in many places, I saw this PR was opened 5 years ago, @topera is there anything pending to push from your side? it seems there are conflicts to resolve on it.

@garydgregory it seems to me that everything was resolved, is there anything that is pending? Thanks!

PR #484, opened 2020-01-07
LANG-1513 ObjectUtils: Get first non-null supplier value

bonii-xx:
Lazy evaluation variant of firstNonNull to allow using it for expensive operations.
When the first value of different options is required, but they're expensive to calculate.

I tried to match the code and test style present.


bonii-xx:
Merged with https://github.com/apache/commons-lang/commit/a606a0328ba4b3061c5bc810a31418fa4251ea16

PR #485, opened 2020-01-19
[LANG-1514] make test more stable by wrapping assertions in hashset

contextshuffling:
This fixes [LANG-1514](https://issues.apache.org/jira/projects/LANG/issues/LANG-1514?filter=allissues)

This PR proposes to replace the expected and actual values of the assertions in the tests by wrapping each value into a HashSet and then checking if the two sets are equal, thus ignoring the order of elements in the collections.


contextshuffling:
@kinow Do you think this change is ready to be merged or is there anything else need to address?

kinow:
Sorry I forgot about it @contextshuffling . Done! Thanks!!

PR #487, opened 2020-02-03
unneccessary parentheses removed

verhas:


garydgregory:
This kind of change for this class has been applied in a different commit.

PR #488, opened 2020-02-03
more unnecessary parentheses removed

verhas:


garydgregory:
@verhas,
Thank you for this PR.
I did this as part of another clean up commit. 
Gary

PR #489, opened 2020-02-03
Code simplification using the methods that we have instead

verhas:
of retyping their functionality. We expect the users of the library to use the provided methods, then we have to use these methods in the library code where it absolutely makes sense.


garydgregory:
@verhas,
May you please resolve conflicts here?
Thank you!
Gary

verhas:
I issued the 

```
git remote add upstream git@github.com:apache/commons-lang.git
git fetch upstream
git checkout functions-vp-20200203-3
git merge upstream/master
```
commands. Then I resolved the conflicts in IntelliJ that were only in `Functions.java` and now it shows three other files changed. I do not understand why. Probably I have to learn more git. Sorry.

I do not know if it really messes something up if it gets merged or just GIT is showing some diffs it should not.

garydgregory:
@verhas this PR looks all messed up compared to what was here before. Learning more about git is a good thing no matter what ;-)

PR #492, opened 2020-02-08
style(org.apache.commons.lang3): rename method from removeAllOccurenc…

BillCindy:
rename method from removeAllOccurences to **removeAllOccurrences**

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -802,8 +802,8 @@ public class ArrayUtils {
 
     /**
      * <p>Adds all the elements of the given arrays into a new array.
-     * <p>The new array contains all of the element of {@code array1} followed
-     * by all of the elements {@code array2}. When an array is returned, it is always
+     * <p>The new array contains all the element of {@code array1} followed
+     * by all the elements {@code array2}. When an array is returned, it is always
      * a new array.
      *
      * <pre>
@@ -831,8 +831,8 @@ public class ArrayUtils {
 
     /**
      * <p>Adds all the elements of the given arrays into a new array.
-     * <p>The new array contains all of the element of {@code array1} followed
-     * by all of the elements {@code array2}. When an array is returned, it is always
+     * <p>The new array contains all the element of {@code array1} followed
+     * by all the elements {@code array2}. When an array is returned, it is always
      * a new array.
      *
      * <pre>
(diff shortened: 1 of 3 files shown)


BillCindy:
> Oh, good catch on the typo. If it's been released, we may have to wait for a new major release before the merge I think.

ok, I will continue to contribute to this project

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -802,8 +802,8 @@ private static Object add(final Object array, final int index, final Object elem
 
     /**
      * <p>Adds all the elements of the given arrays into a new array.
-     * <p>The new array contains all of the element of {@code array1} followed
-     * by all of the elements {@code array2}. When an array is returned, it is always
+     * <p>The new array contains all the element of {@code array1} followed
```
garydgregory: This is still wrong: "The new array contains all the element of {@code array1} followed" -> "The new array contains all the element**s** of {@code array1} followed"

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -5322,7 +5322,7 @@ static Object removeAll(final Object array, final int... indices) {
      * @return A new array containing the existing elements except the occurrences of the specified element.
      * @since 3.5
      */
-    public static byte[] removeAllOccurences(final byte[] array, final byte element) {
+    public static byte[] removeAllOccurrences(final byte[] array, final byte element) {
```
garydgregory: -1 You cannot delete a public API as this will break binary compatibility. You _can_ create a new API with the corrected name and then have the old API delegate to the new one.

src/test/java/org/apache/commons/lang3/ArrayUtilsRemoveTest.java
```diff
@@ -403,189 +403,189 @@ public void testRemoveElementShortArray() {
     @Test
     public void testRemoveAllBooleanOccurences() {
         boolean[] a = null;
-        assertNull(ArrayUtils.removeAllOccurences(a, true));
+        assertNull(ArrayUtils.removeAllOccurrences(a, true));
```
garydgregory: If you create a new API (see my other comment) with the corrected name, either test both the new and old APIs next to each other, or, clone the test methods and also correct their names.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
BillCindy: thanks for your remind! I got it

src/test/java/org/apache/commons/lang3/ArrayUtilsRemoveTest.java
BillCindy: ok，i will do it as you wish

src/main/java/org/apache/commons/lang3/ArrayUtils.java
BillCindy: I will make correct ，thanks

garydgregory:
https://issues.apache.org/jira/browse/LANG-1529

PR #495, opened 2020-02-21
LANG-1519 Add Integer Utils to library

nnivruth:
Create Integer Utils class for useful integer operations and a placeholder for adding more useful/important utility functions

### DIFF
--- a/src/main/java/org/apache/commons/lang3/IntegerUtils.java
+++ b/src/main/java/org/apache/commons/lang3/IntegerUtils.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.lang3;
+
+/**
+ * <p>Operations on int primitives and Integer objects.</p>
+ * <p>
+ * <p>This class tries to handle {@code null} input gracefully.
+ * An exception will not be thrown for a {@code null} input.
+ * Each method documents its behaviour in more detail.</p>
+ * <p>
+ * <p>#ThreadSafe#</p>
+ *
+ * @since 2.0
+ */
+public class IntegerUtils {
+
+    /**
+     * <p>{@code IntegerUtils} instances should NOT be constructed in standard programming.
+     * Instead, the class should be used as {@code IntegerUtils.isPositive(10);}.</p>
+     * <p>
+     * <p>This constructor is public to permit tools that require a JavaBean instance
+     * to operate.</p>
+     */
+    public IntegerUtils() {
+        super();
+    }
+
+    // int Integer methods
+    //-----------------------------------------------------------------------
+
+    /**
+     * <p>Checks if an {@code Integer} value is positive,
+     * handling {@code null} by returning {@code false}.</p>
+     *
+     * @param integer the integer to check, {@code null} returns {@code false}
+     * @return {@code true} only if the input is non-null and greater than 0
+     */
+    public static boolean isPositive(final Integer integer) {
+        return integer != null && integer > 0;
+    }
+
+    /**
+     * <p>Checks if an {@code Integer} value is not positive,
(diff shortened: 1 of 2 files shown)


nnivruth:
https://issues.apache.org/jira/browse/LANG-1519

ameyjadiye:
Shouldn't this be the part of Math-related projects? like common numbers?

garydgregory:
-1 in Commons Lang, we had a `NumberUtil`s class in 2.x but it was deprecated and removed in 3.0. That code now lives in Apache Commons Math IIRC. Also, it does not make sense as typed to `Integer` when you can type to its `Number` superclass.

nnivruth:
@ameyjadiye @garydgregory refactored & moved to NumberUtils

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1822,4 +1824,125 @@ public static int compare(final short x, final short y) {
...
+            return number.byteValue() == BYTE_ZERO;
+        } else if (number instanceof Short) {
+            return number.shortValue() == SHORT_ZERO;
+        } else if (number instanceof Float) {
+            return number.floatValue() == FLOAT_ZERO;
+        } else if (number instanceof Double) {
+            return number.doubleValue() == DOUBLE_ZERO;
+        }
+
+        return false;
```
garydgregory: This is bogus IMO, if I create a subclass of `Number`, this method will always return false even if my `Number`'s inValue() returns 0. Why is this whole method's implementation not simply `return number.doubleValue() == DOUBLE_ZERO` ?

nnivruth: agree, updated the func.

garydgregory:
What should these method return when a double is not-a-number?

nnivruth:
> What should these method return when a double is not-a-number?

@garydgregory, great question. IMO the methods isZero(), isPositive(), isNegative() should return false for float/double NaNs & vice versa for isNotZero(), isNotPositive(), isNotNegative(). I've added these to the test cases as well.

garydgregory:
What about other JRE numbers like atomic numbers and big numbers? How should those be treated?

nnivruth:
@garydgregory one way to handle special numbers (BigInteger, BigDecimal, AtomicInteger, AtomicLong) is to add them as special instance checks (additional else-if blocks in each method) but not sure if that's the best way. do you have any suggestions/feedback?

nnivruth:
> What about other JRE numbers like atomic numbers and big numbers? How should those be treated?

@garydgregory one way to handle special numbers (BigInteger, BigDecimal, AtomicInteger, AtomicLong) is to add them as special instance checks (additional else-if blocks in each method) but not sure if that's the best way. do you have any suggestions/feedback?

garydgregory:
Hi @nnivruth 
In general, since we are a general purpose library, I would say that you should consider all subclasses of `Number` that are built-in the Java 8 JRE with maybe a bias to considering only what is in Java 11's `java.base` module in order to make it easier on us going forward though I would imagine that our requiring Java 11 as a base requirement is a long ways off (just a guess on that one.)

nnivruth:
> Hi @nnivruth
> In general, since we are a general purpose library, I would say that you should consider all subclasses of `Number` that are built-in the Java 8 JRE with maybe a bias to considering only what is in Java 11's `java.base` module in order to make it easier on us going forward though I would imagine that our requiring Java 11 as a base requirement is a long ways off (just a guess on that one.)

Thanks @garydgregory ! I've also added Long Accumulator, Long Adder, Double Accumulator, Double Adder which covers [known Number classes](https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html).

garydgregory:
@nnivruth ,
Thank you for your updates. The code covers all the cases it seems which is nice. The drawback ATM is that it does not feel very OO. I am wondering if all of this could be implemented in terms of a new utility method called compare(Number, Number) which would allow the isZero, isPositive and isNegative to be implemented by calling compare(aNumber, ZERO). More thinking and experimenting needed...

This is all non-trivial, so I think I might cut a 3.10 RC which I am late on and this PR would end up in 3.11.

nnivruth:
@garydgregory ,
Been thinking and trying to make these functions more OO and rely on a single method but seeing some difficulties with type casting (esp for special types of number implementations - BigDecimal, etc), NaNs and custom Number implementations. 

Some examples:

- the isZero() function with the current impl does a direct equality check and returns true when passed with (-0.0d). IMO this is acceptable and correct but with the new impl it'll return false

- the functions isZero(), isPositive(), isNegative() return false for float/double NaNs & vice versa for isNotZero(), isNotPositive(), isNotNegative() but with compare we might have to add special checks for these scenarios


your thoughts/inputs on how to proceed?

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
```diff
@@ -1575,4 +1582,867 @@ public void compareByte() {
         assertEquals(0, NumberUtils.compare((byte) 113, (byte) 113));
         assertTrue(NumberUtils.compare((byte) 123, (byte) 32) > 0);
     }
+
+    @Test
+    void testIsZero() {
```
HubertWo, line 1587: What about splitting this to small tests instead of one long test?

PR #496, opened 2020-02-29
LANG-1523: Avoid unnecessary allocation in StringUtils.wrapIfMissing.

nstdio:
Now `StringUtils#wrapIfMissing(String, char)` and `StringUtils#wrapIfMissing(String, String)` does not allocate buffer when input is already wrapped.

Here are the benchmark results on my machine:

```
Benchmark                    (input)   Mode  Samples    Score  Score error   Units
optimized    /should not be wrapped/  thrpt       20  357.849        4.525  ops/us
optimized          should be wrapped  thrpt       20   33.779        0.246  ops/us
original     /should not be wrapped/  thrpt       20   35.581        0.236  ops/us
original           should be wrapped  thrpt       20   33.671        0.360  ops/us
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -9448,6 +9448,8 @@ public class StringUtils {
      * Wraps a string with a char if that char is missing from the start or end of the given string.
      * </p>
      *
+     * <p>A new {@code String} will not be created if {@code str} is already wrapped.</p>
+     *
      * <pre>
      * StringUtils.wrapIfMissing(null, *)        = null
      * StringUtils.wrapIfMissing("", *)          = ""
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -9508,20 +9514,27 @@ public static String wrapIfMissing(final String str, final char wrapWith) {
...
             return str;
         }
+        
+        final boolean wrapStart = !str.startsWith(wrapWith);
+        final boolean wrapEnd = !str.endsWith(wrapWith);
+        if (!wrapStart && !wrapEnd) {
+            return str;
+        }
+        
         final StringBuilder builder = new StringBuilder(str.length() + wrapWith.length() + wrapWith.length());
```
nstdio: I don't know if there is a practical reason to create `StringBuilder` with exact capacity, because it `str` can start with `wrapWith` but not end with it and vice versa.

kinow: Yeah, looks like this is the maximum capacity, not necessarily the exact capacity we expect to have. Probably harmless/fine?

nstdio:
Done.
@kinow

src/main/java/org/apache/commons/lang3/StringUtils.java
garydgregory: Since the behavior has changed, I would expect a clarification in the Javadoc. Plus a matching test.

nstdio: @garydgregory Sure, that makes sense. Done.

RameshDhungel:
Hello, I just saw this issue in Jira, and if I understand the issue correctly. The  StringUtils.wrapIfMissing is making a StringBuilder even when a string is already wrapped for example Str "xabx" with wrapper "x". Rather than not creating a instance of String Builder. What we can do is check if the string is already wrapped, if so then just return the String. This would avoid the String Builder being created.

garydgregory:
@RameshDhungel I merged this PR four days ago. If you have something else in mind, please offer it up in a new PR. Thank you.

RameshDhungel:
Okay thank you, I did pull an new PR with the changes I described earlier. If you could let me know if I made any mistake I would appreciate it thank you.

PR #499, opened 2020-03-01
LANG-1521 spring clean .gitignore

nhojpatrick:
Handles https://issues.apache.org/jira/browse/LANG-1521

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,20 +1,39 @@
-# Maven build files
-target
-*.log
-maven-eclipse.xml
-build.properties
-site-content
-*~
 
-# IntelliJ IDEA files
-.idea
-.iws
+*~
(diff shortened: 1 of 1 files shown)


nhojpatrick:
> I'm fine adding more items to gitignore. But in the past we had (either on GH or mailing list I think) decided to add IDE and other tools not in our build to the user global gitignore (on Linux ~/.gitignore). Let's wait to see what others think. But thanks for the PR!

Happy to remove Operating System, or IDE specific as should be in users global .gitignore.

What every happens I'll keep the others in sync too;
codec https://github.com/apache/commons-codec/pull/37
collections https://github.com/apache/commons-collections/pull/134
io https://github.com/apache/commons-io/pull/103

PR #501, opened 2020-03-05
Expand Streams functionality

Isira-Seneviratne:
Add failable streams, map and flatMap operations for primitive types and move streams and functional interfaces to org.apache.commons.lang3.stream and org.apache.commons.lang3.function, respectively.


aherbert:
Moving classes is changing the public API and breaking binary compatibility.

Run this to see what is not allowed:
```
mvn clean clirr:check
```

You should revert the changes in the code that use * imports. These will break the code style checks when binary compatibility passes.

PR #502, opened 2020-03-07
 Added 1 and 0 in toBooleanObject(final String str) 

dscham:
I stubmled over this during the week and am wondering why 1 and 0 are not a default true and false representation. So I created this PR. I'm not strongly opnionated about it. So if there is a good reason for 1 and 0 not beeing a good case here, feel free to close it.

I think, parsing the string to an int before calling toBoolean is unnecessary. And it should be common enough that you shouldn't have to call: toBooleanObject(final String str, final String trueString, final String falseString, final String nullString)


garydgregory:
OK by me as it also matches `javax.xml.bind.DatatypeConverter.parseBoolean(String)`

garydgregory:
https://issues.apache.org/jira/browse/LANG-1526

PR #503, opened 2020-03-08
[LANG-1523] Avoid unnecessary allocation in StringUtils.wrapIfMissing

RameshDhungel:
It seems like the String Builder is being created after checking if the str or the wrapper is null. What we can do is check if the str is already wrapped, if so then just return the str. This would avoid the String Builder being created.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -9417,6 +9417,9 @@ public class StringUtils {
         if (isEmpty(str) || wrapWith == CharUtils.NUL) {
             return str;
         }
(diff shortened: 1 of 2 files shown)


garydgregory:
@RameshDhungel 
I believe this PR conflicts with master because this change has already been merged through https://github.com/apache/commons-lang/pull/496

RameshDhungel:
Oh okay, so this issue is already resolved then right?

garydgregory:
> Oh okay, so this issue is already resolved then right?

@RameshDhungel 
It's up to you to verify that your specific use cases are accounted for and then close this PR if you are OK with the state of git master ;-)

garydgregory:
No feedback for a couple of weeks, see comments.

PR #506, opened 2020-03-28
LANG-1530 another method like formatDurationWords with 

mikolajmitura:
added new method 'formatDurationWordsWithMs' for format duration: 
0 days 0 hours 0 minutes 0 seconds 0 milliseconds
still exists old method 'formatDurationWords' which format as below:
0 days 0 hours 0 minutes 0 seconds

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -17,10 +17,13 @@
 package org.apache.commons.lang3.time;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.List;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -160,6 +163,7 @@ public static String formatDuration(final long durationMillis, final String form
 
     /**
      * <p>Formats an elapsed time into a pluralization correct string.</p>
+     * <p>Returns duration in format '0 days 0hours 0 minutes 0 seconds'</p>
```
kinow: s/0hours/0 hours

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -171,56 +175,96 @@ public static String formatDuration(final long durationMillis, final String form
...
+            final boolean suppressLeadingZeroElements,
+            final boolean suppressTrailingZeroElements) {
+        return formatDurationWordsWithMs(durationMillis, false,
+                                         suppressLeadingZeroElements, suppressTrailingZeroElements);
+    }
+
+    /**
+     * <p>Formats an elapsed time into a pluralization correct string.</p>
+     * <p>Returns duration in format '0 days 0 hours 0 minutes 0 seconds 0 milliseconds'</p>
+     * It removes all leading time units with 0 value.
```
kinow: Dangling text. Was it supposed to be within the previous `<p>`?

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
mikolajmitura: fixed all 0hours

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
mikolajmitura: aligned to previous <p>

garydgregory:
Hi All,

Before we make this class heavier, let's step back and consider what Java 8's own [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) and [Period](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) classes have to offer and how this class should interact/interface with these.

kinow:
> Hi All,
> 
> Before we make this class heavier, let's step back and consider what Java 8's own [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) and [Period](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) classes have to offer and how this class should interact/interface with these.

+1

I had a quick look at the code, but didn't want to prepare a review before checking if we run the risk of needing one more test to display time zone with micro seconds, and yet more methods for other things.

Good idea on looking at the SDK classes first.

kinow (commented):
@mikolajmitura whenever you have time, take a look at @garydgregory 's last comment about SDK classes :+1:

PR #508, opened 2020-04-15
LANG-1481 Add StringUtils.split() methods with String[] input.

RatishBansal:
Leveraging the existing split(string,separator) method, to split each element of input array


amarlearning:
@RatishBansal I have just one small doubt, how do you plan to cover these cases mentioned [here](https://issues.apache.org/jira/browse/LANG-1481).
```java
StringUtils.split(["a.b"],["c"], '.')     = ["a", "b", "c"]
StringUtils.split(["a"],["b"],["c"], '.') = ["a", "b", "c"]
```

RatishBansal:
> @RatishBansal I have just one small doubt, how do you plan to cover these cases mentioned [here](https://issues.apache.org/jira/browse/LANG-1481).
> 
> ```java
> StringUtils.split(["a.b"],["c"], '.')     = ["a", "b", "c"]
> StringUtils.split(["a"],["b"],["c"], '.') = ["a", "b", "c"]
> ```

@amarlearning as I mentioned in the [comment](https://issues.apache.org/jira/browse/LANG-1481?focusedCommentId=17078554&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17078554) as well, my understanding of the requirement is to write split method with String[] as input. I don't think the method should have varargs String[]. My understanding might be wrong, but I think the extra square brackets might be a mistake. I think @garydgregory would be the right person to clear the requirement.

PR #509, opened 2020-04-16
Update DurationFormatUtils.java

pedro-borges:
It is not correct to return 0 seconds for a 100ms duration.


garydgregory:
-1 This is wrong IMO, not only we would be changing the format of the returned string, but the information would still be incorrect.

An alternative would be to change:

```
"d' days 'H' hours 'm' minutes 's' seconds'"
```
to
```
"d' days 'H' hours 'm' minutes 's' seconds' 'm' milliseconds"
```

PR #510, opened 2020-04-17
[LANG-1535] Add containsAnyIgnoreCase() method to StringUtils.

Isira-Seneviratne:
Add a containsAnyIgnoreCase() method to StringUtils, together with its associated tests.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsContainsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsContainsTest.java
@@ -293,6 +293,25 @@ public class StringUtilsContainsTest  {
     }
 
     @Test
+    public void testContainsAnyIgnoreCase_StringStringArray() {
+        assertFalse(StringUtils.containsAnyIgnoreCase("", (String[]) null));
+        assertFalse(StringUtils.containsAnyIgnoreCase("", new String[0]));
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/StringUtilsContainsTest.java
```diff
@@ -292,6 +292,28 @@ public void testContainsIgnoreCase_StringString() {
         assertTrue(StringUtils.containsIgnoreCase("xabcz", "ABC"));
     }
 
+    @Test
+    public void testContainsAnyIgnoreCase_StringStringArray() {
+        assertFalse(StringUtils.containsAnyIgnoreCase(null, (String[]) null));
+        assertFalse(StringUtils.containsAnyIgnoreCase(null, new String[0]));
+        assertFalse(StringUtils.containsAnyIgnoreCase(null, new String[] { "hello" }));
```
namannigam, line 299: these would precisely short-circuit at the same condition as with the assertion before these and hence might just be redundant

Isira-Seneviratne: Okay, I'll remove them.

arturobernalg:
@garydgregory 
Already implemented.

https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringUtils.java#L1178

Can close the PR

PR #512, opened 2020-04-17
Simplify some usages of the ternary operator with calls to Math.max() and Math.min()

Isira-Seneviratne:


Isira-Seneviratne:
> 
> 
> Code changes look good. I think for this one we better put in an issue. Could you create an issue in JIRA for lang, and check if it's covered by our unit tests already, please @Isira-Seneviratne ?
> 
> Thanks!

Sure!

PR #514, opened 2020-04-17
[LANG-1537]: Simplify a null check in the private replaceEach() method of StringUtils.

Isira-Seneviratne:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -6723,8 +6723,7 @@ private static String replaceEach(
         // index of replace array that will replace the search string found
         // NOTE: logic duplicated below START
         for (int i = 0; i < searchLength; i++) {
-            if (noMoreMatchesForReplIndex[i] || searchList[i] == null ||
-                    searchList[i].isEmpty() || replacementList[i] == null) {
+            if (noMoreMatchesForReplIndex[i] || isEmpty(searchList[i])|| replacementList[i] == null) {
```
namannigam, line 6726: nit: need to format correctly

Isira-Seneviratne: I missed that, thanks!

kinow:
Looks like a good idea to use the `isEmpty` method of the same class. Created a placeholder issue in JIRA. Let's see what others think :+1: thanks for all your great work @Isira-Seneviratne

PR #515, opened 2020-04-17
Remove unnecessary unboxing in BooleanUtils.

Isira-Seneviratne:


garydgregory:
-1 from my POV. In a low-level library like Commons Lang we should make explicit where we mix primitives and Objects. I want to be made aware of where the boxing and unboxing do in fact take place, which will have consequences depending on some use cases. Boxing is not just syntactical sugar.

kinow:
> -1 from my POV. In a low-level library like Commons Lang we should make explicit where we mix primitives and Objects. I want to be made aware of where the boxing and unboxing do in fact take place, which will have consequences depending on some use cases. Boxing is not just syntactical sugar.

I thought that would be OK given we were checking for `null` in all the cases, but agree on the part about being a low level library and being explicit on primitive vs. Object.

Isira-Seneviratne:
IntelliJ flagged the unboxing operations as unnecessary.

kinow:
> IntelliJ flagged the unboxing operations as unnecessary.

That's OK. There are other warnings in IntelliJ, Eclipse, SonarCloud, and other tools that perform static analysis that are not always addressed in Commons (and in other projects too, run a static analyser through Python Pandas and you will find a lot of warnings).

Commons has several low level libraries, like Lang. Where there is an explicit difference between objects/primitives since a long time ago. The final byte code might look the same, but the code base is probably distinguishing both. So no need to change that :+1:

garydgregory:
Closing...

PR #516, opened 2020-04-17
Simplify null checks in Triple.hashCode() using Objects.hashCode().

Isira-Seneviratne:


kinow:
Thanks for your contributions @Isira-Seneviratne !

PR #519, opened 2020-04-19
Alter assertions in CharUtilsTest to get the class of CharUtils instead of BooleanUtils.

Isira-Seneviratne:


Isira-Seneviratne:
I encountered it while converting BooleanUtils to Kotlin.

PR #520, opened 2020-04-19
Implement Utils classes in Kotlin.

Isira-Seneviratne:
Issue: https://issues.apache.org/jira/projects/LANG/issues/LANG-1538


Isira-Seneviratne:
Maybe it could be implemented as a separate library?

pom.xml
```diff
@@ -74,7 +71,7 @@
       <name>Henri Yandell</name>
       <id>bayard</id>
       <email>bayard@apache.org</email>
-      <organization />
+      <organization/>
```
garydgregory, line 74: @Isira-Seneviratne 
Please do not change this formatting.

pom.xml
```diff
@@ -553,6 +550,12 @@
       <scope>test</scope>
     </dependency>
 
+    <!-- Kotlin -->
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-stdlib-jdk8</artifactId>
+      <version>${kotlin.version}</version>
+    </dependency>
   </dependencies>
```
garydgregory, line 559: -1, that's a bridge too far IMO. If we want to create a Kotlin Commons Lang, we can talk about that on the developer's mailing list...

pom.xml
Isira-Seneviratne: That was formatted by IntelliJ.

garydgregory: It does not matter who formatted what, it's still your PR ;-)

PR #521, opened 2020-04-20
Simplify some if statements in StringUtils.

Isira-Seneviratne:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -330,11 +330,9 @@ public static String abbreviate(final String str, final String abbrevMarker, fin
      * @since 3.6
      */
     public static String abbreviate(final String str, final String abbrevMarker, int offset, final int maxWidth) {
-        if (isEmpty(str) && isEmpty(abbrevMarker)) {
-            return str;
-        } else if (isNotEmpty(str) && EMPTY.equals(abbrevMarker) && maxWidth > 0) {
-            return str.substring(0, maxWidth);
-        } else if (isEmpty(str) || isEmpty(abbrevMarker)) {
+        if (isNotEmpty(str) && EMPTY.equals(abbrevMarker) && maxWidth > 0) {
```
garydgregory, line 333: How about `EMPTY.equals(abbrevMarker)` -> `isEmpty(abbrevMarker)` ?

Isira-Seneviratne: That change would mean that `abbrevMarker` is potentially null.

garydgregory: @Isira-Seneviratne You are correct!

PR #523, opened 2020-04-20
Rewrite the finding of matching Locales in LocaleUtils using the Streams API.

Isira-Seneviratne:


src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -295,14 +297,13 @@ public static boolean isAvailableLocale(final Locale locale) {
...
-            }
+            // Allows Locales to be processed using multiple threads.
+            langs = availableLocaleList().parallelStream()
+                    .filter(locale -> countryCode.equals(locale.getCountry()))
+                    .filter(locale -> locale.getVariant().isEmpty())
+                    .collect(Collectors.toList());
+            // Collectors.toUnmodifiableList() is an alternative, but it is only
+            // available starting with Java 10.
             langs = Collections.unmodifiableList(langs);
             cLanguagesByCountry.putIfAbsent(countryCode, langs);
```
namannigam, line 308: curious, could all of this be replaced with any of the `compute..` variants in `Map`?

Isira-Seneviratne: It should be possible, but I'm not sure how complex the resulting lambda will be.

garydgregory: It's also not "simpler", it's just different, without even thinking about performance consequences....

Isira-Seneviratne: It does make use of parallelStream(), though.

src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -295,14 +297,13 @@ public static boolean isAvailableLocale(final Locale locale) {
...
-            langs = new ArrayList<>();
-            final List<Locale> locales = availableLocaleList();
-            for (final Locale locale : locales) {
-                if (countryCode.equals(locale.getCountry()) &&
-                    locale.getVariant().isEmpty()) {
-                    langs.add(locale);
-                }
-            }
+            // Allows Locales to be processed using multiple threads.
+            langs = availableLocaleList().parallelStream()
```
kinow, line 301: From what I remember (sorry, been doing more Python these days), the `parallelStream` could be good or bad. For a low level library, I think it would be bad.

That method starts executors in the background, and I think it's either a fixed number of executors, or based on the number of cores.

But anyway, the point was that if a library uses `parallelStream`, users of the API would have to be aware that there could be small performance peaks due to these extra executors being created. See [this SO question & answers](https://stackoverflow.com/questions/20375176/should-i-always-use-a-parallel-stream-when-possible).

The place where I would use it, would be some small command line utility processing text/files with thousands or millions of lines in a single computer. There I could perhaps use a parallelStream to do some processing.

So for that, for now, I'd be -1 to `parallelStream` here in common (the code looks nicer by the way :)

PR #525, opened 2020-04-21
Correct Javadocs of methods that use Validate.notNull() and replace some uses of Validate.isTrue() with Validate.notNull().

Isira-Seneviratne:
The Javadocs of some methods that use the `Validate.notNull()` method incorrectly refer to `IllegalArgumentException` instead of `NullPointerException` as the exception thrown when a value is `null`, while the methods' tests check for `NullPointerException`. In addition, some of said methods use `Validate.isTrue()` instead of `Validate.notNull()` for null checking.

This PR fixes both issues as they are closely related.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharUtils.java
+++ b/src/main/java/org/apache/commons/lang3/CharUtils.java
@@ -131,7 +131,7 @@ public class CharUtils {
      *
      * @param ch  the character to convert
      * @return the char value of the Character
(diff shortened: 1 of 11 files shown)


garydgregory:
@Isira-Seneviratne 
May you please rebase on master if still appropriate?

Isira-Seneviratne:
> 
> 
> @Isira-Seneviratne
> May you please rebase on master if still appropriate?

Of course.

PR #528, opened 2020-04-25
Combine two if conditions in ClassUtils.isAssignable().

Isira-Seneviratne:


garydgregory:
-1 I think the code is easier to maintain with one "if" per type.

PR #529, opened 2020-04-29
[LANG-1545] CharSequenceUtils.regionMatches is wrong dealing with Georgian.

XenoAmess:
see details in tests.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -3304,9 +3304,20 @@ public class StringUtilsTest {
 
     @Test
     public void testGeorgianSample() {
-        char[] arrayI = new char[]{(char) 305, (char) 1012};
-        char[] arrayJ = new char[]{(char) 304, (char) 977};
+        char[] arrayI = new char[]{
+                //ı
+                (char) 0x131,
 
+                //ϴ
+                (char) 0x3f4
+        };
+        char[] arrayJ = new char[]{
+                //İ
+                (char) 0x130,
+
+                //ϑ
+                (char) 0x3d1
+        };
         for (char i : arrayI) {
             for (char j : arrayJ) {
                 String si = "" + i;
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -3301,4 +3301,36 @@ public class StringUtilsTest {
             Locale.setDefault(defaultLocales);
         }
     }
+
+    /**
+     //ignore the full test because it is correct but far too slow.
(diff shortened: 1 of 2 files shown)


XenoAmess:
@garydgregory Hi gary

garydgregory:
Hi @XenoAmess ,
Can the test be simplified to test a single character?
Is this related to https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/ ?

XenoAmess:
> Hi @XenoAmess ,
> Can the test be simplified to test a single character?
Actually this test has already been simplified from[0 , Character.MAX_VALUE] to [0,1000]
I'm not sure if it is good to do more simplity work as it might hide some errors.
However if you really want to simplify it, I can try to write codes to generate a list of suspicious chars pairs, and only check them.

> Is this related to https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/ ?
I know nothing about Turkey but it seems like similar type of problem (according to your descriptions in that url.)
I'm not a philologist though, I cannot make it for sure.

garydgregory:
@XenoAmess It would be best to test what we know is wrong instead of some blanket "Something in here goes wrong and we don't know exactly where."

XenoAmess:
> @XenoAmess It would be best to test what we know is wrong instead of some blanket "Something in here goes wrong and we don't know exactly where."

Agreed. Test  tighten now.

XenoAmess:
@garydgregory  Is the test now acceptable? Should I rebase the two commits to 1 commit? or any other thing I shall do?

XenoAmess:
@garydgregory , anything we can do to get this moving?

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -3301,4 +3301,22 @@ public void testToRootUpperCase() {
             Locale.setDefault(defaultLocales);
         }
     }
+
+    @Test
+    public void testGeorgianSample() {
+        char[] arrayI = new char[]{(char) 305, (char) 1012};
```
garydgregory, line 3307: Please use hexadecimal for the values with a comment that shows the Unicode name for each character. Otherwise, there is not way to tell what this is doing without knowing the Unicode values from memory.

XenoAmess: @garydgregory done.

XenoAmess:
> On that page, the U+01D3 is "Latin Capital Letter U with caron" on this GitHub page it looks like "Greek theta symbol". Which one are you testing?
@garydgregory originally I found it with something uncorrect about [Greek Capital Letter Theta] and [Greek Theta Symbol] .
And then I started a fully test of every pair of unicode chars and found the bug also appears when [Latin Small Letter dotless I] and [Latin Capital Letter I with dot above].

PR #530, opened 2020-05-16
LANG-1542: ToStringBuilder.reflectionToString - Wrong JSON format for List of enums

TranNgocKhoa:
I add some code to fix an issue about **ToStringBuilder.reflectionToString - Wrong JSON format when object has a List of Enum**.
Link issue: https://issues.apache.org/jira/browse/LANG-1542

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
@@ -16,16 +16,16 @@
  */
 package org.apache.commons.lang3.builder;
 
+import org.apache.commons.lang3.ClassUtils;
+import org.apache.commons.lang3.ObjectUtils;
+import org.apache.commons.lang3.StringEscapeUtils;
+import org.apache.commons.lang3.StringUtils;
+
 import java.io.Serializable;
 import java.lang.reflect.Array;
 import java.util.Collection;
 import java.util.Map;
 import java.util.WeakHashMap;
-
-import org.apache.commons.lang3.ClassUtils;
-import org.apache.commons.lang3.ObjectUtils;
-import org.apache.commons.lang3.StringEscapeUtils;
-import org.apache.commons.lang3.StringUtils;
 
 /**
  * <p>Controls {@code String} formatting for {@link ToStringBuilder}.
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/lang3/builder/ToStringStyle.java
```diff
@@ -635,6 +635,14 @@ protected void appendDetail(final StringBuffer buffer, final String fieldName, f
      *  {@code toString}, not {@code null}
      */
     protected void appendDetail(final StringBuffer buffer, final String fieldName, final Collection<?> coll) {
+        if (coll != null && !coll.isEmpty()) {
+            coll.stream().findFirst()
+                    .map(Object::getClass)
+                    .filter(Class::isEnum)
```
swarajsaaj, line 641: This will restrict only enums to be appended in a collection, a list of other types (e.g. Collection<Integer> or Collection<String>) will be ignored. This filter should be removed.

garydgregory: We should avoid using streams here IMO, it will likely make performance worse in what can already be slow when reflection is involved.

TranNgocKhoa: @garydgregory @swarajsaaj I added some code following your comments. Please take a look. Thank you!

swarajsaaj: This should fix the case I mentioned earlier. Seems good.

TranNgocKhoa: @swarajsaaj, I added test case for it

kinow:
Will leave it to @garydgregory to do the final review as he had requested changes before :+1:

kinow:
BTW, @TranNgocKhoa if you could rebase it, the build should pass. There were checkstyle issues on master until recently.

XenoAmess:
Hi.
I don't think the changes in ToStringStyleTest.java is necessary...

src/test/java/org/apache/commons/lang3/builder/ToStringStyleTest.java
```diff
@@ -16,10 +16,10 @@
  */
 package org.apache.commons.lang3.builder;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
XenoAmess, line 21: Hi.
I don't think the changes in ToStringStyleTest.java is necessary...

TranNgocKhoa: @XenoAmess fixed, thank you!

garydgregory:
-1: See below, I will provide an alternate implementation.
- The new method makes a copy of the collection as a new array, this can be done without the extra allocation.
- The arguments to assertEquals() are backward: The order is (expected, actual). If you do not follow the proper argument order, the failure messages are REALLY confusing ;-)
- The new tests do not cover edge cases: 
-- null array of enums
-- empty array of enums
-- null list of enums
-- empty list of enums
- New protected method is missing Javadoc since tag.

PR #535, opened 2020-05-31
[LANG-1549] implement TODO in CharSequenceUtils.lastIndexOf : remake it.

XenoAmess:
Hi.
If this change is accectable then I will start CharSequenceUtils.indexOf using same way.
Though, need there be more tests about this function?


XenoAmess:
Hi.
According to `72 Hour Lazy Consensus` rule, if nobody against this pr, I will continue developing codes about this pr in 50 hours later, means the same functin in indexOf.

chtompki:
Merged manually

PR #542, opened 2020-06-01
[LANG-1557] Change a Pattern to a static final field, for not letting it compile each time the function invoked.

XenoAmess:
change a Pattern to a static final field, for not letting it compile each time the function invoked

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8186,6 +8186,8 @@ public class StringUtils {
         return stripEnd(str, stripChars);
     }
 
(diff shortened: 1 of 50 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -8186,6 +8186,8 @@ public static String strip(String str, final String stripChars) {
         return stripEnd(str, stripChars);
     }
 
+    private static final Pattern STRIP_ACCENTS_PATTERN = Pattern.compile("\\p{InCombiningDiacriticalMarks}+"); //$NON-NLS-1$
```
garydgregory, line 8189: All constants live at the top of the file.

XenoAmess: @garydgregory 
done.

PR #551, opened 2020-06-01
[LANG-1550] Optimize ArrayUtils::isArrayIndexValid method.

nstdio:


src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -3110,19 +3110,22 @@ public static int indexOf(final short[] array, final short valueToFind, int star
     }
 
     /**
-     * Returns whether a given array can safely be accessed at the given index.
+     * <p>Returns whether a given array can safely be accessed at the given index.
```
garydgregory, line 3113: Paragraph tags are not needed for the first paragraph of a Javadoc comment. When you do use HTML tags (elsewhere), you should close them.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -3110,19 +3110,22 @@ public static int indexOf(final short[] array, final short valueToFind, int star
...
+     *
+     * <pre>
+     * ArrayUtils.isArrayIndexValid(null, 0)       = false
+     * ArrayUtils.isArrayIndexValid([], 0)         = false
+     * ArrayUtils.isArrayIndexValid(["a"], 0)      = true
+     * </pre>
```
garydgregory, line 3119: Unfortunately, Javadoc created with a closed pre tag after an EOL appear with an extra blank line. so pre tags should be closed at the end of the last line.

XenoAmess: @garydgregory 
I think this is not a big problem, as most of the `</pre>` usage in this repo (especially in StringUtils) do it this way, means not closed at the end of the last line, but close at a new line.
Should we arrange a fix for all of that usages?

XenoAmess: @garydgregory 
an example:
```
    /**
     * <p>Remove the last character from a String.</p>
     *
     * <p>If the String ends in {@code \r\n}, then remove both
     * of them.</p>
     *
     * <pre>
     * StringUtils.chop(null)          = null
     * StringUtils.chop("")            = ""
     * StringUtils.chop("abc \r")      = "abc "
     * StringUtils.chop("abc\n")       = "abc"
     * StringUtils.chop("abc\r\n")     = "abc"
     * StringUtils.chop("abc")         = "ab"
     * StringUtils.chop("abc\nabc")    = "abc\nab"
     * StringUtils.chop("a")           = ""
     * StringUtils.chop("\r")          = ""
     * StringUtils.chop("\n")          = ""
     * StringUtils.chop("\r\n")        = ""
     * </pre>
     *
     * @param str  the String to chop last character from, may be null
     * @return String without last character, {@code null} if null String input
     */
    public static String chop(final String str) {
```

nstdio: Sorry, I don't really get what is the problem with `</pre>`.

XenoAmess: @nstdio 
In short gary think your docs should be 
```
     * ArrayUtils.isArrayIndexValid(null, 0)       = false
     * ArrayUtils.isArrayIndexValid([], 0)         = false
     * ArrayUtils.isArrayIndexValid(["a"], 0)      = true</pre>
```

instead of 

```
     * ArrayUtils.isArrayIndexValid(null, 0)       = false
     * ArrayUtils.isArrayIndexValid([], 0)         = false
     * ArrayUtils.isArrayIndexValid(["a"], 0)      = true
     * </pre>
```

But I think the current docs is OK.
So suggest just wait.

garydgregory: Not worth changing today...

PR #554, opened 2020-06-13
fix checkstyle.

XenoAmess:
There is a trailing space in one class who blocks the checkstyle.
(Which might caused by my pr so I come here to fix it.)
Sorry about that.


XenoAmess:
update: due to track, 
that space is caused by commit
306dbb0cb01f565bf04d8879cad9b413d94d4af5

kinow:
Change in a test file and only for formatting. No need to wait other reviewers, nor no need for a JIRA IMO. LGTM

PR #555, opened 2020-06-14
Add binary compatibility check to Maven default goal

XenoAmess:
by changing defaultGoal in pom.xml


garydgregory:
If you add japicmp from the default goal, then please remove clirr. Clirr is not Java 8 aware anyway.

PR #557, opened 2020-06-18
fix checkstyles.

XenoAmess:
as title.


aherbert:
I have just fixed all these in master. Please rebase and check if anything is missing.

PR #558, opened 2020-06-19
refine test output for FastDateParserTest

XenoAmess:
FastDateParserTest fails on openjdk-ea
Hi.
I updated my local java 15 to newest version and reproduced the error.
This error is caused by a new added locale (in the latest versions of java15) called "ff_LR_#Adlm".
Unfortunately I found no documents about it, and I have no idea about why it happened, as I am not a "ff_LR_#Adlm" user.
(I doubt if there really exist any "ff_LR_#Adlm" user actually)
And I have a feeling that this is not a bug by commons-lang, but a bug by jdk15, as we are using SimpleDateFormat to format a Date to a String, and use SimpleDateFormat to parse that String, and caused an error. SimpleDateFormat is a class from JDK.

So what I can do is just refine output for tests, for showing what locale caused what error, and wait for oracle guys to fix it.
So, should we do something to inform openjdk guys about this buggy thing we found? And how?
@garydgregory


XenoAmess:
see the new added java15BuggyLocaleTest.
It errors in line 733, means `sdf.parse(formattedDate.toUpperCase(buggyLocale));`
Seems this locale in jdk-15 cannot process a UpperCase date String.

chtompki:
@XenoAmess - regarding informing the OpenJDK guys, I would think we would contact Rory O'Donnell as a response to: https://lists.apache.org/thread.html/rd619f4bc54e6d85cd84d7b9ac5628000c5d787543fe8b01c21c1c9c5%40%3Cdev.commons.apache.org%3E

I'm happy to do that if you want.

chtompki:
Done

PR #561, opened 2020-06-27
[LANG-1543] Implemented feature of writing Maps into JSON

rwanderc:
Implemented the feature of navigating through Maps to write in JSON format.
Looking forward to your feedback.
The issue is described at https://issues.apache.org/jira/browse/LANG-1543


garydgregory:
I applied a different version of the PR. Thank you for your help.

PR #564, opened 2020-06-28
add passing varargs Supplier to ObjectUtils.getIfNull instead of Supplier single param

ytanikin:


ytanikin:
checks failed because of incompatibility, that is actually no true, after changing single param to varargs it is compatible
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.14.3:cmp (default-cli) on project commons-lang3: There is at least one incompatibility: org.apache.commons.lang3.ObjectUtils.getIfNull(java.lang.Object,java.util.function.Supplier):METHOD_REMOVED -> [Help 1]

XenoAmess:
> checks failed because of incompatibility, that is actually no true
No it is.
you can reproduce by:
0. compile original commons-lang to original-commons-lang.jar
1. create another repo who use original-commons-lang.jar as a lib, named A.
2. compile A into a jar.
3. make sure the A.jar can run correctly with original-commons-lang.jar
4. apply your changes to the codes in commons-lang.
5. compile commons-lang to modified-commons-lang.jar
6. run A.jar with modified-commons-lang.jar
I strongly guess it will fail.
Varargs is not that convienent, my friend.
If you really want a varargs version of that function you can create one, but don't delete the original function.

garydgregory:
What is your use-case for this? It seems rather obtuse to me. Perhaps a real example would help.

ytanikin:
`ObjectUtils.getIfNull(
                number,
                () -> Support.getMasterNumber(file),
                () -> Support.getPredefinedNumber(file),
                this::takeFreeNumber,
                Support::getUndefinedNumber
        );`
it is a lazy calculation, don't need to calculate all possible values and use ObjectUtils.firstNonNull

PR #568, opened 2020-06-29
[LANG-1579] refine StringUtils.convertRemainingAccentCharacters

XenoAmess:
as title.
reduce array copy, for two times.


XenoAmess:
Hi.

PR #569, opened 2020-06-29
[LANG-1580] refine StringUtils.deleteWhitespace

XenoAmess:
don't create more empty String.


XenoAmess:
Hi.

PR #570, opened 2020-06-29
[LANG-1581] refine StringUtils.getFuzzyDistance

XenoAmess:
use char array insteadof String for multi-layer loop.


XenoAmess:
jmh test shows this pr is meaningless.

PR #573, opened 2020-06-29
[LANG-1584] Refine StringUtils.isNumericSpace

XenoAmess:
reduce calling charAt several times.
also change order to try to reduce function calls.


XenoAmess:
Hi.

PR #579, opened 2020-07-07
Speed up Travis build

oehme:
I saw that the Travis builds are taking over 10min, much more than what I saw locally. I made some modifications to the Travis file to improve that:

- I've activated caching of `.m2/repository`, so the dependencies don't have to be redownloaded every time
- I've removed the unnecessary `mvn install -DskipTest` which Travis does by default
- I've combined the two build steps into one - The only behavioral change here is that there is no longer a test without jacoco instrumentation. If you're concerned about that and are okay with spending double the time on testing (once with and once without jacoco), I can revert that commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ matrix:
   allow_failures:
     - jdk: openjdk-ea
 
+install: skip
+
 script:
   - mvn
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@
 # limitations under the License.
 
 language: java
+
+cache:
+  directories:
+    - $HOME/.m2/repository
 jdk:
   - openjdk8
   - openjdk11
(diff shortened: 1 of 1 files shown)


oehme:
Wow, 16min and the build isn't even queued yet. Maybe instead of fixing the Travis build, a migration to Github actions would be better. Those start immediately from my experience.

XenoAmess:
> Wow, an hour later and the build isn't even queued yet. Maybe instead of fixing the Travis build, a migration to Github actions would be better. Those start immediately from my experience.
> 
> Edit: Just saw you're already using Github actions too - Any reason not to switch over completely?

travis-ci have more operations to use.
github-ci, on the other hand, is really fast.(without pending and pending)
But in my experience github-ci broke more frequently than travis-ci.

oehme:
Closing since there seems to be no interest.

PR #580, opened 2020-07-07
Add StringUtils.startsWithIgnoreCase

LinuxClient:
As the name of the method says: This method is the same as String.startsWith except that it is case insensitive.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -869,7 +869,7 @@ public class StringUtils {
      * @since 3.5
      */
     public static int compare(final String str1, final String str2, final boolean nullIsLess) {
-        if (str1 == str2) {
+        if (str1.equals(str2)) {
             return 0;
         }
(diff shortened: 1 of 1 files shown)


garydgregory:
I am not a fan of adding more and more "ignore case" versions of APIs.

For 4.0, we could consider a new StringUtils class that has two default instances, one that is case-sensitive, the other not, which you would use like:
```
StringUtils.DEFAULT.theApiYouWant(...); // case-sensitive
StringUtils.NAME_CASE_SENSITIVE.theApiYouWant(...);
StringUtils.NAME_CASE_INSENSITIVE.theApiYouWant(...);
```
The alternative for now might be to override the existing API with a boolean or maybe an enum that says what kind of case-sensitivity you want.

Any thoughts from the community?

LinuxClient:
Yes, that's a very good suggestion. _I wonder why I wrote equalsIgnoreCase several times instead of ignoreCase._

LinuxClient:
I have overlooked the fact that such a method already exists. Therefore I close this pull request.

PR #582, opened 2020-07-14
Remove redundant parameter: str.length()

bhawna94:
Remove redundant parameter:str.length() as by default substring(int beginIndex) method returns the substring from the specified index and extends till string length.


bhawna94:
@kinow Could you please review the PR and let me know your thoughts.

XenoAmess:
Hi.
I agree with this pr, but I think you can do it more aggressively.
You see, in this function, param `str` is final, thus `str.length()` is final.
Why don't we let a `final int length = str.length()` to avoid calling function `str.length()` sevaral times?
Any ideas guys?

garydgregory:
Seems reasonable to only call length() once if str is final.

bhawna94:
Agreed! @XenoAmess @garydgregory

bhawna94:
Updated the PR. Please have a look

PR #583, opened 2020-07-18
[LANG-1544] Fixes null pointer exception in distance method in MethodUtils

pixelr:
This Pull request fixed Null Pointer Exception in the 
MethodUtils:774 :  NPE when classArray[offset] is null
[https://issues.apache.org/jira/browse/LANG-1544](https://issues.apache.org/jira/browse/LANG-1544)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
@@ -782,7 +782,7 @@ public class MethodUtils {
         }
         for (int offset = 0; offset < classArray.length; offset++) {
             // Note InheritanceUtils.distance() uses different scoring system.
-            if (classArray[offset].equals(toClassArray[offset])) {
+            if (null == classArray[offset] || classArray[offset].equals(toClassArray[offset])) {
                 continue;
             } else if (ClassUtils.isAssignable(classArray[offset], toClassArray[offset], true)
                     && !ClassUtils.isAssignable(classArray[offset], toClassArray[offset], false)) {
--- a/src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
@@ -1015,6 +1015,8 @@ public class MethodUtilsTest {
         assertEquals(0, distanceMethod.invoke(null, new Class[]{Date.class}, new Class[]{Date.class}));
         assertEquals(1, distanceMethod.invoke(null, new Class[]{Integer.class}, new Class[]{ClassUtils.wrapperToPrimitive(Integer.class)}));
         assertEquals(2, distanceMethod.invoke(null, new Class[]{Integer.class}, new Class[]{Object.class}));
+        assertEquals(-1, distanceMethod.invoke(null, new Class[]{Integer.class, null}, new Class[]{Object.class}));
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -782,7 +782,7 @@ private static int distance(final Class<?>[] classArray, final Class<?>[] toClas
         }
         for (int offset = 0; offset < classArray.length; offset++) {
             // Note InheritanceUtils.distance() uses different scoring system.
-            if (classArray[offset].equals(toClassArray[offset])) {
+            if (Objects.isNull(classArray[offset]) || classArray[offset].equals(toClassArray[offset])) {
```
garydgregory, line 785: Objects#isNull is meant to be used with lambdas, use a `==` test here instead.

pixelr: I have update the commit with suggested changes. Please have a look. thanks !

pixelr: @garydgregory Please review and merge if changes are correct

garydgregory:
Hi @pixelr 

After looking again, I'm not sure this PR makes sense. The test you patched calls a private method, and this private method is not going to be called with just any value. In this case, the private method `distance` is only called from `MethodUtils.getMatchingMethod(Class<?>, String, Class<?>...)` where our `getMatchingMethod` calls `java.lang.reflect.Method.getParameterTypes()`. 

Do you have a case where `java.lang.reflect.Method.getParameterTypes()`returns an array containing a null? Can you show this by calling a public method like `org.apache.commons.lang3.reflect.MethodUtils.getMatchingMethod(Class<?>, String, Class<?>...)`?

Otherwise, this PR just make the `distance` method more complicated than it has to be, granted not by much, but there would then be no sense if making the whole code base more NPE proof where it does not need to be.

Let us know your findings.

XenoAmess (commented):
Is there example user-side codes which can lead to NPE before this pr?

pixelr:
> Is there example user-side codes which can lead to NPE before this pr?

@XenoAmess I took this a first good issue to fix. But, when I tried reproducing it, I am not able to do so. I have asked the reporter on JIRA for info on this. If we don't find a real use case, will close the PR

pixelr:
> Hi @pixelr
> 
> After looking again, I'm not sure this PR makes sense. The test you patched calls a private method, and this private method is not going to be called with just any value. In this case, the private method `distance` is only called from `MethodUtils.getMatchingMethod(Class<?>, String, Class<?>...)` where our `getMatchingMethod` calls `java.lang.reflect.Method.getParameterTypes()`.
> 
> Do you have a case where `java.lang.reflect.Method.getParameterTypes()`returns an array containing a null? Can you show this by calling a public method like `org.apache.commons.lang3.reflect.MethodUtils.getMatchingMethod(Class<?>, String, Class<?>...)`?
> 
> Otherwise, this PR just make the `distance` method more complicated than it has to be, granted not by much, but there would then be no sense if making the whole code base more NPE proof where it does not need to be.
> 
> Let us know your findings.
you are right. I don't think complicating code would be right. I have asked bug reporter for further info else will close the PR.

pixelr:
closing. Please re-open if need. Thanks for reviewing :)

PR #584, opened 2020-07-19
Removing uncessary unboxing at several places

mehwhatever:


garydgregory:
-1 In a low level level component like this one, it is important IMO to explicitly see and be aware of where boxing and unboxing takes place since this could impact performance.

PR #585, opened 2020-07-20
Remove redudant argument from substring call

bhawna94:
Remove redundant parameter:str.length() as by default substring(int beginIndex) method returns the substring from the specified index and extends till string length. 
And as str.length() is final.So, storing str.length() into a variable final int length = str.length() to avoid calling function str.length() sevaral times.


bhawna94:
@britter @garydgregory @chtompki @kinow Please review the PR and let me know your thoughts as well.

bhawna94:
Thanks @kinow . Are we good to merge this or waiting for other approvals

kinow:
Not a problem @bhawna94 , I think it should be ready to be merged, I'm just looking at other PR's at the moment, but can probably merge in some moments :+1:

bhawna94:
Thanks for the update @kinow

kinow:
Merged. Thanks @bhawna94 !

PR #595, opened 2020-07-31
Add method to remove all numbers from a String 

muriloalvesdev:
### Motivation:

I came across a problem in which I had to remove all the numbers contained in a String. After researching in some places I saw that in the stackoverflow there are many questions about this same problem. I use the StringUtils class a lot and I haven't found any method capable of solving this problem, so I decided to create this Pull Request to contribute to the project.

### Solution:
Create a method capable of receiving any String and returning it without any numeric characters.

- [x] This PR contains tests of the method created.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -180,6 +180,11 @@ public class StringUtils {
     private static final int PAD_LIMIT = 8192;
 
     /**
+     * <p>Finds numbers 0 through 9. </p>
+     */
+    private static final String REGEX_FIND_NUMBERS = "[0-9]";
+    
+    /**
      * Pattern used in {@link #stripAccents(String)}.
      */
     private static final Pattern STRIP_ACCENTS_PATTERN = Pattern.compile("\\p{InCombiningDiacriticalMarks}+"); //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -221,6 +226,27 @@ public static String abbreviate(final String str, final int maxWidth) {
...
+     * StringUtils.removeAllNumbers("abc123")    = "abc"
+     * StringUtils.removeAllNumbers("123abc123") = "abc"
+     * StringUtils.removeAllNumbers("a1b2c3")    = "abc"
+     * </pre>
+     * 
+     * @param str String to check, may be null
+     * @return The String without the numbers
+     */
+    public static String removeAllNumbers(String str) {
+      return !isBlank(str) ? str.replaceAll(REGEX_FIND_NUMBERS, EMPTY) : EMPTY;
```
XenoAmess, line 247: 1. if isBlank(str) should not return EMPTY, but return str itself.
2. maybe using a hand-write loop will be faster than using regex. please do a benchmark.

muriloalvesdev: Hello XenoAmess!

My English is unfortunately still intermediate, did you suggest that I do this validation manually?

Was it something like that?

```
public static String removeAllNumbers(String str) {
        String result = "";
        if(!isEmpty(str)) {
          char[] array = str.toCharArray();
          for(int i = 0 ; i < array.length; i++) {
              if(!Character.isDigit(array[i])) {
                result += array[i];
              }
          }
          return result;
        }
        return str;
}
```

muriloalvesdev: XenoAmess I am creating a new commit with this change. I also corrected the checkstyle. Could you review again?

XenoAmess:
Hi.
you can try something like this
```
    public static String removeAllNumbers(String str) {
        final int originalLength = length(str);
        if (originalLength == 0) {
            return str;
        }
        char[] charArray = str.toCharArray();
        int i = originalLength - 1;
        int j = i;
        while (i >= 0) {
            if (charArray[i] >= '0' && charArray[i] <= '9') {
                --i;
                continue;
            }
            charArray[j] = charArray[i];
            --i;
            --j;
        }
        return new String(charArray, j + 1, originalLength - 1 - j);
    }
```

XenoAmess:
however I don't know why you want to change " " to "" using this function, by its name I think we should only try to remove the numbers, and " " does not sound like a number...

XenoAmess:
Besides, you need to write some test cases for the new added function.
for example, org.apache.commons.lang3.StringUtilsContainsTest is test for StringUtils.contains functions.

XenoAmess:
And finally (or maybe should do it first actually), go jira system (https://issues.apache.org/jira) and open a issue ticket for the new function, and wait for what the committers say. they are people who can decide whether we can merge this function/

src/main/java/org/apache/commons/lang3/StringUtils.java
XenoAmess: @muriloalvesdev Hi.
I re-reviewed this pr and have no more suggestions about the implementation.
Please open a JIRA ticket for this pr, and add the string of that ticket in front of the pr's name 
(example: https://github.com/apache/commons-lang/pull/524)

Next step is waiting for some people who have merge access(means, Committer) to review this pr, and see if we should merge this.

yoshi44:
Unsubscribe please

garydgregory:
-1

Use the stock Java classes: `"12ab".replaceAll("[\\d]", "")` return `"ab"`

In general I think we need to move this kind of text processing to the Commons Text component instead of adding more and more to Commons Lang, especially when it comes to manipulating Strings that are not generic, in this case, the domain is numbers.

PR #602, opened 2020-08-10
Made String Reverse Method Faster

nishantc1527:
Right now, StringUtils.reverse() is done the lazy way by initializing a new Stringbuilder, calling reverse, then converting back into a string. I filled out the reverse method by doing it manually.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -7101,7 +7101,16 @@ public class StringUtils {
         if (str == null) {
             return null;
         }
-        return new StringBuilder(str).reverse().toString();
+
+        char[] chars = str.toCharArray();
+
+        for (int i = 0, j = chars.length - 1; i < j; i++, j--) {
+            char temp = chars[i];
+            chars[i] = chars[j];
+            chars[j] = temp;
+        }
+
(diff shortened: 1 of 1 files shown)


XenoAmess:
Hi.
It actually breaks BC...
the original code can handle surrogate pairs.
you can see details in StringBuilder's source.

XenoAmess:
besides, you should run a jmh test to show the performance incresement.
see HashSetvBitSetTest for example.

nishantc1527:
> Hi.
> It actually breaks BC...
> the original code can handle surrogate pairs.
> you can see details in StringBuilder's source codes.

Sorry, I don't know what surrogate pairs are... do you mind explaining?

> besides, you should run a jmh test to show the performance incresement.
> see HashSetvBitSetTest for example.

Doing that now!

garydgregory:
@XenoAmess 
Let's be careful with the terms we use here: this PR does NOT break BC. BC is about API signatures, not behavior.

nishantc1527:
Never mind, I think ```StringBuilder```'s reverse method is better. Sorry for wasting anyone's time.

XenoAmess:
> @XenoAmess
> Let's be careful with the terms we use here: this PR does NOT break BC. BC is about API signatures, not behavior.

@garydgregory Oh yes. thanks for the clarification.

PR #603, opened 2020-08-11
[LANG-1595] Add ArrayUtils methods that accept a Predicate.

Isira-Seneviratne:


garydgregory:
Closing: No feedback and it also feels like a duplication of applying predicates to a `Stream.of(...)`.

PR #608, opened 2020-08-15
Takari maven wrapper

nhojpatrick:
Maven 3.7.0 will introduce a maven wrapper that is based upon the takari maven wrapper.

The advantage of using a maven wrapper is that it puts the control over what maven version to be use into source control, and each branch or tag controls the version it requires. Both developers and CI/CD no longer have to play russian roulette over what version needs to be installed.

This PR changes mvn CI/CD commands to mvnw include documentation and comments.

I've been using takari for about 4 years and have tested the version which will released as part of maven v3.7.0. Moving to the apache maven version will simply be a case of properties file url which happens for any maven upgrade and the mvnw scripts mainly around the license header.


nhojpatrick:
Withdrawing PR due to negative email voting regarding mvnw as being an improvement.

PR #615, opened 2020-09-01
[LANG-1606] StringUtils.countMatches returns incorrect value while handling intersecting substrings

GaliFFun:


sebbASF:
Thanks for the PR, but it is the Javadoc that must be fixed, not the code.
If necessary, raise a separate issue for the a version of the method that supports counting overlapping occurrences.

PR #629, opened 2020-10-17
[LANG-1608] Include junit-bom for dependency version alignment.

nstdio:
Alternatively, we can rollback to `org.junit.jupiter:junit-jupiter:5.6.2`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -513,13 +513,24 @@
     </contributor>
   </contributors>
 
+  <dependencyManagement>
+    <dependencies>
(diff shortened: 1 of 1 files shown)


aherbert:
This looks OK to me. There is an intermittent failure with some DateTime parsing tests on Java 15 that appears in the history of GH builds hence the failures. This change fixes the ability to run the tests as intended.

Can you force push to this branch to change the commit message to use `LANG-1608` then it is good to merge.

PR #633, opened 2020-10-25
Close/flush the `OutputStream` before calling `toByteArray()` on underlying `ByteArrayOutputStream`

HubertWo:
This PR is related to: https://github.com/apache/commons-lang/pull/400 

I have pulled most recent changes from repository and used ```try-with-resources``` to correctly close resources.


garydgregory:
-1 Implemented differently, see git master.

PR #634, opened 2020-10-25
NumberUtilsTest - incorrect types in min/max tests

HubertWo:
Some of the tests related to finding min and max values does not have types declared. As a result min/max of value ```int``` methods are being tested multiple times without invoking target methods ```byte```, ```long``` etc.
I fixed types and added casting where required.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
@@ -696,11 +696,11 @@ public class NumberUtilsTest {
 
     @Test
     public void testMinLong() {
-        assertEquals(5, NumberUtils.min(5), "min(long[]) failed for array length 1");
-        assertEquals(6, NumberUtils.min(6, 9), "min(long[]) failed for array length 2");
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
```diff
@@ -16,20 +16,15 @@
...
+import org.junit.jupiter.api.Test;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.math.RoundingMode;
 
-import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.*;
```
garydgregory: Please don't use star imports.

garydgregory (changes requested):

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
HubertWo: Fixed.

PR #636, opened 2020-10-28
LANG-1610: Fix StringUtils.unwrap throws StringIndexOutOfBoundsException

liangshuche:
When wrapToken.length() * 2 is larger than str.length(), and str starts with and ends with wrapToken, the function calls str.substring(beginIndex, endIndex) with beginIndex larger than endIndex, which throws a StringIndexOutOfBoundsException.

It should return the original string since it is not quoted properly with the wrapToken according to documentation.


namannigam (approved):
This change looks good.

garydgregory:
Thank you @liangshuche ! :-)

PR #638, opened 2020-11-02
fix javadoc for isJavaVersionAtMost

JohnRDOrazio:
the javadoc for isJavaVersionAtMost was the same as for isJavaVersionAtLeast! They sure don't do the same thing :)


namannigam (approved):
Makes sense.

garydgregory:
Thank you @JohnRDOrazio !

PR #640, opened 2020-11-08
[LANG-1612] testGetAllFields and testGetFieldsWithAnnotation sometimes fail

tongxin97:
Issue link: https://issues.apache.org/jira/browse/LANG-1612

Don't use `assertArrayEquals()` because it would fail when the array returned by `FieldUtils.getAllFields(...)` is not ordered in the way the test assumes in `annotatedFields` and `ArrayUtils.addAll(fieldsInteger, fieldsNumber)`. 

Instead, do order-independent assertions on: 1) the size of the returned array, 2) the element it contains.


garydgregory:
Hi @tongxin97 
Thank you for your PR.
There is a simpler solution: sorting, see git master

tongxin97:
Ah I see. I didn't do it this way since `Field` is not comparable. I saw your workaround with the `sort()` method. Thanks for keeping me updated.

PR #630, opened 2020-10-17
[LANG-1609] Case-insensitive variants for startsWithAny and endsWithAny.

nstdio:

PR #641, opened 2020-11-09
[LANG-1609] - Case-insensitive variants for startsWithAny and endsWithAny

arturobernalg:
Add methods to StringUtils class
StringUtils::startsWithAnyIgnoreCase(CharSequence, CharSequence...)
StringUtils::endsWithAnyIgnoreCase(CharSequence, CharSequence...)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8110,6 +8110,73 @@ public class StringUtils {
         return false;
     }
 
+
+    /**
+     * <p>Check if a CharSequence starts with any of the ignored case-sensitive prefixes.</p>
+     *
+     * <pre>
+     * StringUtils.startsWithAnyIgnoreCase(null, null)                                  = false
+     * StringUtils.startsWithAnyIgnoreCase(null, new String[] {"abc"})                  = false
+     * StringUtils.startsWithAnyIgnoreCase", null)                                      = false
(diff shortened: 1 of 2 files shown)


arturobernalg:
Hi @garydgregory 
ping here ;)

XenoAmess:
@garydgregory @arturobernalg I remembered gary had a plan to due with all ignoreCase functions differently, in lang4. 
Also, should this function be added directly to commons-text instead of commons-lang?

garydgregory:
> @garydgregory @arturobernalg I remembered gary had a plan to due with all ignoreCase functions differently, in lang4.
> Also, should this function be added directly to commons-text instead of commons-lang?

Good reminder @XenoAmess 
I want us to think about what belongs in Lang vs. Text. IMO when we talk about "case", we talk about Text. As opposed to a String which is a wrapper for an array. So I think I'd like to see case operations in Text. Needs more study... I certainly don't want to keep growing StringUtils beyond it's current kitchen sink style.

arturobernalg:
Hi @garydgregory  @XenoAmess 
I really didn't know that. How can I know those things?
So, and now? What do I do with the PR?do i need close the PR?

garydgregory:
In general, information is spread about on the dev mailing list, PR comments, and JIRA tickets. The dev ML has the most history information I would say but JIRAs and PRs are also important. Remember that this is a volunteer based organization and that we do our best :-)

nstdio:
Is this same #630 ?

PR #642, opened 2020-11-09
[LANG-1613] - BigDecimal is created when you pass it the min and max values

arturobernalg:
… of Double


garydgregory:
Fixed while keeping explicit unboxing and more tests compared to the PR.

PR #643, opened 2020-11-09
[LANG-1420] - TypeUtils.isAssignable returns wrong result for GenericAr…

arturobernalg:
…rayType and ParameterizedType comparison

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java
@@ -424,6 +424,11 @@ public class TypeUtils {
             return true;
         }
 
+        //the component type should not be assignable to the array type
+        if (type instanceof GenericArrayType) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java
```diff
@@ -424,6 +424,11 @@ private static boolean isAssignable(final Type type, final ParameterizedType toP
             return true;
         }
 
+        //the component type should not be assignable to the array type
```
garydgregory, line 427: This comment is backward.

I've committed a variation of the PR to git master with more tests.

I am wondering about other Type extensions. What do you think about `ypeUtilsTest.testIsAssignableGenericArrayTypeToWildercardType()`? See `isAssignable(Type, Type, Map<TypeVariable<?>, Type>)`.

garydgregory:
Closing, see https://github.com/apache/commons-lang/pull/643#discussion_r523783371

PR #644, opened 2020-11-09
[LANG-1231] - StringUtils#indexOfAny() methods with start position argu…

arturobernalg:
…ment

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -2855,6 +2855,39 @@ public class StringUtils {
     }
 
     /**
+     * <p>Search a CharSequence to find the first index of any
+     * character in the given set of characters starting with
+     * a given index.</p>
+     *
+     * <p>A {@code null} String will return {@code -1}.
+     * A {@code null} search string will return {@code -1}.</p>
+     *
+     * <pre>
+     * StringUtils.indexOfAny(null, 0, *)            = -1
+     * StringUtils.indexOfAny("", 0, *)              = -1
+     * StringUtils.indexOfAny(*, 0, null)            = -1
+     * StringUtils.indexOfAny(*, 0, "")              = -1
+     * StringUtils.indexOfAny("zzabyycdxx", 0, "za") = 0
+     * StringUtils.indexOfAny("zzabyycdxx", 0, "by") = 3
+     * StringUtils.indexOfAny("aba", 0, "z")         = -1
+     * StringUtils.indexOfAny("aba", 1, "a")         = 1
+     * StringUtils.indexOfAny("aba", -1, "a")        = -1
+     * </pre>
+     *
+     * @param cs  the CharSequence to check, may be null
+     * @param beginIndex  the start position to search
+     * @param searchChars  the chars to search for, may be null
+     * @return the index of any of the chars, -1 if no match,null input or  the beginIndex smaller than 0
+     * @since 3.2
+     */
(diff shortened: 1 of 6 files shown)


kinow (commented):
Looks good to me. Might have to review with calm and the IDE to test for surrogate bytes, and see if there's any case that we didn't think of, or haven't tested.

But looks really good @arturobernalg , thanks for the PR!

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2854,6 +2854,39 @@ public static int indexOfAny(final CharSequence cs, final String searchChars) {
...
+     * StringUtils.indexOfAny("aba", 0, "z")         = -1
+     * StringUtils.indexOfAny("aba", 1, "a")         = 1
+     * StringUtils.indexOfAny("aba", -1, "a")        = -1
+     * </pre>
+     *
+     * @param cs  the CharSequence to check, may be null
+     * @param beginIndex  the start position to search
+     * @param searchChars  the chars to search for, may be null
+     * @return the index of any of the chars, -1 if no match,null input or  the beginIndex smaller than 0
+     * @since 3.2
```
garydgregory, line 2881: Wrong since tag.

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/StringUtils.java
arturobernalg: Hi @garydgregory Which would be correct?

arturobernalg: True. Changed to 3.12

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2854,6 +2854,39 @@ public static int indexOfAny(final CharSequence cs, final String searchChars) {
...
+     * @param beginIndex  the start position to search
+     * @param searchChars  the chars to search for, may be null
+     * @return the index of any of the chars, -1 if no match,null input or  the beginIndex smaller than 0
+     * @since 3.12
+     */
+    public static int indexOfAny(final CharSequence cs, int beginIndex, final String searchChars) {
+        if (isEmpty(cs) || isEmpty(searchChars) || beginIndex < 0) {
+            return INDEX_NOT_FOUND;
+        }
+        return indexOfAny(cs.toString().substring(beginIndex), searchChars.toCharArray());
```
kinow, line 2887: Looks good I think. Only thing I'd like to confirm is what happens with those surrogate pairs. (Can't recall if that affects the length of the string, in which case it could give unexpected results here?)

https://garygregory.wordpress.com/2015/11/03/java-lowercase-conversion-turkey/

kinow: Maybe we should just document that it's not supported for now. Or spend some more time making it support special locales/chars?

```java
Locale turkish = Locale.forLanguageTag("tr");
assertEquals(1, StringUtils.indexOfAny("TITLE".toLowerCase(turkish), 1, "i"));
```

WDYT @garydgregory ?

@arturobernalg if you have any thoughts too, but in the meantime, could you squash your commits? Your change is very well written, and concise. But there are now 13 commits due to review feedback. We can squash it too when merging, but it'd be simpler if you could do it, please.

Thanks!

kinow: Although I think other similar methods make no remarks about locales/surrogate pairs... so we can probably ignore what I just said I guess ☺

PR #645, opened 2020-11-10
[LANG-1614] - Split fixed-length elements

arturobernalg:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7731,6 +7731,48 @@ public static String rotate(final String str, final int shift) {
...
+     * StringUtils.splitEvery("",10)                  = [""]
+     * StringUtils.splitEvery(null,4)                 = {@code null}
+     * StringUtils.splitEvery("testAbgTestsABG",2)    = ["te","st","Ab","gT","es","ts","AB","G"]
+     * </pre>
+     *
+     * @param str {@link String} to split into fixed length elements.
+     * @param length length of the elements into which the String will be divided.
+     * @return a {@code String []} of fixed-length elements.
+     */
+    public static String[] splitEvery( final String str, final int length ) {
```
aakashgupta96, line 7750: I think it'll be better to throw `IllegalArgumentException` in case `length <1  || length > str.length()`. @arturobernalg  Thoughts?

arturobernalg: Hi 

usually in the others method return null . An. example of that -->

`  public static String abbreviateMiddle(final String str, final String middle, final int length) {
        if (isAnyEmpty(str, middle) || length >= str.length() || length < middle.length()+2) {
            return str;
        }`

PR #646, opened 2020-11-13
[LANG-1345] - Add methods to enhance non-empty strings

arturobernalg:


arturobernalg:
I really like this method and the proposal. I think it is quite useful and should be included in the next version. I create the PR with the name change proposal.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7274,6 +7274,52 @@ public static String rightPad(final String str, final int size, String padStr) {
...
+     * StringUtils.defaultIfNotBlank("bat", "pre-", "-post") = "pre-bat-bost"
+     * StringUtils.defaultIfNotBlank("bat", null, "-post")      = "bat-post"
+     * StringUtils.defaultIfNotBlank("bat", "pre-", null)      = "pre-bat"
+     * </pre>
+     * @param str the String to check, may be null
+     * @param prefix  the string to prepend if not blank. Null will be converted to empty string.
+     * @param suffix  the string to append if not blank. Null will be converted to empty string.
+     * @return the passed in String with prefix and suffix added, or empty string
+     * @see StringUtils#defaultString(String, String)
+     */
```
garydgregory, line 7296: And since tag.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -7274,6 +7274,52 @@ public static String rightPad(final String str, final int size, String padStr) {
...
+    public static String defaultIfNotBlank(final String str, final String prefix, final String suffix) {
+        return isBlank(str) ? "" : defaultString(prefix) + str + defaultString(suffix);
```
garydgregory, line 7298: Use EMPTY constant.

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/StringUtils.java
arturobernalg: Done

src/main/java/org/apache/commons/lang3/StringUtils.java
arturobernalg: Add * @since 3.12

src/main/java/org/apache/commons/lang3/StringUtils.java
arturobernalg: HI @garydgregory Should i make any other changes?

PR #647, opened 2020-11-13
[LANG-1541] - ArrayUtils.contains() fails to handle Double.NaN

arturobernalg:


garydgregory:
Hm, now that I look at the code, the change is in a deeper method that returns an index but you are not testing the method you changed, and this is likely why you did not find the bug in your implementation ;-)
Please see git master for the proper fix and test. I you want to contribute a Jira and PR for the Float version, that would be most welcome :-)

arturobernalg:
Hi @garydgregory 

I really appreciate the comments.  

i've created the Jira and the PR with the fix.

[LANG-1615](https://issues.apache.org/jira/browse/LANG-1615)

PR #654, opened 2020-11-20
[LANG-1533] - RandomStringUtils.random() should not include \u0000 or control characters

arturobernalg:
Add Character.CONTROL


arturobernalg:
HI @garydgregory 
 PR created for --> https://issues.apache.org/jira/browse/LANG-1533

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -404,6 +404,7 @@ public static String random(int count, int start, int end, final boolean letters
                 case Character.UNASSIGNED:
                 case Character.PRIVATE_USE:
                 case Character.SURROGATE:
+                case Character.CONTROL:
```
garydgregory, line 407: That does not look right IMO because it excludes the TAB character.

PR #657, opened 2020-11-21
[LANG-1524] : Added cycle detection check in toString for classes

aakashgupta96:
1. To String method earlier used to give stack overflow error in case of cyclic reference.
2. Added a cycle detection check before calling toString recursively on type parameters.
https://issues.apache.org/jira/browse/LANG-1524

### DIFF
--- a/src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
@@ -21,6 +21,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.awt.Insets;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


aakashgupta96:
@arturobernalg Could you please review this PR? 
Since I'm new to open source, let me know in case I missed something or I need to make any changes.

arturobernalg:
rebase from master to solve the conflict

aakashgupta96:
@arturobernalg Done rebasing changes with master. Can you check now?

aakashgupta96:
@arturobernalg @garydgregory Done rebasing changes with master. Can you please check now?

garydgregory:
-1: A better and proper fix would actually return a String. If can the test code compiles, then the API call should return a "Java-esque" String per the Javadoc.

aakashgupta96:
@garydgregory 
According to me, we cannot represent the cyclic dependency in extended string form, hence I thought it would be better to detect this kind of scenario and then throw an exception in order to avoid cyclic method calls and StackOverflow Exception. 

Can you please suggest how the 'java-esque' string would look like?

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
```diff
@@ -949,4 +957,11 @@ public void testWrap() {
 
         assertEquals(String.class, TypeUtils.wrap(String.class).getType());
     }
+
+    @Test
+    public void testLand1524() {
```
garydgregory, line 962: typo -> "Land" -> "Lang"

garydgregory:
> @garydgregory
> According to me, we cannot represent the cyclic dependency in extended string form, hence I thought it would be better to detect this kind of scenario and then throw an exception in order to avoid cyclic method calls and StackOverflow Exception.
> 
> Can you please suggest how the 'java-esque' string would look like?

In the Eclipse debugger, `AAAAClass.BBBBClass.CCCClass.class` is displayed as "class org.apache.commons.lang3.reflect.AAAAClass$BBBBClass$CCCClass". We don't display the `class` prefix IIRC. I do not know if Eclipse does not display generics because it never does or because doing so in this case is problematic.

aakashgupta96:
@garydgregory I agree that the problem is coming only because TypeUtils Class is trying to display the information of type parameters too. 
Java as language allows having type parameters on inner nested classes too, but we surely can't display type information in these scenarios where we have circular reference.

Both IntelliJ and Eclipse don't show type information for the classes and hence they never go in that cyclic loop which java as language supports.

PR #658, opened 2020-11-21
[LANG-1616] - Add lastIndexOfAnyChar

arturobernalg:


arturobernalg:
HI @garydgregory 

any thoughts about this implementation?

PR #659, opened 2020-11-21
LANG-1617 - Improve/clean math package

arturobernalg:


src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -32,43 +32,43 @@
...
+    public static final Byte BYTE_ONE = (byte) 1;
     /** Reusable Byte constant for minus one. */
-    public static final Byte BYTE_MINUS_ONE = Byte.valueOf((byte) -1);
+    public static final Byte BYTE_MINUS_ONE = (byte) -1;
     /** Reusable Double constant for zero. */
-    public static final Double DOUBLE_ZERO = Double.valueOf(0.0d);
+    public static final Double DOUBLE_ZERO = 0.0d;
     /** Reusable Double constant for one. */
-    public static final Double DOUBLE_ONE = Double.valueOf(1.0d);
+    public static final Double DOUBLE_ONE = 1.0d;
```
garydgregory, line 63: Hi @arturobernalg 
Thank you for your PR.
-1: For low-level libraries like most Apache Commons libraries, I like to keep boxing and unboxing explicit to make it easier to understand where objects are created and unwrapped.

garydgregory (changes requested):
See my previous comment: -1 don't change the boxing/unboxing.

PR #666, opened 2020-12-02
[LANG-1621] - Add JAVA_VM_COMPRESSED_MODE

arturobernalg:
Add  "java.vm.compressedOopsMode" properties  from system.

Do we need to return the value? getCompreesMode maybe ??


garydgregory:
Do you have an example use-case where this system property is needed?

arturobernalg:
HI,

I don't have a tangible example. I just checked that that property was missing and I thought that maybe at some point it would be useful

garydgregory:
> HI,
> 
> I don't have a tangible example. I just checked that that property was missing and I thought that maybe at some point it would be useful

Then let's not add it, there are a LOT of properties we don't cover anyway.

PR #667, opened 2020-12-03
[LANG-1622] Corrected reference to right methods.

kanaksony:
Corrected:
StringUtils.isNumericSpace incorrectly refer to isNumeric
BooleanUtils.toBoolean incorrectly refer to toBooleanObject


src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -548,9 +548,9 @@ public static Boolean toBooleanObject(final int value, final int trueValue, fina
      * if unboxed to a {@code boolean}.</p>
      *
      * <pre>
-     *   BooleanUtils.toBoolean(Integer.valueOf(0))    = Boolean.FALSE
```
kanaksony, line 551: @garydgregory I understand you merged PR 668 as it was reported by original reporter and had fix too.  If you will review my PR, I have corrected similar issue thing on other lines too that's from L551-L553.

Should I make another PR for same?

PR #668, opened 2020-12-03
[LANG-1622] Javadoc of some methods incorrectly refers to another method

anomen-s:


garydgregory:
https://issues.apache.org/jira/browse/LANG-1622

PR #669, opened 2020-12-03
Adding method to convert and format date

iashok22:
given Unix epoch time will be converted to date string with given date format.


garydgregory:
Hi @iashok22 

Thank you for your PR.

This seems to duplicate `FastDateFormat.getInstance(pattern).format(long)` which uses a format cache. 

The new method is the same as `new SimpleDateFormat(format).format(long)` (which does not use a cache) so I do not see the value here.

Am I missing something?

garydgregory:
No response in 2 months, closing.

PR #675, opened 2020-12-14
LANG-1444

bndhanush:
NumberUtils.createNumber(), BigDecimal for decimal fractions tending to zero

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -798,13 +798,14 @@ public class NumberUtils {
         try {
             final Float f = createFloat(str);
             final Double d = createDouble(str);
+            final BigDecimal b = createBigDecimal(str);
             if (!f.isInfinite()
                     && !(f.floatValue() == 0.0F && !allZeros)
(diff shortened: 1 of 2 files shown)


garydgregory:
@bndhanush 
This PR breaks the build, please see https://github.com/apache/commons-lang/pull/675/checks?check_run_id=1549559036

bndhanush:
Hi @garydgregory,

I have fixed the formatting issue. I see it is ok for the 16-ea build to fail, is that correct understanding? 
I am pretty new to OS, can you please help me in this?

garydgregory:
@chtompki 
May you please validate this one?

bndhanush:
@chtompki and @garydgregory. 
Can you please help in checking if the PR is valid. Also suggest if any changes are required.

garydgregory:
Hi @bndhanush 
Thank you for your PR.
In the new test, I do not think it is enough to check that an instance of right class is created, you want to check that the right _value_ is created to make sure we do not have the opposite issue: loosing precision.

bndhanush:
hi @garydgregory 

Thanks for the input. 
I added test case to ensure that precision is not lost and we maintain the value. 

Please validate the PR.

PR #677, opened 2020-12-19
LANG-1628: Javadoc correction for RandomStringUtils.random()

aviProgrammer:
Javadoc correction for RandomStringUtils.random() letters and numbers parameters to make it consistent with other overloaded random()  method Javadoc for the same parameters.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -345,8 +345,10 @@ public class RandomStringUtils {
      * @param count  the length of random string to create
      * @param start  the position in set of chars to start at (inclusive)
      * @param end  the position in set of chars to end before (exclusive)
-     * @param letters  only allow letters?
-     * @param numbers  only allow numbers?
+     * @param letters  if {@code true}, generated string may include
+     *  alphabetic characters
+     * @param numbers  if {@code true}, generated string may include
+     *  numeric characters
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -309,8 +309,10 @@ public static String random(final int count, final int start, final int end, fin
      * @param count  the length of random string to create
      * @param start  the position in set of chars to start at
      * @param end  the position in set of chars to end before
-     * @param letters  only allow letters?
-     * @param numbers  only allow numbers?
+     * @param letters  if {@code true}, generated string may include
+     *  alphabetic characters
+     * @param numbers  if {@code true}, generated string may include
+     *  numeric characters
```
kinow, line 315: Hi @aviProgrammer , thanks for your pull request.

It looks like a good contribution! And I was ready to approve it, but I noticed we can fix it in another line further down, where the same comments appear.

If you are using the GitHub UI to read this reply, you can click on the blue arrow here twice, and on lines 348 and 349 you should find these same parameters I think.

aviProgrammer: Hi @kinow , Thanks for your observation. 
The same fix of Javadoc has been applied at line 348 and 349. 

Let me know in case you find any other improvements that can be done in the same Javadoc.

kinow (approved):

kinow:
Merged. Thanks for your contribution @aviProgrammer !

PR #680, opened 2020-12-21
[LANG-1544] MethodUtils.invokeMethod NullPointerException in case of null in args list

mdbuck77:
- Null guards in place to handle one or more nulls specified as one of the parameters of the method to invoke.
- Check for an exact match of the actual parameter types against all of the methods on the class. This prevents picking an "upcasted" method (i.e. int specified but a method with a double is chosen).
- Throw an IllegalStateException with a helpful message if multiple candidate methods were found. This happens when multiple Methods had the same "distance" from the desired parameter types. Before this change the algorithm would just chose the first one.
- Tests for the above.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
@@ -30,13 +30,16 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.ClassUtils.Interfaces;
 import org.apache.commons.lang3.Validate;
+
+import static java.util.stream.Collectors.toList;
 
 /**
  * <p>Utility reflection methods focused on {@link Method}s, originally from Commons BeanUtils.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -742,35 +745,51 @@ public static Method getMatchingMethod(final Class<?> cls, final String methodNa
...
-        return inexactMatch;
+
+        final List<Method> bestCandidates = candidates.values().iterator().next();
+        if (bestCandidates.size() == 1) {
+            return bestCandidates.get(0);
+        }
+
+        final String target = methodName + Arrays.stream(parameterTypes).map(String::valueOf).collect(Collectors.joining(",", "(", ")"));
+        final String strCandidates = bestCandidates.stream().map(Method::toString).collect(Collectors.joining("\n  "));
+        throw new IllegalStateException("Found multiple candidates for method " + target + " on class " + cls + ":\n  " + strCandidates);
```
garydgregory, line 785: Let's not hard-code Linux-specific new-line chars in error messages, but, if you really want new-lines, use `%n` with `String.format()`.

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
```diff
@@ -1018,4 +1019,83 @@ public void testDistance() throws Exception {
...
+        assertThrows(IllegalStateException.class,
+                () -> MethodUtils.getMatchingMethod(GetMatchingMethodClass.class, "testMethod4", null, null));
+    }
+
+    private static final class GetMatchingMethodClass {
+        public String testMethod() {
+            return "testMethod";
+        }
+
+        public String testMethod(Long aLong) {
```
garydgregory, line 1061: Use final where you can, like here for parms.

garydgregory:
https://issues.apache.org/jira/browse/LANG-1544

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
```diff
@@ -1018,4 +1019,83 @@ public void testDistance() throws Exception {
...
+                GetMatchingMethodClass.class.getMethod("testMethod3", Long.class, Long.TYPE));
+
+        assertEquals(MethodUtils.getMatchingMethod(GetMatchingMethodClass.class, "testMethod3", Long.TYPE, null),
+                GetMatchingMethodClass.class.getMethod("testMethod3", Long.TYPE, Long.class));
+
+        assertThrows(IllegalStateException.class,
+                () -> MethodUtils.getMatchingMethod(GetMatchingMethodClass.class, "testMethod4", null, null));
+    }
+
+    private static final class GetMatchingMethodClass {
```
garydgregory, line 1056: Please make the test fixture as simple as possible: All of these methods can just return `void`.

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -742,35 +745,51 @@ public static Method getMatchingMethod(final Class<?> cls, final String methodNa
...
     }
 
     /**
      * <p>Returns the aggregate number of inheritance hops between assignable argument class types.  Returns -1
      * if the arguments aren't assignable.  Fills a specific purpose for getMatchingMethod and is not generalized.</p>
-     * @param classArray
-     * @param toClassArray
+     * @param classArray the Class array to calculate the distance from.
```
garydgregory, line 791: Maybe rename to `fromClassArray` to match `toClassArray`?

garydgregory (changes requested):
@mdbuck77

Gary

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
mdbuck77: I replaced the newlines with opening and closing brackets.

Michael

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
mdbuck77: Done.

Michael

src/test/java/org/apache/commons/lang3/reflect/MethodUtilsTest.java
mdbuck77: Done.

Michael

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
mdbuck77: Good suggestion. Done.

Michael

PR #682, opened 2020-12-22
[LANG-1631] - Check if Character is defined

arturobernalg:
Crash when the the char to be searched value it's out of range >=  '\u0000' <= '\uFFFF'
 Getting the next error -->
 
`java.lang.IllegalArgumentException: Not a valid Unicode code point: 0xFFFFF936`

Basically the problem is that when it is verified that the character to search is less than MIN_SUPPLEMENTARY_CODE_POINT, the search is not exited of the loop. It continues searching but this time values less than or equal to Character.MAX_CODE_POIN

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
+++ b/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
@@ -109,6 +109,7 @@ public class CharSequenceUtils {
                     return i;
                 }
             }
+            return NOT_FOUND;
         }
         //supplementary characters (LANG1300)
         if (searchChar <= Character.MAX_CODE_POINT) {
@@ -195,6 +196,7 @@ public class CharSequenceUtils {
                     return i;
                 }
             }
+            return NOT_FOUND;
         }
         //supplementary characters (LANG1300)
         //NOTE - we must do a forward traversal for this to avoid duplicating code points
@@ -244,7 +246,7 @@ public class CharSequenceUtils {
         }
 
         if (start < 0 || len2 < 0 || len2 > len1) {
-            return -1;
+            return NOT_FOUND;
         }
 
         if (len2 == 0) {
@@ -272,7 +274,7 @@ public class CharSequenceUtils {
             while (cs.charAt(i) != char0) {
                 i--;
                 if (i < 0) {
-                    return -1;
+                    return NOT_FOUND;
                 }
             }
             if (checkLaterThan1(cs, searchChar, len2, i)) {
@@ -280,7 +282,7 @@ public class CharSequenceUtils {
             }
             i--;
             if (i < 0) {
-                return -1;
+                return NOT_FOUND;
             }
         }
     }
--- a/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
@@ -21,6 +21,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.nio.CharBuffer;
 import java.util.Locale;
 
 import org.hamcrest.core.IsNot;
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -506,42 +510,57 @@ public void testIndexOfIgnoreCase_StringInt() {
...
+        charArray[1] = '&';
+        charArray[7] = '{';
+        charArray[3] = '.';
+        charArray[4] = 'c';
+        charArray[5] = '.';
+        charArray[6] = '0';
+        charArray[7] = 'r';
+        charArray[8] = 'o';
+        CharBuffer charBuffer = CharBuffer.wrap(charArray);
+        assertEquals(NOT_FOUND, StringUtils.lastIndexOf (charBuffer, (-1738), 982));
```
garydgregory, line 543: Hi @arturobernalg 

Thank you for your PR. 

PS: It's harder to deal with a PR like this because it mixes a change with some clean ups (-1 -> NOT_FOUND) so I am not sure where the new test is without trolling through all the noise :-(

Why is the number in parentheses? If this is a test for the actual bug fix, then you should add a sanity check to validates that the char is indeed bogus.

TY.

arturobernalg: HI @garydgregory 
sorry for the noise. When I created the PR it seemed logical.  But you're right. Better keep the PRs separate

garydgregory: Hi @arturobernalg 
You've not answered all my question.
Also why are you building the char[] one char at a time? Why not call CharBuffer.wrap(String)?

arturobernalg: mmm 

- The parentheses was a mistake
- The Character in the test there are not valid unicode values
- I'll change  the test for --> CharBuffer.wrap(String)

garydgregory:
@arturobernalg 
Now, I'm confused. There are now no changes here for "Check if Character is defined".

garydgregory (changes requested):
@arturobernalg
Now, I'm confused. There are now no changes here for "Check if Character is defined".

src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
```diff
@@ -195,6 +196,7 @@ static int lastIndexOf(final CharSequence cs, final int searchChar, int start) {
                     return i;
                 }
             }
+            return NOT_FOUND;
```
arturobernalg, line 199: mm . Maybe it's my fault. 
The idea it's exit the loop and return -1 (NOT_FOUND) if not found the charset

src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
```diff
@@ -109,6 +109,7 @@ static int indexOf(final CharSequence cs, final int searchChar, int start) {
...
```
arturobernalg, line 112: idem @garydgregory

arturobernalg (commented):
The idea it's exit the loop and return -1 (NOT_FOUND) if not found the charset

src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
garydgregory: ah, I see, this is a new line, and the other mods are updates. ok, I'll take a look tomorrow probably...

arturobernalg: sorry. my fault for mixing.

src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -48,6 +49,8 @@
 
     private static final String[] FOOBAR_SUB_ARRAY = new String[] {"ob", "ba"};
 
+    private static final int NOT_FOUND = -1;
```
garydgregory, line 52: This is already defined in `ArrayUtils`.

arturobernalg: true. changed

PR #683, opened 2020-12-25
[LANG-1632] -Implementation of indexOf

arturobernalg:


XenoAmess:
So why this but not [LANG-1572]? Is there any strong evidence that shows this version works faster?

PR #684, opened 2020-12-28
[LANG-1634] Add ObjectUtils #applyIfNonNull and #applyFirstNonNull

bindul:
Utility methods that take a java.util.function.Consumer and possibly null value(s). The consumer is invoked if the value is not null or with the first non-null value, respectively.

See [LANG-1634](https://issues.apache.org/jira/browse/LANG-1634) and discussion at [\[lang\] ObjectUtils enhancement - consumer with non-null value](https://lists.apache.org/thread.html/r5c02039bc54d455f032d06bc196aa7c59d7673a100684272961007e3%40%3Cdev.commons.apache.org%3E)


src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -226,6 +227,62 @@ public static boolean anyNotNull(final Object... values) {
...
+     * Invokes the given {@code consumer's} {@link Consumer#accept(Object)} with the first {@code non-null} value from
+     * {@code objects}. If all the values are null, the consumer is not invoked.
+     * </p>
+     *
+     * <p>
+     * The caller is responsible for thread-safety and exception handling of consumer.
+     * </p>
+     *
+     * <pre>
+     * ObjectUtils.applyIfNonNull(bean::setValue, null)                 - setValue not invoked
```
sgflt, line 241: typo in method name

bindul: Corrected, thanks.

garydgregory:
It seems to me the true test of a new API like `applyIfNonNull` is where can it be used within Commons Lang, if we don't eat our own dog food, it seems odd to tell others to do so. IOW, I'd like to see this PR include using this API.

I'm not convinced as to the utility of `applyFirstNonNull`, see above.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -226,6 +227,62 @@ public static boolean anyNotNull(final Object... values) {
         return firstNonNull(values) != null;
     }
 
+    /**
+     * <p>
+     * Invokes the given {@code consumer's} {@link Consumer#accept(Object)} with the first {@code non-null} value from
```
garydgregory, line 232: IMO, you're not "invoking", you're "accepting", so the Javadoc should say "Accepts...". Actually, the method is misnamed since the underlying method is `Consumer#accept()`, this method should be use "accept", not "apply".

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -226,6 +227,62 @@ public static boolean anyNotNull(final Object... values) {
...
+     * ObjectUtils.applyIfNonNull(v -&gt; bean.setValue(v), "abc")  - setValue invoked with "abc"
+     * </pre>
+     *
+     * @param <T> the type of the object
+     * @param object the {@code Object} to test, may be {@code null}
+     * @param consumer the consumer operation to invoke with {@code object} if it is {@code non-null}
+     * @see #applyFirstNonNull(Consumer, Object...)
+     * @since 3.12
+     */
+    public static <T> void applyIfNonNull(final Consumer<T> consumer, final T object) {
```
garydgregory, line 280: See "accept" comment above and other comment in the Conversation stream of this PR.

src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
```diff
@@ -764,4 +791,13 @@ public int compare(final CharSequence o1, final CharSequence o2) {
 
     }
 
+    static final class ApplyIfNonNullBean {
+        private String value;
+        public String getValue() {
+            return value;
+        }
+        public void setValue(String value) {
+            this.value = value;
```
garydgregory, line 800: You can add a call to `Objects.requireNonNull()` here instead of the blind assignment:
```
this.value = Objects.requireNonNull(value, "value");
```

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/ObjectUtils.java
bindul: I have changed the method name to 'accept...' and updated the Javadoc comment accordingly.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
bindul: I have changed the method name to 'accept...', switched the parameter order and updated the Javadoc comment accordingly.

src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
bindul: This inner class was meant as a simple Java bean with a setter and getter, hence did not have any checks in it. I have changed it now.

bindul:
@garydgregory 

I have renamed the methods to accept from apply and switched the parameter order where possible.

`acceptFirstNonNull` is the analogous method to existing `ObjectUtils.firstNonNull` or `StringUtils.firstNonBlank`. If I remember correctly, I have had one use for this method in my real life code, where a configuration parameter value could be supplied at multiple levels (specific instance, application default or global default; the first two could be null) and I have used a similar method to handle this. If you don't want to include it, I am happy to remove this.

In a real life scenario, the use case for the `acceptIfNonNull` method is usually when setting configuration values. The driver (say for something like redis, solr, etc.) or library (like a HTTP client library) accepts a configuration bean with some default values set. Your application accepts configuration values, and you only want to set those in the configuration bean if they are not null (or you lose the default). Anyways, I have updated 3 existing commons-lang classes to show how it can reduce 3 lines of code to 1. I did not go and update every possible occurrence, as that would make this PR large and lose focus on the primary change in the PR.

Please review and let me know if I should make further modifications.

garydgregory:
@bindul Please rebase on master. TY!

bindul:
@garydgregory Done!
Please let me know if you want me to squash commits into logical groups (I am thinking 1 with the change and the changes for review comments and another where the new method is used elsewhere in lang)

singhbaljit:
We could really use this method. This is helpful because it removes the if/else code branch from the application codes, and therefore, devs don't have to write trivial unit tests to meet the code coverage requirements.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -226,6 +227,63 @@ public static boolean anyNotNull(final Object... values) {
...
+     * ObjectUtils.acceptIfNonNull("abc", v -&gt; bean.setValue(v))  - setValue invoked with "abc"
+     * </pre>
+     *
+     * @param <T> the type of the object
+     * @param object the {@code Object} to test, may be {@code null}
+     * @param consumer the consumer operation to invoke with {@code object} if it is {@code non-null}
+     * @see #acceptFirstNonNull(Consumer, Object...)
+     * @since 3.12
+     */
+    public static <T> void acceptIfNonNull(final T object, final Consumer<T> consumer) {
```
singhbaljit, line 281: more user-friendly: `Consumer<? super T> consumer`. 

Also, `requireNonNull(consumer, "consumer")`.

PR #685, opened 2020-12-29
[LANG-1635] - Count True function improvement

arturobernalg:
Function that returns the number of values that are true


sgflt:
It seems we have similar use case, also BooleanUtils.countTrue(T, Predicate<? super T>...) could be more flexible.

garydgregory:
Add a LINK to the JIRA in description please, otherwise it's a pain to find what this is about.

arturobernalg:
I close the PR. Thank you

PR #686, opened 2020-12-29
[LANG-1636] - Boolean Join Function

arturobernalg:
Join Implementation it's missing for primitive of type boolean

https://issues.apache.org/jira/browse/LANG-1636


garydgregory:
Indeed, boolean[] is missing for this API, thanks @arturobernalg

PR #691, opened 2021-01-03
DiffBuilder.java: remove unnecessary boxing for primitives

akarimin:


garydgregory:
-1: We keep boxing and unboxing in low-level libraries like Commons Lang to make explicit where object allocation may occur and where NPEs may be thrown.

akarimin:
Ok thanks, will abandon PR.

PR #693, opened 2021-01-06
[LANG-1614] - Split fixed-length elements

arturobernalg:
Create a new method that split a string into an array of fixed-length elements


arturobernalg:
any thoughts here  @garydgregory  ;)
should/could it  added to commons-lang?

PR #696, opened 2021-01-11
LANG-1639 Add getEnumByPredicate

bonii-xx:
Adds a method to obtain a matching enum value based on a predicate. Quite useful when mapping values to enums.

I used Optional as the return type since it allows the removal of a default value without removing any functionality, the caller has to handle the returned value if it's not found.


src/test/java/org/apache/commons/lang3/EnumUtilsTest.java
```diff
@@ -163,6 +165,26 @@ public void test_getEnumIgnoreCase_defaultEnum() {
...
+    @Test
+    public void test_getEnumByPredicate_nullInputs() {
+        assertFalse(EnumUtils.getEnumByPredicate(null, WithExtraData::isA).isPresent());
+        assertFalse(EnumUtils.getEnumByPredicate(WithExtraData.class, null).isPresent());
+    }
+
+    @Test
+    public void test_getEnumByPredicate_nonEnumClass() {
+        final Class rawType = Object.class;
+        assertFalse(EnumUtils.getEnumByPredicate(rawType, null).isPresent());
```
garydgregory, line 185: This test does not cover checking the class' enum aspect because the predicate is null.

bonii-xx: Thanks, had it named differently before, but decided to align the naming with the other enum tests 🤦

bonii-xx: Ah, I misread the comment. Yes you're right, the test is faulty.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -264,6 +266,35 @@ public static <E extends Enum<E>> E getEnumIgnoreCase(final Class<E> enumClass,
...
+    public static <E extends Enum<E>> Optional<E> getEnumByPredicate(final Class<E> enumClass, final Predicate<E> predicate) {
+        if (predicate == null || enumClass == null || !enumClass.isEnum()) {
+            return Optional.empty();
+        }
+        for (final E each : enumClass.getEnumConstants()) {
+            if (predicate.test(each)) {
+                return Optional.of(each);
+            }
+        }
+        return Optional.empty();
```
biya-bi, line 295: Since you are using Java 8 here, I think lines 290 to 295 could be replaced with the below:

`return Arrays.stream(enumClass.getEnumConstants()).filter(predicate).findFirst()`

The findFirst() method will never return null in this case because null is not a valid value of any enum.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -264,6 +266,35 @@ public static <E extends Enum<E>> E getEnumIgnoreCase(final Class<E> enumClass,
...
+     * or context information regarding the enum values.</p>
+     *
+     * @param <E>         the type of the enumeration
+     * @param enumClass   the class of the enum to query, not null
+     * @param predicate   the predicate used to test for the value, not null
+     * @return an Optional containing the enum or an empty optional if none was found, or any input is invalid
+     * @since 3.12
+     */
+    public static <E extends Enum<E>> Optional<E> getEnumByPredicate(final Class<E> enumClass, final Predicate<E> predicate) {
+        if (predicate == null || enumClass == null || !enumClass.isEnum()) {
```
biya-bi, line 287: I think we can remove `!enumClass.isEnum()` check because generic type (`<E extends Enum<E>> Optional<E>`) in the method signature already ensures that enumClass is indeed an enum. If the class type is not an enum, there will be a compile time error.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -264,6 +266,35 @@ public static <E extends Enum<E>> E getEnumIgnoreCase(final Class<E> enumClass,
...
+            return Optional.empty();
+        }
```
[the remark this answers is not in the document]
biya-bi: From my understanding, the idea is to pick elements that satisfy the passed in predicate. We may or may not be interested in all matches. In certain scenarios, just knowing that one occurance exists is enough for certain processing. The short-circuiting terminal operation findFirst() helps in such cases.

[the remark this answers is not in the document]
biya-bi: Didn't write my comment at the right place. I was trying to reply to your earlier comment on picking the first element.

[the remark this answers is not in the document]
biya-bi: I however agree that the bad parameters should be reported with an exception.

biya-bi:
Did you see my first comment on about replacing lines 290 to 295 by a single line of code?

biya-bi:
What is mean is rather than writing many lines of code (the for loop and the return statment) to return a single value, we can take advantage of short-circuiting to accomplish the same thing in a more concise way.

biya-bi:
The question is what output are we expecting from the method?
`Arrays.stream(enumClass.getEnumConstants()).filter(predicate);` will return a stream. We don't want to give more work to the caller of the getEnumByPredicate method.
If we want the returned value to be a list, we need to append `.collect(Collectors.toList())`.
If we want the returned value to be a single value, we need to pick the first element from the filter's result, in one way or the other.

biya-bi:
Ok, so your proposal is returning a list?

biya-bi:
That's fine. It is true that this method does not really add value to the library.

PR #704, opened 2021-01-31
Added new percentage check circuit breaker

saurabhsinghwww:
Added new circuit breaker implementation (PercentageCheckCircuitBreaker), which is based on percentage checks on the events.

A circuit breaker can be used to protect an application against unreliable services or unexpected load.
A newly created {@code PercentageCheckCircuitBreaker} object is initially in state closed meaning that no problem has been detected.
When the application encounters specific events (like errors or service timeouts), it tells the circuit breaker to increment the internal counters.
If the number of events reported results in a higher than acceptable percentage for a window size, the circuit breaker changes into state open.
This means that there is a problem with the associated sub system, the application should no longer call it, but give it some time to settle down.
The circuit breaker can be configured to switch back or reset to closed state after a certain time frame.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/PercentageCheckCircuitBreaker.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/PercentageCheckCircuitBreaker.java
@@ -1,0 +1,257 @@
+package org.apache.commons.lang3.concurrent;
+
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


saurabhsinghwww:
.

saurabhsinghwww:
By mistake created this Pull Request on the main repository instead of my forked repository.

PR #712, opened 2021-02-17
Fix broken Javadoc links to commons-text

smowton:
All were missing a subpackage


garydgregory:
Merged. TY @smowton !

PR #713, opened 2021-02-19
Let mid(null), left(null) and right(null) return empty

senderic:
I'm not sure if there is reasoning behind `left(null, n)` returning `null`, but other `StringUtils` functions tends to return empty strings if their inputs are null. Likewise, I think it makes sense that `mid(null, n)`, `left(null, n)` or `right(null, n)` returns an empty string instead of returning null. My use case is lambda logging. I have code like this:

```.java
log.debug("My output (first 256 chars): {}", StringUtils.left(output, 256)::toString);
```

I get an NullPointerExceptions when I do and not properly wrap the call, etc. So I can't take advantage of both the `left` function as well as lambda based logging.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -5221,7 +5221,7 @@ public class StringUtils {
      * an exception. An empty String is returned if len is negative.</p>
      *
      * <pre>
-     * StringUtils.left(null, *)    = null
+     * StringUtils.left(null, *)    = ""
      * StringUtils.left(*, -ve)     = ""
      * StringUtils.left("", *)      = ""
      * StringUtils.left("abc", 0)   = ""
@@ -5231,13 +5231,10 @@ public class StringUtils {
      *
      * @param str  the String to get the leftmost characters from, may be null
      * @param len  the length of the required String
-     * @return the leftmost characters, {@code null} if null String input
+     * @return the leftmost characters, {@code EMPTY} if null String input
      */
     public static String left(final String str, final int len) {
-        if (str == null) {
-            return null;
-        }
-        if (len < 0) {
+        if (str == null || len < 0) {
             return EMPTY;
         }
         if (str.length() <= len) {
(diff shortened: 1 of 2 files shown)


esend7881:
> 
> [Coverage Status](https://coveralls.io/builds/37285941)
> 
> Coverage decreased (-0.002%) to 94.955% when pulling **821a4a5b04f6fb4bbdcec014be48bf42df5f48ca on esend7881:patch-1** into **5689c91cf2a3a27f0a3a8a362857f2cf0919d4f6 on apache:master**.
> 

I think that is because the total number of lines have been reduced (smaller denominator) which makes it seem like there is less coverage.

garydgregory:
Not sure about the impact here of changing the behavior... needs a bit more study...

garydgregory:
We are not consistent in `StringUtils`, some methods return `null` on `null` input, others return an empty string. Changing this would likely lead to unexpected results for call sites in apps and other libraries. If we change anything, we should do so consistently in the whole class. At least the Javadocs say what to expect. We should probably do this for 4.0 IMO.

Any thoughts?

ppkarwasz:
> We are not consistent in `StringUtils`, some methods return `null` on `null` input, others return an empty string. Changing this would likely lead to unexpected results for call sites in apps and other libraries. If we change anything, we should do so consistently in the whole class. At least the Javadocs say what to expect. We should probably do this for 4.0 IMO.

Personally I don't like method that return `null` on `null` input. Among the nullness checkers I know, only the Checker Framework supports this behavior with the [`@PolyNull` annotation](https://checkerframework.org/manual/#nullness-annotations).

The "polynull" pattern has also the tendency to contaminate your code: you write a lot of "polynull" methods, so that you can compose those methods. You could save many LoCs by just checking for `null` once a having methods that don't return `null`.

garydgregory:
Should we start a discussion on the dev ML as to whether all StringUtils should be normalized null to null vs. null to "" and if so for 3.x or 4.0?

ppkarwasz:
Absolutely,

We should discuss it on `dev@commons`.

garydgregory:
OK, I'll start a thread.

garydgregory:
I think this PR should be moved to draft mode (or closed?) pending the discussion on the dev mailing list.

ppkarwasz:
Almost certainly these changes can only be applied in `4.x`, so we should probably close it.

garydgregory:
Closing per the above discussion .

PR #716, opened 2021-02-23
Reuse the boolean array validations.

arturobernalg:
The idea it's reuse a single method  validate that the specified argument array is neither null  nor a length of zero (no elements). otherwise throwing an exception

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BooleanUtils.java
+++ b/src/main/java/org/apache/commons/lang3/BooleanUtils.java
@@ -29,7 +29,18 @@ import org.apache.commons.lang3.math.NumberUtils;
  * @since 2.0
  */
 public class BooleanUtils {
-
+    /**
+     * The default message for null array.
+     *
+     * @since 3.12.0
+     */
+    private static final String DEFAULT_NULL_ARRAY_EX_MESSAGE = "The Array must not be null";
+    /**
+     * The default message for empty array.
+     *
+     * @since 3.12.0
+     */
+    private static final String DEFAULT_EMPTY_ARRAY_EX_MESSAGE = "Array is empty";
     /**
      * The false String {@code "false"}.
      *
@@ -92,12 +103,7 @@ public class BooleanUtils {
      */
     public static boolean and(final boolean... array) {
         // Validates input
-        if (array == null) {
-            throw new IllegalArgumentException("The Array must not be null");
-        }
-        if (array.length == 0) {
-            throw new IllegalArgumentException("Array is empty");
-        }
+        notNullOrEmpty(array);
         for (final boolean element : array) {
             if (!element) {
                 return false;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -263,6 +267,23 @@ public static Boolean negate(final Boolean bool) {
...
+    /**
+     * <p>Validate that the specified argument array is neither {@code null}
+     * nor a length of zero (no elements); otherwise throwing an exception.
+     *
+     * @param array  an array of {@code boolean}s
+     * @throws IllegalArgumentException if {@code array} is {@code null}
+     * @throws IllegalArgumentException if {@code array} is empty.
+     */
+    private static void notNullOrEmpty(final boolean... array){
+        if (array == null) {
```
garydgregory, line 279: Hi @arturobernalg 
I think we should converge all null checks to `Objects.requireNonNull(arg, "arg")`, which also avoids making our exception messages too English-centric. In the same vein, I'd call this new method `requireNotEmpty`.

src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -29,7 +29,18 @@
  * @since 2.0
  */
 public class BooleanUtils {
-
+    /**
+     * The default message for null array.
+     *
+     * @since 3.12.0
```
garydgregory, line 35: No need for since tag for private elements.

src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -29,7 +29,18 @@
...
+     */
+    private static final String DEFAULT_NULL_ARRAY_EX_MESSAGE = "The Array must not be null";
+    /**
+     * The default message for empty array.
+     *
+     * @since 3.12.0
```
garydgregory, line 41: No need for since tag for private elements.

src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -29,7 +29,18 @@
...
+     */
+    private static final String DEFAULT_EMPTY_ARRAY_EX_MESSAGE = "Array is empty";
```
garydgregory, line 43: Simplify to "Empty array".

garydgregory (changes requested):
Hi @arturobernalg 

arturobernalg:
> Hi @arturobernalg
> Thank you for your PR. Please see my comments.
> Ses if you can reuse:
> 
> * `ObjectUtils.isEmpty(Object)`
> * `ObjectUtils.isNotEmpty(Object)`
>   We could add:
> * `ObjectUtils.requireNotEmpty(Object)`
> * `ObjectUtils.requireNotEmpty(Object, String)`

Hi @garydgregory 

I'll check how 

> Hi @arturobernalg
> Thank you for your PR. Please see my comments.
> Ses if you can reuse:
> 
> * `ObjectUtils.isEmpty(Object)`
> * `ObjectUtils.isNotEmpty(Object)`
>   We could add:
> * `ObjectUtils.requireNotEmpty(Object)`
> * `ObjectUtils.requireNotEmpty(Object, String)`

Hi @garydgregory 

Thank you for your tips. Is this what you are looking for? I've created a new PR to maintain clean the other one in case I screwed up ;)
https://github.com/apache/commons-lang/pull/717

garydgregory:
Closing in favor of a different implementation in git master which mirror the style and names of java.lang.Objects.requireNon*() methods.

PR #718, opened 2021-02-24
Use own validator ObjectUtils.anyNull to check null String input

arturobernalg:


garydgregory:
Eating our own dog food :-)

PR #719, opened 2021-02-25
SimpleStopWatch is alternative for overcomplicated and buggy StopWatch

michaldo:
In https://github.com/apache/commons-lang/pull/709 I proposed add `lap()` to StopWatch.

During discussion I realized that StopWatch is overcomplicated and better is option is no-brainer SimpleStopWatch. 
With a `lap()`, of course


garydgregory:
Closing: Old PR, no real need IMO, and duplicates functionality from a user's POV.

PR #720, opened 2021-02-25
Reuse ObjectUtils.anyNull

arturobernalg:


arturobernalg:
Reuse ObjectUtils.anyNull when need to instead of check if (obj == null || obj == null) {

PR #725, opened 2021-03-02
Add toLocalDate to CalendarUtils

asgh:
Using an Instance to create a LocalDate from a Calendar does not work on dates before the Gregorian/Julian cutover.

So use this code instead, and add tests for years before the cutover and after.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
@@ -22,7 +22,6 @@ import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
-
 import java.util.Calendar;
 import java.util.Locale;
 import java.util.Locale.Category;
--- a/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
@@ -17,10 +17,12 @@
 
 package org.apache.commons.lang3.time;
 
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+
 import java.util.Calendar;
 import java.util.Locale;
 import java.util.Locale.Category;
@@ -223,4 +225,11 @@ public class CalendarUtils {
         return toZonedDateTime(calendar);
     }
 
+     * Creates a LocalDate from a Calendar
+     *
+     * @return a LocalDate
+     */
+    public LocalDate toLocalDate() {
+        return LocalDate.of(getYear(), getMonth()+1, getDayOfMonth());
+    }
 }
--- a/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
@@ -224,9 +224,10 @@ public class CalendarUtils {
         return toZonedDateTime(calendar);
     }
 
-     * Creates a LocalDate from a Calendar
+     * Converts this instance to a {@link LocalDate}.
      *
      * @return a LocalDate
+     * @since 3.17.0
      */
     public LocalDate toLocalDate() {
         return LocalDate.of(getYear(), getMonth()+1, getDayOfMonth());
--- a/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
@@ -224,12 +224,12 @@ public class CalendarUtils {
         return toZonedDateTime(calendar);
     }
 
-     * Converts this instance to a {@link LocalDate}.
+    /** Converts this instance to a {@link LocalDate}.
      *
      * @return a LocalDate
      * @since 3.17.0
      */
     public LocalDate toLocalDate() {
-        return LocalDate.of(getYear(), getMonth()+1, getDayOfMonth());
+        return LocalDate.of(getYear(), getMonth() + 1, getDayOfMonth());
     }
 }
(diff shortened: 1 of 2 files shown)


asgh:
The failing tests appear to be from other code, not the new code in the pull request.

garydgregory:
@asgh 

Would you please rebase on master?

asgh:
@garydgregory 

Done.

garydgregory:
@asgh 
Thank you for the update. Please run 'mvn' by itself before you push to catch all errors.

asgh:
@garydgregory 

I don't have mvn on this computer, I hope my last commit fixed things, if not I'll have to work on it a little later.

src/test/java/org/apache/commons/lang3/time/CalendarUtilsTest.java
```diff
@@ -83,4 +87,12 @@ public void testGetYear() {
         assertEquals(Calendar.getInstance().get(Calendar.YEAR), CalendarUtils.INSTANCE.getYear());
     }
 
+    @Test
+    public void testToLocalDate() {
+        Calendar calendar = new GregorianCalendar(TimeZone.getTimeZone("GMT"));
```
garydgregory, line 92: FYI: You can reuse `org.apache.commons.lang3.time.TimeZones.GMT` here.

garydgregory (commented):
I'm not sure the implementation is correct. The following fails with this PR:
```java
    @ParameterizedTest
    @MethodSource("java.util.TimeZone#getAvailableIDs()")
    public void testToLocalDate(final String id) {
        final TimeZone timeZone = TimeZone.getTimeZone(id);
        final Calendar calendar = new GregorianCalendar(timeZone);
        // sanity check
        assertEquals(LocalDateTime.ofInstant(calendar.toInstant(), timeZone.toZoneId()).toLocalDate(), new CalendarUtils(calendar).toLocalDate());
        // actual test
        calendar.setTimeInMillis(-27078001200000L);
        assertEquals(LocalDate.of(1111, Month.DECEMBER, 1), new CalendarUtils(calendar).toLocalDate());
        calendar.setTimeInMillis(1614700215000L);
        assertEquals(LocalDate.of(2021, Month.MARCH, 2), new CalendarUtils(calendar).toLocalDate());
    }
```

src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
```diff
@@ -151,4 +152,13 @@ String[] getStandaloneShortMonthNames() {
     public int getYear() {
         return calendar.get(Calendar.YEAR);
     }
+
+    /**
+     * Creates a LocalDate from a Calendar
```
garydgregory, line 157: The new objects is not based on "a" Calendar", it's based on _this instance_.

garydgregory:
@asgh 
Please rebase on git master to pick up the latest from this class.

src/test/java/org/apache/commons/lang3/time/CalendarUtilsTest.java
asgh: fixed

asgh:
@garydgregory 

> I'm not sure the implementation is correct. The following fails with this PR:

That's exactly why I added this!

Dealing with local dates that cross timezones is a mess, LocalDate specifically has no timezone, the other way also does not correctly handle early dates. It's the other code that is incorrect.

If you feel it's important I can try to dive into the debugger and find where your pr goes wrong.

src/test/java/org/apache/commons/lang3/time/CalendarUtilsTest.java
```diff
@@ -128,5 +131,13 @@ public void testToZonedDateTime(final String id) {
         final ZonedDateTime zdt1 = ZonedDateTime.of(1, 2, 3, 4, 5, 6, 0, zoneId);
         calendar.setTimeInMillis(zdt1.toInstant().toEpochMilli());
         assertEquals(ZonedDateTime.ofInstant(zdt1.toInstant(), calendar.getTimeZone().toZoneId()), new CalendarUtils(calendar).toZonedDateTime());
+
+    @Test
```
garydgregory, line 135: Hello @asgh,

Thank you for your reply and update.

I don't think the implementation is correct because it does not take into account the Calendar's time zone.

You should be able to see this if you test with time zones and adjust the test as in my example test using:

```java
    @ParameterizedTest
    @MethodSource(TimeZonesTest.TIME_ZONE_GET_AVAILABLE_IDS)
```

See also the other tests in this class.

TY!

asgh: @garydgregory 

I'm really not following you. For this to be wrong this.getDayOfMonth() would have to be wrong. Are you saying that function returns the wrong day of the month?

Also your Assert is comparing an object with a Time to one without a Time - obviously they are going to be different!

And the entire point is that LocalDate doesn't *have* a timezone. I'm only setting the timezone in the test so that it correctly converts the UnixTime into a Month and Day (otherwise it could vary +- a day depending on what timezone you pick).

If you look at the actual function in the code there is no timezone there are all.

garydgregory: Hello @asgh 

One of the reasons for confusion perhaps is that the test data does not look right.

How did you get from `-27078001200000` to `LocalDate.of(1111, Month.DECEMBER, 1)` when:
```
    public void testToLocalDate() {
        final Calendar calendar = new GregorianCalendar(TimeZone.getTimeZone(TimeZones.GMT_ID));
        calendar.setTimeInMillis(-27078001200000L);
        assertEquals("1111-12-08T05:00:00Z", calendar.toInstant().toString());
        assertEquals(LocalDate.of(1111, Month.DECEMBER, 8), new CalendarUtils(calendar).toLocalDate());
        calendar.setTimeInMillis(1614700215000L);
        assertEquals(LocalDate.of(2021, Month.MARCH, 2), new CalendarUtils(calendar).toLocalDate());
    }
```
?
See [`CalendarUtilsTest.testToLocalDate()`](https://github.com/apache/commons-lang/blob/3a8293af2b8cdab191a0e7ca6e306e0a2ac1e9e8/src/test/java/org/apache/commons/lang3/time/CalendarUtilsTest.java#L100-L107).

garydgregory:
Closing: [org.apache.commons.lang3.time.CalendarUtils.toLocalDate()](https://github.com/apache/commons-lang/blob/6cf0236d4e5687cd29c6351ca54e9a2f250fff50/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java#L203) will be in 3.18.0, see git master and snapshot builds in https://repository.apache.org/content/repositories/snapshots/

PR #729, opened 2021-03-05
LANG-1647 - Check if an throwable is a checked exception

arturobernalg:
The idea it's have a function that check if a given throwable is a checked exception. There are some similar function in ConcurrentUtils, but have package visibility and return the same exception. 
Seem logic have this verification in the class that have this responsibility - `ExceptionUtils`

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -45,10 +45,12 @@ The <action> type attribute can be add,update,fix,remove.
   </properties>
   <body>
 
-  <release version="3.12.1" date="2021-MM-DD" description="New features and bug fixes (Java 8).">
+  <release version="3.13.0" date="2021-MM-DD" description="New features and bug fixes (Java 8).">
     <!-- FIX -->
     <action issue="LANG-1645" type="fix" dev="aherbert" due-to="Alex Herbert">NumberUtils.createNumber to recognise hex integers prefixed with +.</action>
     <action issue="LANG-1646" type="fix" dev="aherbert" due-to="Alex Herbert">NumberUtils.createNumber to return requested floating point type for zero.</action>
+    <!-- ADD -->
+    <action issue="LANG-1647" type="add" dev="abernal"  due-to="Arturo Bernal">Check if an throwable is a checked exception.</action>
   </release>
 
   <release version="3.12.0" date="2021-02-26" description="New features and bug fixes (Java 8).">
--- a/src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
+++ b/src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
@@ -586,6 +586,30 @@ public class ExceptionUtils {
         return indexOf(throwable, type, fromIndex, true);
     }
 
+    /**
+     * Return whether the given throwable is a checked exception:
+     * that is, neither a RuntimeException nor an Error.
+     *
+     * <pre>
+     * ExceptionUtils.isChecked(new Exception())                    = true
+     * ExceptionUtils.isChecked(new SQLException())                 = true
+     * ExceptionUtils.isChecked(new RuntimeException())             = false
+     * ExceptionUtils.isChecked(new IllegalArgumentException(""))   = false
+     * ExceptionUtils.isChecked(new Throwable())                    = true
+     * ExceptionUtils.isChecked(null)                               = false
+     * </pre>
+     *
+     * @param throwable the throwable to check
+     * @return whether the throwable is a checked exception
+     * @see java.lang.Exception
+     * @see java.lang.RuntimeException
+     * @see java.lang.Error
+     * @since 3.13.0
+     */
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
```diff
@@ -586,6 +586,30 @@ public static int indexOfType(final Throwable throwable, final Class<? extends T
...
+     *
+     * @param throwable the throwable to check
+     * @return whether the throwable is a checked exception
+     * @see java.lang.Exception
+     * @see java.lang.RuntimeException
+     * @see java.lang.Error
+     * @since 3.12.1
+     */
+    public static boolean isCheckedException(final Throwable throwable) {
+        return throwable != null && !(throwable instanceof RuntimeException || throwable instanceof Error);
```
aherbert, line 610: `instanceof` handles null. Does this work if you remove the `throwable != null &&`?

I would expect in the majority of cases this would be called with a non-null object so this simplifies the code.

src/changes/changes.xml
```diff
@@ -100,6 +100,7 @@ The <action> type attribute can be add,update,fix,remove.
     <action                   type="add" dev="jochen">Introduce the use of @Nonnull, and @Nullable, and the Objects class as a helper tool.</action>
     <action                   type="add" dev="ggregory" due-to="Arturo Bernal, Gary Gregory">Add and use true and false String constants #714.</action>
     <action                   type="add" dev="ggregory" due-to="Arturo Bernal, Gary Gregory">Add and use ObjectUtils.requireNonEmpty() #716.</action>
+    <action issue="LANG-1647" type="add" dev="abernal"  due-to="">Check if an throwable is a checked exception.</action>
```
aherbert, line 103: This is in the wrong place

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
arturobernalg: I prefer to leave all options open. it is more versatile. ...

src/changes/changes.xml
arturobernalg: true. thanks. @aherbert

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
aherbert: My point was that you do not need the `throwable != null`. It should still pass your unit tests (i.e. return false if null is passed).

arturobernalg: No. Return true without the null check

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
```diff
@@ -586,6 +586,30 @@ public static int indexOfType(final Throwable throwable, final Class<? extends T
...
+     * </pre>
+     *
+     * @param throwable the throwable to check
+     * @return whether the throwable is a checked exception
+     * @see java.lang.Exception
+     * @see java.lang.RuntimeException
+     * @see java.lang.Error
+     * @since 3.12.1
+     */
+    public static boolean isCheckedException(final Throwable throwable) {
```
garydgregory, line 609: The API name is misleading since it takes more than Exception as input. 
Going further, I think the type name in the method name is redundant and calling the API 'isChecked' should be enough.

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
```diff
@@ -586,6 +586,30 @@ public static int indexOfType(final Throwable throwable, final Class<? extends T
...
+     * ExceptionUtils.isCheckedException(new Throwable())                    = true
+     * ExceptionUtils.isCheckedException(null)                               = false
+     * </pre>
+     *
+     * @param throwable the throwable to check
+     * @return whether the throwable is a checked exception
+     * @see java.lang.Exception
+     * @see java.lang.RuntimeException
+     * @see java.lang.Error
+     * @since 3.12.1
```
garydgregory, line 607: Adding a new API is not a maintenance version change per semantic versioning guidelines, the version should be the next minor one.

arturobernalg: HI @garydgregory 

Don't understand. what version should u put in order to maintain the guidelines?

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
arturobernalg: Changed

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
arturobernalg: Hi @garydgregory 
Changed

PR #730, opened 2021-03-14
EnumUtils: Add more-generic getEnumMap() method.

MaxwellCody:
The method accepts a Function object used to supply keys for the map, allowing for keys other than the value names. This may be useful when mapping constants to data.

The behavior of the single-parameter `getEnumMap()` can be emulated with this method by using a reference to the `Enum::name` method. For example,

```groovy
Map<String, Traffic> map = EnumUtils.getEnumMap(Traffic.class);
```

would become
```groovy
Map<String, Traffic> map = EnumUtils.getEnumMap(Traffic.class, Traffic::name);
```

and the resulting map would be identical.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/EnumUtils.java
+++ b/src/main/java/org/apache/commons/lang3/EnumUtils.java
@@ -23,6 +23,7 @@ import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 /**
  * <p>Utility library to provide helper methods for Java enums.</p>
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -294,6 +295,25 @@
...
+     * <p>Gets the {@code Map} of enums.</p>
+     *
+     * <p>This method is useful when you need a map of enums.</p>
+     *
+     * @param <E> the type of enumeration
+     * @param <K> the type of the map key
+     * @param enumClass the class of the enum to query, not null
+     * @param keyFunction the function to query for the key, not null
+     * @return the modifiable map of enums, never null
+     */
```
garydgregory, line 308: Hi @MaxwellCody and thank you for your PR. Some comments:

- Add `@since 3.13.0`
- "Eat own dog food" and reimplement `EnumUtils.getEnumMap(Class<E>)` to use this method.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -294,6 +295,25 @@
...
+    public static <E extends Enum<E>, K> Map<K, E> getEnumMap(final Class<E> enumClass, final Function<E, K> keyFunction) {
+        final Map<K, E> map = new LinkedHashMap<>();
```
garydgregory, line 310: Add `Objects.requireNonNull(enumClass, "enumClass")` as the 1st line of this method.

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/EnumUtils.java
aherbert: Perhaps optimise the size of the hashmap:
```java
final E[] elements = enumClass.getEnumConstants();
final Map<K, E> map = new LinkedHashMap<>(elements.length);
for (final E e : elements) {
```

garydgregory:
Hi @MaxwellCody 
May you please rebase on master and address the comments?
TY!

garydgregory:
I pushed a different implementation to master.

PR #731, opened 2021-03-17
LANG-1649 - Check if method is Overridable

arturobernalg:
Create a method that Check if the given  Method is overridable in the given a  Class

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -983,6 +983,59 @@ public class ClassUtils {
     }
 
     /**
+     * Check if the given {@code Method} is overridable in the given target {@code Class}.
+     *
+     * <p>This method returns {@code false} for any {@code null} input or method with the static or final modifiers.</p>
+     * <pre>
+     * <code>
+     *   private static class ABG {
+     *        private void privateMethod() {
+     *        }
+     *        void defaultMethod() {
+     *        }
+     *        public void publicMethod() {
+     *        }
+     *        public final void finalMethod() {
+     *        }
+     *        public static void  staticMethod() {
+     *        }
+     *    }
+     *   Method publicMethod  = ABG.class.getDeclaredMethod("publicMethod", null)
+     *   Method privateMethod = ABG.class.getDeclaredMethod("privateMethod", null)
+     *   Method defaultMethod = ABG.class.getDeclaredMethod("defaultMethod", null)
+     * </code>
+     *
+     * ClassUtils.isOverridable(null, null)                = false
+     * ClassUtils.isOverridable(null, Class)               = false
+     * ClassUtils.isOverridable(Class, null)               = false
+     * ClassUtils.isOverridable(privateMethod, ABG.class)) = false
+     * ClassUtils.isOverridable(defaultMethod, ABG.class)) = true
+     * ClassUtils.isOverridable(publicMethod, ABG.class))  = true
+     * ClassUtils.isOverridable(finalMethod, ABG.class))   = false
+     * ClassUtils.isOverridable(staticMethod, ABG.class))  = false
+     * </pre>
+     *
+     * @param method the method to check, may be {@code null}
+     * @param cls the target class to check against, may be {@code null}
+     * @return {@code true} if the is overridable, false if not or any {@code null} input
+     * @since 3.13.0
+     */
+    public static boolean isOverridable(final Method method, final Class<?> cls) {
+        if (ObjectUtils.anyNull(method, cls)) {
+            return false;
+        }
+        final int mod = method.getModifiers();
+        if ((mod & (Modifier.STATIC | Modifier.FINAL | Modifier.PRIVATE)) != 0) {
+            return false;
+        }
+        if ((mod & (Modifier.PUBLIC | Modifier.PROTECTED)) != 0) {
+            return true;
+        }
+        // default scope: the package must be the same to allow override
+        return getPackageName(method.getDeclaringClass()).equals(getPackageName(cls));
+    }
+
+    /**
      * <p>
      * Gets the class name minus the package name from a {@code Class}.
      * </p>
--- a/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
@@ -39,6 +39,7 @@ import java.util.Set;
 import java.util.TreeMap;
 
 import org.apache.commons.lang3.ClassUtils.Interfaces;
+import org.apache.commons.lang3.reflect.testbed.ABG2;
 import org.apache.commons.lang3.reflect.testbed.GenericConsumer;
 import org.apache.commons.lang3.reflect.testbed.GenericParent;
 import org.apache.commons.lang3.reflect.testbed.StringParameterizedChild;
@@ -89,6 +90,15 @@ public class ClassUtilsTest  {
             // empty
         }
     }
+
+    private static class ABG extends ABG2 {
+        public void publicMethod() {}
+        private void privateMethod() {}
+        void defaultMethod() {}
+        public final void finalMethod() {}
+        public static void staticMethod() {}
+    }
+
 
     private void assertGetClassReturnsClass( final Class<?> c ) throws Exception {
         assertEquals( c, ClassUtils.getClass( c.getName() ) );
@@ -1126,6 +1136,27 @@ public class ClassUtilsTest  {
         assertTrue(ClassUtils.isAssignable(Boolean.TYPE, Boolean.TYPE), "boolean -> boolean");
     }
 
+    /**
+     * Test for {@link ClassUtils#isOverridable(Method, Class)}.
+     */
+    @Test
+    public void test_isOverridable() throws Exception {
+        final Method publicMethod = ABG.class.getDeclaredMethod("publicMethod", null);
+        assertTrue(ClassUtils.isOverridable(publicMethod, Object.class));
+        assertFalse(ClassUtils.isOverridable(null, Object.class));
+        final Method privateMethod = ABG.class.getDeclaredMethod("privateMethod", null);
+        assertFalse(ClassUtils.isOverridable(privateMethod, ABG.class));
+        final Method defaultMethod = ABG.class.getDeclaredMethod("defaultMethod", null);
+        assertTrue(ClassUtils.isOverridable(defaultMethod, ABG.class));
+        final Method finalMethod = ABG.class.getDeclaredMethod("finalMethod", null);
+        assertFalse(ClassUtils.isOverridable(finalMethod, ABG.class));
+        final Method staticMethod = ABG.class.getDeclaredMethod("staticMethod", null);
+        assertFalse(ClassUtils.isOverridable(staticMethod, ABG.class));
+        final Method defaultMethodABG2 = ABG2.class.getDeclaredMethod("defaultMethodABG2", null);
+        assertFalse(ClassUtils.isOverridable(defaultMethodABG2, ABG.class));
+    }
+
+    // -------------------------------------------------------------------------
     @Test
     public void test_isInnerClass_Class() {
         assertTrue(ClassUtils.isInnerClass(Inner.class));
--- a/src/test/java/org/apache/commons/lang3/reflect/testbed/ABG2.java
+++ b/src/test/java/org/apache/commons/lang3/reflect/testbed/ABG2.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.lang3.reflect.testbed;
+
+import org.apache.commons.lang3.ClassUtilsTest;
+
+/**
+ * Test class for {@link ClassUtilsTest#test_isOverridable()}}.
+ */
+public class ABG2 {
+    void defaultMethodABG2() {}
+}


aherbert:
There are more rules to check here. You should also check the method is not **final** using `Modifier.isFinal` which prevents being overridden.

How to handle static methods is an issue. You cannot create an **instance** method in a subclass with the same signature as a **static** method from the super class. But you can create a **static** method with the same signature. The static methods are not dispatched on an object at runtime. The method to use is chosen by the compiler.

So perhaps this should return false for any method with the static or final modifiers.

arturobernalg:
> There are more rules to check here. You should also check the method is not **final** using `Modifier.isFinal` which prevents being overridden.
> 
> How to handle static methods is an issue. You cannot create an **instance** method in a subclass with the same signature as a **static** method from the super class. But you can create a **static** method with the same signature. The static methods are not dispatched on an object at runtime. The method to use is chosen by the compiler.
> 
> So perhaps this should return false for any method with the static or final modifiers.

Hi @aherbert 
I completely forgot about the final. I will check the static case.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -88,6 +88,14 @@
             // empty
         }
     }
+    private static class ABG {
```
aherbert, line 91: Add an extra line

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -88,6 +88,14 @@
...
+        public void publicMethod(){}
+        private void privateMethod(){}
+        void defaultMethod(){}
+        public final void finalMethod(){}
+        public static void  staticMethod(){}
```
aherbert, line 96: Remove double whitespace after void. The same change in the javadoc of the method in ClassUtils.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -1131,6 +1139,25 @@ public void test_isAssignable_Widening() {
...
+        assertFalse(ClassUtils.isOverridable(null, Object.class));
+        final Method privateMethod = ABG.class.getDeclaredMethod("privateMethod", null);
+        assertFalse(ClassUtils.isOverridable(privateMethod, ABG.class));
+        final Method defaultMethod = ABG.class.getDeclaredMethod("defaultMethod", null);
+        assertTrue(ClassUtils.isOverridable(defaultMethod, ABG.class));
+        final Method finalMethod = ABG.class.getDeclaredMethod("finalMethod", null);
+        assertFalse(ClassUtils.isOverridable(finalMethod, ABG.class));
+        final Method staticMethod = ABG.class.getDeclaredMethod("staticMethod", null);
+        assertFalse(ClassUtils.isOverridable(staticMethod, ABG.class));
+
```
aherbert, line 1158: Remove extra line here

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -935,6 +935,60 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
         return toClass.isAssignableFrom(cls);
     }
 
+    /**
+     * <p>Check if the given {@code Method} is overridable in the given target {@code Class}.</p>
```
aherbert, line 939: No `<p>` tags on the first paragraph

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -935,6 +935,60 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
...
+     *
+     * <p>This method returns {@code false} for any {@code null} input or method with the static or final modifiers.</p>
+     * <pre>
+     * <code>
+     *   private static class ABG {
+     *        private void privateMethod(){
```
aherbert, line 945: Add whitespace before the curly brace `{` (to match our coding style)

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
arturobernalg: Done

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
arturobernalg: Done

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
arturobernalg: Done

src/main/java/org/apache/commons/lang3/ClassUtils.java
arturobernalg: Done

src/main/java/org/apache/commons/lang3/ClassUtils.java
arturobernalg: Done

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -935,6 +935,60 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
...
+     * @param method the method to check, may be {@code null}
+     * @param cls the target class to check against, may be {@code null}
+     * @return {@code true} if the is overridable, false if not or any {@code null} input
+     * @since 3.13.0
+     */
+    public static boolean isOverridable(final Method method, final Class<?> cls) {
+        if (ObjectUtils.anyNull(method, cls)) {
+            return false;
+        }
+        if (Modifier.isStatic(method.getModifiers())) {
```
aherbert, line 980: Given that `Modifiers` exposes the bit flag constants can this all be done with 2 comparisons:
```java
final int mod = method.getModifiers();
if ((mod & (Modifier.STATIC | Modifier.FINAL | Modifier.PRIVATE) != 0) {
   return false;
}
if ((mod & (Modifier.PUBLIC | Modifier.PROTECTED) != 0) {
   return true;
}
// default scope: the package must be the same to allow override
return getPackageName(method.getDeclaringClass()).equals(getPackageName(cls));
```

arturobernalg: & cannot be applied to int

aherbert: Using `&` here is a bitwise AND operation (see [Bitwise Operators](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/op3.html)). Look at the source code for Modifier.isStatic(...) for an example.

Did you try the code? It has a missing parenthesis so this should be correct:
```java
final int mod = method.getModifiers();
if ((mod & (Modifier.STATIC | Modifier.FINAL | Modifier.PRIVATE)) != 0) {
   return false;
}
if ((mod & (Modifier.PUBLIC | Modifier.PROTECTED)) != 0) {
   return true;
}
// default scope: the package must be the same to allow override
return getPackageName(method.getDeclaringClass()).equals(getPackageName(cls));
```

arturobernalg:
HI @aherbert 
Hi, thank you for the test case tip. Regarding with use bitwise operator IMO it's more complex
TY

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -1131,6 +1140,24 @@ public void test_isAssignable_Widening() {
...
+     * Test for {@link ClassUtils#isOverridable(Method, Class)}.
+     */
+    @Test
+    public void test_isOverridable() throws Exception {
+        final Method publicMethod = ABG.class.getDeclaredMethod("publicMethod", null);
+        assertTrue(ClassUtils.isOverridable(publicMethod, Object.class));
+        assertFalse(ClassUtils.isOverridable(null, Object.class));
+        final Method privateMethod = ABG.class.getDeclaredMethod("privateMethod", null);
+        assertFalse(ClassUtils.isOverridable(privateMethod, ABG.class));
+        final Method defaultMethod = ABG.class.getDeclaredMethod("defaultMethod", null);
```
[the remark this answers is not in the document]
arturobernalg: Done

src/main/java/org/apache/commons/lang3/ClassUtils.java
arturobernalg: Done

arturobernalg:
hi @aherbert 
Do i have to do something else?
ty

src/test/java/org/apache/commons/lang3/reflect/testbed/ABG2.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.lang3.reflect.testbed;
+
+import org.apache.commons.lang3.ClassUtilsTest;
+
+/**
+ * Test class for {@link ClassUtilsTest#test_isOverridable()}}.
+ */
+public class ABG2 {
```
aherbert, line 24: Looking at the rest of the testbed package, this does not have a very descriptive name. I would suggest `ClassWithPackagePrivateMethod` and the method to be `packagePrivateMethod`. The method is not a default method (which applies to interfaces). The purpose of the testbed class is then self-documenting.

The test class that extends this can then be named whatever you want but I would suggest `IsOverridableClass`.

src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -90,6 +91,15 @@ private class DeeplyNested {
         }
     }
 
+    private static class ABG extends ABG2 {
+        public void publicMethod() {}
+        private void privateMethod() {}
+        void defaultMethod() {}
```
aherbert, line 97: Can you put all the method modifiers in the method name:

```Java
void packagePrivateMethod() {}
public final void publicFinalMethod() {}
public static void publicStaticMethod() {}
```

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -982,6 +982,59 @@ public static String getShortCanonicalName(final String canonicalName) {
...
+    /**
+     * Check if the given {@code Method} is overridable in the given target {@code Class}.
+     *
+     * <p>This method returns {@code false} for any {@code null} input or method with the static or final modifiers.</p>
+     * <pre>
+     * <code>
+     *   private static class ABG {
+     *        private void privateMethod() {
+     *        }
+     *        void defaultMethod() {
```
aherbert, line 994: Update this to `packagePrivateMethod`

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -982,6 +982,59 @@ public static String getShortCanonicalName(final String canonicalName) {
...
+     *        public void publicMethod() {
+     *        }
+     *        public final void finalMethod() {
+     *        }
+     *        public static void  staticMethod() {
+     *        }
+     *    }
+     *   Method publicMethod  = ABG.class.getDeclaredMethod("publicMethod", null)
+     *   Method privateMethod = ABG.class.getDeclaredMethod("privateMethod", null)
+     *   Method defaultMethod = ABG.class.getDeclaredMethod("defaultMethod", null)
```
aherbert, line 1005: Update this to `packagePrivateMethod`

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -982,6 +982,59 @@ public static String getShortCanonicalName(final String canonicalName) {
...
+     * </code>
+     *
+     * ClassUtils.isOverridable(null, null)                = false
+     * ClassUtils.isOverridable(null, Class)               = false
+     * ClassUtils.isOverridable(Class, null)               = false
+     * ClassUtils.isOverridable(privateMethod, ABG.class)) = false
+     * ClassUtils.isOverridable(defaultMethod, ABG.class)) = true
```
aherbert, line 1012: Update this to `packagePrivateMethod`

The limitations of the result for packagePrivateMethod should be documented. For example:

```
<p>Note that the result of {@code isOverridable} when called on a package-private method is
sensitive to the package of the target class. When invoked on the declaring class or a child
class within the same package the result will be {@code true} as the method is visible within 
the package. When invoked on a child class within a different package the result will be
{@code false}.
```

src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -982,6 +982,59 @@ public static String getShortCanonicalName(final String canonicalName) {
...
+            return false;
+        }
+        final int mod = method.getModifiers();
+        if ((mod & (Modifier.STATIC | Modifier.FINAL | Modifier.PRIVATE)) != 0) {
+            return false;
+        }
+        if ((mod & (Modifier.PUBLIC | Modifier.PROTECTED)) != 0) {
+            return true;
+        }
+        // default scope: the package must be the same to allow override
```
aherbert, line 1034: package-private scope:

garydgregory:
What is the use case? I do not mean "what does the code do". Are you writing or maintaining Java tools lile Maven plugins or an IDE or something like that? Or, is this PR a "this is just something I think is cool and maybe someone will use it"?

arturobernalg:
> What is the use case? I do not mean "what does the code do". Are you writing or maintaining Java tools lile Maven plugins or an IDE or something like that? Or, is this PR a "this is just something I think is cool and maybe someone will use it"?

It is something that in my opinion is interesting to have in a ClassUtils. We have a bunch of method in the class that help to any developer when they need manipulate some class. IMO this function might help.

garydgregory:
Hi @arturobernalg 
"IMO this function might help." Or it might not, or it might if it took some unknown additional parameter. I am -1 to add code that "might" help; IOW, unless there is an actual practical need for functionality, let's not add it just because we can, this would lead to endless bloat.

arturobernalg:
Hi @garydgregory 
I close the PR. but it is not clear to me who says or under what guidelines it is "real practical need for functionality". I thought that the whole idea of the "commons" library was to have several functions to help developers in their daily tasks.

The apache commons libraries allow me know if a class is public or if is a inner class, in fact, all this options can i use

```
comparator
convertClassesToClassNames
convertClassNamesToClasses
getAbbreviatedName
getAbbreviatedName
getAllInterfaces
getAllInterfaces
getAllSuperclasses
getCanonicalName
getCanonicalName
getCanonicalName
getCanonicalName
getCanonicalName
getClass
getClass
getClass
getClass
getComponentType
getName
getName
getName
getName
getPackageCanonicalName
getPackageCanonicalName
getPackageCanonicalName
getPackageName
getPackageName
getPackageName
getPublicMethod
getShortCanonicalName
getShortCanonicalName
getShortCanonicalName
getShortClassName
getShortClassName
getShortClassName
getSimpleName
getSimpleName
getSimpleName
getSimpleName
hierarchy
hierarchy
isAssignable
isAssignable
isAssignable
isAssignable
isInnerClass
isPublic
isPrimitiveOrWrapper
isPrimitiveWrapper
primitivesToWrappers
primitiveToWrapper
toCanonicalName
toClass
useFull
wrappersToPrimitives
wrapperToPrimitive
```

but what it not a "practical need for functionality" it's if a method is Overridable. 

I don't know.

PR #733, opened 2021-03-20
Fixed NPE getting Stack Trace if Throwable is null

arturobernalg:


src/test/java/org/apache/commons/lang3/exception/ContextedExceptionTest.java
```diff
@@ -48,6 +48,12 @@ public void testContextedException() {
         assertTrue(StringUtils.isEmpty(message));
     }
 
+    @Test
+    public void testNullException() {
+        final String trace = ExceptionUtils.getStackTrace(null);
```
garydgregory, line 53: In-line `trace`.

arturobernalg: Hi @garydgregory 
Changed
TY

src/test/java/org/apache/commons/lang3/exception/ContextedExceptionTest.java
```diff
@@ -48,6 +48,11 @@ public void testContextedException() {
         assertTrue(StringUtils.isEmpty(message));
     }
 
+    @Test
+    public void testNullException() {
+        assertTrue(StringUtils.isEmpty(ExceptionUtils.getStackTrace(null)));
```
garydgregory, line 53: I think the test can be simplified by not using StringUtils.

arturobernalg: HI @garydgregory 
SomeThins like --> `assertNotNull(ExceptionUtils.getStackTrace(null));` ??
TY

garydgregory: > 
> 
> HI @garydgregory
> SomeThins like --> `assertNotNull(ExceptionUtils.getStackTrace(null));` ??
> TY

Hello @arturobernalg 
No, assert what you expect to happen on null input. Your suggestion above would accept a return value of "foo". Is that OK? I think not.

arturobernalg: HI @garydgregory 
ok i think i understand what you mean.
TY

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
```diff
@@ -359,14 +359,16 @@ public static String getRootCauseMessage(final Throwable th) {
...
-     *  {@code printStackTrace(PrintWriter)} method
+     * {@code printStackTrace(PrintWriter)} method, or an empty String if {@code null} input
      */
     public static String getStackTrace(final Throwable throwable) {
         final StringWriter sw = new StringWriter();
         final PrintWriter pw = new PrintWriter(sw, true);
-        throwable.printStackTrace(pw);
+        if (throwable != null) {
+            throwable.printStackTrace(pw);
+        }
```
garydgregory, line 371: We do not need to create the writers at all. IMO, just guard the method by starting it with:
```
        if (throwable == null) {
            return StringUtils.EMPTY;
        }
```

garydgregory (changes requested):
@arturobernalg 
See my 1 comment and rebase on master, please.

garydgregory (changes requested):
See https://github.com/apache/commons-lang/pull/733/files#r820237567

src/main/java/org/apache/commons/lang3/exception/ExceptionUtils.java
garydgregory: ping.

arturobernalg: HI @garydgregory 
I don't see the way to just use String and not break the binary.

PR #736, opened 2021-03-22
LANG-1652 - Check whether the given class has a public constructor

arturobernalg:
Determine whether the given class has a public constructor with the given signature, checking accessibility.


garydgregory:
> > @arturobernalg
> > I am curious: Can this be used within Commons Lang itself?
> 
> Hi @garydgregory We have a complete class dedicated to manipulate and access to the method of a class through reflections, so, IMO why not.

Not, because (1) It's not useable within Commons Lang itself, because current callers of getAccessibleConstructor(*) test and use the result of the call. So using this new API internally would cause getAccessibleConstructor() to be called more than once instead of once. And (2) because "why not" is not a use-case, so YAGNI.

PR #743, opened 2021-04-14
Lang 1656: first non blank returned

xuxiao415:
add new feature for first non blank returned by supplier array


arturobernalg:
Shouldn't these functions be in the `ObjectUtils` class?

xuxiao415:
> Shouldn't these functions be in the `ObjectUtils` class?

because the type of value returned by function is subclass of `CharSequence`

arturobernalg:
> > Shouldn't these functions be in the `ObjectUtils` class?
> 
> because the type of value returned by function is subclass of `CharSequence`

Sounds fair.

PR #745, opened 2021-04-16
Lang 1656 first non blank returned

xuxiao415:


garydgregory:
Closing: Too narrow of a use case I think.

PR #751, opened 2021-05-09
create API for searching for numbers within string - solving LANG-165…

Mikhail-Polivakha:
Solving Jira LANG-1658 issue. Unit test with multiple assertions is included.
Jira issue reference: https://issues.apache.org/jira/browse/LANG-1658

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -9639,6 +9639,79 @@ public class StringUtils {
     }
 
     /**
+     * Method that assembles all the numbers, form the passed string and returns them as list.
+     * It is important to note here, is that bu 'number' method assume any digit sequence, that
+     * can (but not necessary at all) contains dot within it (I mean just plain old floats,
+     * something like 51.82)
+     *
+     * For example, you may pass a string "21.2 days 3 minutes 22 seconds". For this particular string
+     * the result list of doubles will look like this : [21.2, 3.0, 22.0]
+     *
+     * if string contains invalid numbers (for example this string contains
+     * not valid number: "My height is 1234.23.13" This is invalid because it
+     * is not clear how to parse this part - 1234.23.13), {@link NumberFormatException}
+     * will be thrown. Though if string will contain number, where right
+     * after second dot resides not a number, or, any other char, then this
+     * case will be considered as valid. For example, this string contains
+     * only valid numbers: "My pulse is 90.123. and weight is 78.2"
+     * In this case sequence "90.123." will be considered as "90.123", as well as
+     * sequence "90." (imagine that there is no digit right after dot) will be
+     * considered as 90.0 double.
+     *
+     * @param stringThatContainsNumbers - string, that contains number or several numbers.
+     *                                 Not necessary integers, may be numbers with float point.
+     * @return - list of numbers, that this particular string contains
+     *
+     * @throws NumberFormatException - see documentation clarification about cases when thrown above
+     */
+    public static List<Double> extractNumbersFromString(String stringThatContainsNumbers) {
+        boolean hasDigitAlreadyStarted = false;
+        boolean alreadyMetDotInThisNumber = false;
+
+        List<Double> resultList = new ArrayList<>();
+
+        StringBuilder currentNumberAsStringBuilder = new StringBuilder("");
+
+        for (int i = 0; i < stringThatContainsNumbers.length(); i++) {
+            char currentSymbol = stringThatContainsNumbers.charAt(i);
+            if (Character.isDigit(currentSymbol)) {
+                if (!hasDigitAlreadyStarted) {
+                    hasDigitAlreadyStarted = true;
+                }
+                currentNumberAsStringBuilder.append(currentSymbol);
+                continue;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -9638,6 +9638,79 @@ public static String wrapIfMissing(final String str, final String wrapWith) {
...
+
+        for (int i = 0; i < stringThatContainsNumbers.length(); i++) {
+            char currentSymbol = stringThatContainsNumbers.charAt(i);
+            if (Character.isDigit(currentSymbol)) {
+                if (!hasDigitAlreadyStarted) {
+                    hasDigitAlreadyStarted = true;
+                }
+                currentNumberAsStringBuilder.append(currentSymbol);
+                continue;
+            } else if (currentSymbol == '.') {
```
garydgregory, line 9683: This will not work on locales that use a different character for the decimal marker.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -9638,6 +9638,79 @@ public static String wrapIfMissing(final String str, final String wrapWith) {
...
+     * sequence "90." (imagine that there is no digit right after dot) will be
+     * considered as 90.0 double.
+     *
+     * @param stringThatContainsNumbers - string, that contains number or several numbers.
+     *                                 Not necessary integers, may be numbers with float point.
+     * @return - list of numbers, that this particular string contains
+     *
+     * @throws NumberFormatException - see documentation clarification about cases when thrown above
+     */
+    public static List<Double> extractNumbersFromString(String stringThatContainsNumbers) {
```
garydgregory, line 9667: The API name says Number but the return type says Double. They should match.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -3338,6 +3341,95 @@ public void testToRootUpperCase() {
         }
     }
 
+    @Test
+    void extractNumbersFromString() {
```
garydgregory, line 3345: By convention test method have the "test" prefix.

You should break up each use case in their own method IMO.

There are no tests for edge cases: null and empty string. Also a non empty string with no numbers.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -3338,6 +3341,95 @@ public void testToRootUpperCase() {
...
+        expectedList.add(299.0);
+        Assertions.assertThrows(NumberFormatException.class, () -> StringUtils.extractNumbersFromString("737 another is 42.23.32"));
+        expectedList.clear();
+
+        expectedList.add(737.0);
+        expectedList.add(29.1);
+        expectedList.add(299.0);
+        input = "737. something29.1. 299";
+        Assertions.assertEquals(expectedList, StringUtils.extractNumbersFromString(input));
+        expectedList.clear();
```
garydgregory, line 3405: Why is this an error condition? Sentences end in a period.

How would you treat commas in a sentence when in a locale like French that uses the comma as the decimal separator?

garydgregory (changes requested):
Also what about the thousands separator as in "10,000" for 10 thousand? What about the Unicode decimal separator? A true general solution is much more complex than one that focuses on a narrower case hard coded to the period decimal separator. 

The type of string parsing is too specific for Commons Lang IMO, maybe for Commons Text.

I'm curious as to what the community thinks.

src/main/java/org/apache/commons/lang3/StringUtils.java
Mikhail-Polivakha: You propose to consume separator as method parameter?

src/main/java/org/apache/commons/lang3/StringUtils.java
Mikhail-Polivakha: So, ok, and what is exactly your proposition?) Btw all the types you mentioned extends Number class, that`s why from my point of view it is valid. Explain please more to me why it does not work

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
Mikhail-Polivakha: Valid points, I will add some. Thanks!

src/main/java/org/apache/commons/lang3/StringUtils.java
garydgregory: You can implement an API to do that if you want but my main point is that the code should run with inputs from any locale which means handling period and commas as both the decimal and thousands separators. But, the other point I was attempting to make is that I do not feel this code belongs in Commons Lang, it feels too much like NLP code to me. It might be something for Commons Text, but it seems quite a specific use case, too much like NLP, not generic enough for a Commons library. I encourage others in the community to opine. The NLP nature makes me wonder how you would handle input like "I 'd like 10,000 nuts, 9,000 bolts, but only up to $10.5." and "Send $4.50.5 apples too please."

garydgregory:
Closing: No reply, and this kind of text processing does not belong in Commons Lang but could be in Commons Text.

PR #758, opened 2021-05-16
Use own validator ClassUtils.getSimpleName to get String input

arturobernalg:


garydgregory:
@arturobernalg Would you rebase on master please?

arturobernalg:
> @arturobernalg Would you rebase on master, please?

done
TY

src/main/java/org/apache/commons/lang3/time/FastDateParser.java
```diff
@@ -564,7 +565,7 @@ boolean parse(final FastDateParser parser, final Calendar calendar, final String
          */
         @Override
         public String toString() {
-            return getClass().getSimpleName() + " [pattern=" + pattern + "]";
+            return ClassUtils.getSimpleName(this) + " [pattern=" + pattern + "]";
```
garydgregory, line 568: We should not do this since a "this" will never be null.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -67,7 +67,7 @@
     private static <E extends Enum<E>> Class<E> checkBitVectorable(final Class<E> enumClass) {
         final E[] constants = asEnum(enumClass).getEnumConstants();
         Validate.isTrue(constants.length <= Long.SIZE, CANNOT_STORE_S_S_VALUES_IN_S_BITS,
-            Integer.valueOf(constants.length), enumClass.getSimpleName(), Integer.valueOf(Long.SIZE));
+            Integer.valueOf(constants.length), ClassUtils.getSimpleName(enumClass), Integer.valueOf(Long.SIZE));
```
garydgregory, line 70: This does not make sense since a null enumClass will cause an exception before we even get here.

garydgregory (changes requested):
Please review all call sites keeping in mind the 2 comments I left in this PR. In general, it does not make sense to call the util method when the argument will never be null.

garydgregory (changes requested):
Hi @arturobernalg Ping?

PR #760, opened 2021-05-16
ObjectUtils: Add nullSafeEquals method for Object compare. 

LeeKyoungIl:
When we compare Object values, if it enters more than 2depth, it may become a bit annoying.

For example

```java
class Bar {
    String testBarString = "test";
}

class Foo {
    Bar bar = new Bar();
}

Foo foo = new Foo();

if ( foo.getBar() != null && "test".equals(foo.getBar().getTestBarString()) ) {
    // TODO
}
```

So I made a feature that can be compared at once and is null safe.

```java
// AS-IS
if ( foo.getBar() != null && "test".equals(foo.getBar().getTestBarString()) ) {
    // TODO
}

// TO-BE
if ( ObjectUtils.nullSafeEquals(foo, "bar.testBarString", "test') ) {
    // TODO
}
```

I made it because it didn't seem to have a similar function like this before.
Do you have a similar method in ObjectUtils?

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -755,7 +755,8 @@ public class ObjectUtils {
      * @param targetPath object target path to be checked.
      * @param value value to compare.
      * @return {@code true} if the objects are same, {@code false} otherwise
-     * @throws IllegalAccessException, IllegalArgumentException if an Access or Argument error occurs.
+     * @throws IllegalAccessException if an Access error occurs.
+     * @throws IllegalArgumentException if an Argument error occurs.
      */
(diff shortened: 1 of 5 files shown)


garydgregory:
Hi @LeeKyoungIl 
This seems redundant with https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java

LeeKyoungIl:
@garydgregory 
Hi! 

In my personal opinion, reflectionEquals is uncomfortable to use, but you can close it if you think it is similar.
Thanks. !

garydgregory:
Hi @LeeKyoungIl Going back to basics: If `Foo` and `Bar` implement `equals()`/`hashCode()`, you don't need any of this. What am I missing? You cannot change `Foo` and `Bar`? If you cannot, then `EqualsBuilder` is your friend.

garydgregory:
Closing, no more feedback from OP.

PR #762, opened 2021-05-19
LANG-1661 - Add methods to ClassUtils to get the class type names in a null-safe manner

arturobernalg:
Add methods ClassUtils to get the class names type in a null-safe manner:

- getTypeName(Class)
- getTypeName(Object)
- getTypeName(Class, String)
- getTypeName(Object, String)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -346,6 +346,56 @@ public class ClassUtils {
     }
 
     /**
+     * <p>Null-safe version of {@code cls.getTypeName()}</p>
+     *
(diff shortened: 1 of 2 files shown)


garydgregory:
Can this be used within Commons Lang? 
Also, would you rebase on master, please?

arturobernalg:
> Can this be used within Commons Lang?
Why not. There are commons class utils operations.
> Also, would you rebase on master, please?
Rebased.

arturobernalg:
> Can this be used within Commons Lang?
> Also, would you rebase on master, please?

HI @garydgregory 
Maybe not in Commons Lang itself, but definitely its a utility that should be in ClassUtils be used by end users.

PR #765, opened 2021-05-29
Check if the given class represents an array of primitives

arturobernalg:
null-safe

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -791,6 +791,24 @@ public class ClassUtils {
     }
 
     /**
+     * <p>Returns whether the given {@code type} is a primitive array class ({@code boolean}, {@code byte}, {@code char},
+     * {@code short}, {@code int}, {@code long}, {@code double}, {@code float}).</p>
+     *
+     * <p>The method returns {@code false} if a {@code null} type is passed in.</p>
+     *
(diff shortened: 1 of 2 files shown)


garydgregory:
Can this be used within Commons Lanf itself?

arturobernalg:
> Can this be used within Commons Lanf itself?

HI @garydgregory 
i.e in -->  [ObjectUtil](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/ObjectUtils.java#L249)

garydgregory:
> 
> 
> > Can this be used within Commons Lanf itself?
> 
> HI @garydgregory
> i.e in --> [ObjectUtil](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/ObjectUtils.java#L249)

That does not seem like a very good example because it reuses the component type just 2 lines below.

arturobernalg:
> > > Can this be used within Commons Lanf itself?
> > 
> > 
> > HI @garydgregory
> > i.e in --> [ObjectUtil](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/ObjectUtils.java#L249)
> 
> That does not seem like a very good example because it reuses the component type just 2 lines below.
HI @garydgregory 
That's true. Maybe it's not the best example, but proves that the functionality it's used and the function `isPrimitiveArray`  could be used
TY

PR #766, opened 2021-05-29
LANG-1172: Support dash as a delimiter in locales

c-w:
This pull request is a partial fix for [LANG-1172](https://issues.apache.org/jira/browse/LANG-1172) adding support for dashes as delimiters for locales as per [BCP47](https://tools.ietf.org/html/bcp47#section-4) thus enabling parsing of locales such as `en-GB` in addition to the previously supported (but not to spec) `en_GB`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/LocaleUtils.java
+++ b/src/main/java/org/apache/commons/lang3/LocaleUtils.java
@@ -250,9 +250,9 @@ public class LocaleUtils {
             return new Locale(str);
         }
 
-        final String[] segments = str.indexOf(DASH) != -1
-            ? str.split(String.valueOf(DASH), -1)
-            : str.split(String.valueOf(UNDERSCORE), -1);
+        final String[] segments = str.indexOf(UNDERSCORE) != -1
+            ? str.split(String.valueOf(UNDERSCORE), -1)
+            : str.split(String.valueOf(DASH), -1);
         final String language = segments[0];
         if (segments.length == 2) {
             final String country = segments[1];
--- a/src/main/java/org/apache/commons/lang3/LocaleUtils.java
+++ b/src/main/java/org/apache/commons/lang3/LocaleUtils.java
@@ -248,7 +248,9 @@ public class LocaleUtils {
             return new Locale(str);
         }
 
-        final String[] segments = str.split("_", -1);
+        final String[] segments = str.contains("-")
+            ? str.split("-", -1)
+            : str.split("_", -1);
         final String language = segments[0];
         if (segments.length == 2) {
             final String country = segments[1];
@@ -289,6 +291,7 @@ public class LocaleUtils {
      *   LocaleUtils.toLocale("")           = new Locale("", "")
      *   LocaleUtils.toLocale("en")         = new Locale("en", "")
      *   LocaleUtils.toLocale("en_GB")      = new Locale("en", "GB")
+     *   LocaleUtils.toLocale("en-GB")      = new Locale("en", "GB")
      *   LocaleUtils.toLocale("en_001")     = new Locale("en", "001")
      *   LocaleUtils.toLocale("en_GB_xxx")  = new Locale("en", "GB", "xxx")   (#)
      * </pre>
@@ -300,7 +303,7 @@ public class LocaleUtils {
      * <p>This method validates the input strictly.
      * The language code must be lowercase.
      * The country code must be uppercase.
-     * The separator must be an underscore.
+     * The separator must be an underscore or a dash.
      * The length must be correct.
      * </p>
      *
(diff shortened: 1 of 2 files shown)


garydgregory:
@c-w 
May you please rebase on master?
TY!

c-w:
@garydgregory Just following up on this. Is there anything further I can do to speed up getting this reviewed? Thanks in advance for your time!

c-w:
@garydgregory Do you have feedback on this pull request? Anything holding this back from getting merged?

src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -248,7 +248,9 @@ private static Locale parseLocale(final String str) {
             return new Locale(str);
         }
 
-        final String[] segments = str.split("_", -1);
+        final String[] segments = str.contains("-")
+            ? str.split("-", -1)
```
sgoeschl, line 252: Blame my weak eyes - could we introduce constants here, e.g. "UNDERSCORE"?

src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -324,7 +327,7 @@ public static Locale toLocale(final String str) {
             throw new IllegalArgumentException("Invalid locale format: " + str);
         }
         final char ch0 = str.charAt(0);
-        if (ch0 == '_') {
+        if (ch0 == '_' || ch0 == '-') {
```
sgoeschl, line 330: We could use "UNDERSCORE" or "DASH" here or introduce a helper method, e.g. "isDelimiter"? What do you think?

src/main/java/org/apache/commons/lang3/LocaleUtils.java
c-w: Done in 11a6d1d52.

src/main/java/org/apache/commons/lang3/LocaleUtils.java
c-w: Added the constant in 11a6d1d52. I'd prefer not to add the helper method to avoid changing too much of the structure as part of this fairly minor overall change.

kinow (commented):
I am not sure whether we should modify the current methods to support this locale, or have separate methods. But the code looks really good :+1: Maybe someone else can chime in and share their thoughts on supporting both here. Thanks for the PR @c-w !

c-w:
@kinow Given that the change is backwards compatible I'd say introducing a new method would be more confusing as the developer would have to know/remember to pick the correct function.

@garydgregory Any thoughts or other comments on the PR?

src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -248,7 +250,9 @@ private static Locale parseLocale(final String str) {
             return new Locale(str);
         }
 
-        final String[] segments = str.split("_", -1);
+        final String[] segments = str.indexOf(DASH) != -1
+            ? str.split(String.valueOf(DASH), -1)
```
garydgregory, line 254: Does it matter that dash is tested before underscore?

c-w: It's been a while since I wrote this code, but I don't think so. Should be possible to switch them around without issues.

c-w: All the unit tests still pass with the comparison swapped. I made the change in 0319d84d1. It's probably the right thing to do to optimize for the previous delimiter as that'll be the vast majority of use-cases.

c-w:
Another friendly ping @garydgregory @kinow @sgoeschl. Could you take another look and let me know if you have any concerns about this change? Thanks in advance!

c-w:
Thanks for the merge! :tada:

garydgregory:
You should build git master locally to make sure you can use the jar in your app.

seadbrane:
This change has resulted in some broken expectations in the differences between locale and language tags.  While there is probably value in having a method that is open to taking either in order to produce a Locale object, it is not clear that should be this method.  In fact, the original and current comments for the method say "This method takes the **string format of a locale** and creates the locale object from it".   In addition, the rationale for why this specific change was made appears to based on some inaccurate understanding - BCP 47 is the spec for language tags, not locales and BCP47 does not support 'underscore', which was this method only supported before this change.

seadbrane:
Any update?

garydgregory:
Hi @seadbrane 
Feel free to provide a PR.

seadbrane:
The PR is to revert this commit unless the intention has changed since the method was created.

On Feb 6, 2024, at 15:34, Gary Gregory ***@***.***> wrote:

﻿

Hi @seadbrane<https://github.com/seadbrane>
Feel free to provide a PR.

seadbrane:
pr to mostly revert this change.  

https://github.com/apache/commons-lang/pull/1271

PR #773, opened 2021-06-20
Object Casting

johnson-abraham:
Casts an object from one type to the specified type in a null-safe manner.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -229,6 +229,21 @@ public class ObjectUtils {
     }
 
     /**
+     * Casts the object to the targeted class.
+     * @param value the object to cast
+     * @param castClass the class to which the object must be cast to
+     * @param <T> the type of the cast object
+     * @return the cast object, otherwise {@code null}
+     */
+    public static <T> T cast(Object value, Class<T> castClass) {
+        if ((castClass != null) && castClass.isInstance(value)) {
+            return castClass.cast(value);
+        }
+        return null;
+    }
+
+
+    /**
      * <p>Clone an object.</p>
      *
      * @param <T> the type of the object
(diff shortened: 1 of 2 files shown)


garydgregory:
There is a mismatch here between the code, the Javadoc, and the PR description. Where is the actual intent expressed? Specifically, the description and Javadoc do not state that class cast exceptions are effectively escaped into null returned values. This is quite unexpected since the PR and Javadoc only talk about the method as a null-safe shorthand utility. 

If the method is to only be a null-safe shorthand wrapper then 'cast' is fine as the method name, but if the internet is to add logic to return null values for illegal casts, then the name needs to be different in order to allow for a method that only does the null safe check but allows for CCE.

In the end, the real value of such a low-level shorthand method is better evaluated if Commons Lang itself is made to use it internally.  Are there sites within Lang that can make use of this utility?

I'll sprinkle comments in the code as well.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -228,6 +227,18 @@ public static boolean anyNull(final Object... values) {
         return !allNotNull(values);
     }
 
+    /**
+     * Casts the object to the targeted class.
+     * @param value the object to cast
+     * @param castClass the class to which the object must be cast to
+     * @param <T> the type of the cast object
+     * @return the cast object, otherwise {@code null}
+     */
```
garydgregory, line 236: New main public and protected methods need a Javadoc since tag.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -228,6 +227,18 @@ public static boolean anyNull(final Object... values) {
...
+    public static <T> T cast(Object value, Class<T> castClass) {
+        return (castClass != null) && castClass.isInstance(value) ? castClass.cast(value) : null;
```
garydgregory, line 238: Match the style of the file: Use final where possible, no need for extra parentheses.

garydgregory (changes requested):
@johnson-abraham 
See comments. Looking for comments from others in the community as well.

PR #777, opened 2021-07-10
LANG-1664 - Javadoc for StringUtils.substringBefore(String str, int separator) doesn't mention that the separator is an Int

amarlearning:
Ready for review

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8700,7 +8700,7 @@ public class StringUtils {
 
     /**
      * <p>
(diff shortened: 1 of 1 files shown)


oneiros-de:
This is wrong - the "separator" is not an index but a char used by `String#indexof(int)`.
See also the unit test `StringUtilsSubstringTest.testSubstringBefore_StringInt`

So the description of the method is mostly correct (the description of `separator` is wrong, should be "character (Unicode code point", not "String"); the examples simply don't match the code (the unit test has working examples).

An `@see String#indexOf(int)` would help.

And this should be done for similar functions.

amarlearning:
@oneiros-de sorry my mistake, I got confused. Closing my PR.

PR #780, opened 2021-07-18
LANG-1665 - Modify Tuple equals to take BigDecimal comparison

toyopilgrim:
LANG-1665

Perhaps as possibly ignorable suggestion, it'd be more helpful if the Tupple supports BgDecimal numeric equality rather than simply checking object.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
@@ -30,6 +30,7 @@ import static org.junit.jupiter.api.Assertions.fail;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
+import java.math.BigDecimal;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
```diff
@@ -403,6 +404,20 @@ public void testEquals() {
         assertTrue(ObjectUtils.equals(FOO, FOO), "ObjectUtils.equals(\"foo\", \"foo\") returned false");
     }
 
+    @Test
+    public void TestEqualsOrCompareBigDecimal(){
+        assertTrue(ObjectUtils.equalsOrCompareBigDecimal(null,null));
```
garydgregory, line 409: Fix formatting: space after comma.

garydgregory (commented):
-1: Not a good idea IMO because this effectively overrides the implementation of `BigDecimal#equals()` which means that you get different results comparing two `BigDecimal` and comparing the same two `BigDecimal` if they happen to live inside a `Triple` or `Pair` which would, for me, violate the principle of least surprise.

aherbert:
This issue is related to the BigDecimal equals method only returning true to binary equality. The same issue arises for Double or Float where there are two representations of zero:
```java
Double d1 = Double.valueOf(-0.0);
Double d2 = Double.valueOf(0.0);
assert d1.equals(d2) == false;
Float f1 = Float.valueOf(-0.0f);
Float f2 = Float.valueOf(0.0f);
assert f1.equals(f2) == false;
```
I am not aware of other objects that can have a comparison that is 0 but not be equal. This applies to Numbers. I would suggest expanding the method to cover these cases of numbers:
```java
public static boolean equalsOrCompareNumber(final Object object1, final Object object2);
```

I have put an initial implementation on the Jira ticket. The details should be discussed there.

garydgregory:
@aherbert 
See my previous comment please, it does not make sense to me to have equality mean different things in different places. If anything, this issue suggests the need for a custom set of Number subclasses with a policy enum that dictates what kind of equality you want. But such a class belongs in Commons Numbers/Math, not Lang.

aherbert:
@garydgregory I had inferred that the use in Tuple has been ruled out (which I agree with; equals should always use equals). This is a suggestion to downgrade to just a static method to compute equality. I am not convinced of the utility of such a method anyway. My point was that handling BigDecimal without handling other cases seems wrong. So without any use case perhaps this is a dead end.

garydgregory:
@aherbert I agree, this PR is a dead end as it stands.

toyopilgrim:
@garydgregory  @aherbert 
Thank you guys for the feedback. I'm convinced of all the discussions above. 

>This issue suggests the need for a custom set of Number subclasses with a policy enum that dictates what kind of equality you want. But such a class belongs in Commons Numbers/Math, not Lang.

This kind of feedback is exactly what I wanted. An intention of that ignorable PR is to receive a point of view like if there's a right place for it. Do you think that it'd be worth adding such a class in a right package?

XenoAmess:
> @garydgregory @aherbert
> Thank you guys for the feedback. I'm convinced of all the discussions above.
> 
> > This issue suggests the need for a custom set of Number subclasses with a policy enum that dictates what kind of equality you want. But such a class belongs in Commons Numbers/Math, not Lang.
> 
> This kind of feedback is exactly what I wanted. An intention of that ignorable PR is to receive a point of view like if there's a right place for it. Do you think that it'd be worth adding such a class in a right package?

@toyopilgrim I think you'd better ask this question in the mailing list, means 
```
Commons Developers List <dev@commons.apache.org>
```
because if my memory be correct the guy mainly maintaining commons-math do not use github quite much

PR #782, opened 2021-07-28
[LANG-1666] Implemented ObjectUtils.equalsAny

felhag:
[LANG-1666](https://issues.apache.org/jira/browse/LANG-1666)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.TreeSet;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.exception.CloneFailedException;
 import org.apache.commons.lang3.function.Suppliers;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -634,6 +634,35 @@ public static boolean equals(final Object object1, final Object object2) {
...
+     * ObjectUtils.equalsAny(Boolean.TRUE, Boolean.TRUE)                = true
+     * ObjectUtils.equalsAny(Boolean.TRUE, Boolean.FALSE, Boolean.TRUE) = true
+     * </pre>
+     *
+     * @param object to compare, may be {@code null}
+     * @param searchObjects a vararg of objects, may be {@code null}.
+     * @return {@code true} if {@code object} is equal to any of the {@code searchObjects}.
+     */
+    public static boolean equalsAny(final Object object, final Object... searchObjects) {
+        if (ArrayUtils.isEmpty(searchObjects)) {
```
garydgregory, line 655: Or just `searchObjects== null ? false : Stream.of(searchObjects).anyMatch(e -> Objects.equals(object, e));`
-1 anyway since this duplicated the functionality in `ArrayUtils.contains(Object[], Object)`

garydgregory:
See also the method `org.apache.commons.lang3.ArrayUtils.containsAny(Object[], Object...)` in git master.

felhag:
I see, but I think it doesn't fully replace functionality for `ObjectUtils.equalsAny`. In my opinion building an inline array creates a lot of clutter, for example: 
`ArrayUtils.containsAny(new String[] {"foo", "bar"}, someObject.getSomeGetter());`
vs
`ObjectUtils.equalsAny(someObject.getSomeGetter(), "foo", "bar");`

In that case I think I would prefer something like `Arrays.asList("foo", "bar").contains(someObject.getSomeGetter())`.

garydgregory:
So... problem solved then? You can do:

Arrays.asList("foo", "bar").contains(someObject.getSomeGetter())

Instead of:

ObjectUtils.equalsAny(someObject.getSomeGetter(), "foo", "bar");

In one case you build an array, in the other a list.

felhag:
Yeah, I know but I think the second option looks better. But nvm, I will close this issue for now, thanks for your response.

PR #784, opened 2021-08-04
[LANG-1593] Common behavior for StringUtils join APIs when called with char or String delimiter

HubertWo:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4858,6 +5127,38 @@ public static String join(final short[] array, final char delimiter, final int s
         return join(elements, null);
     }
 
+    public static String join(byte[] elements, String separator) {
```
garydgregory, line 5130: All public methods should be Javadoc'd.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4818,13 +5052,48 @@ public static String join(final short[] array, final char delimiter) {
      * @since 3.2
      */
     public static String join(final short[] array, final char delimiter, final int startIndex, final int endIndex) {
+        return join(array, String.valueOf(delimiter), startIndex, endIndex);
+    }
+
+    /**
+     * <p>
```
garydgregory, line 5059: The 1st para of a Javadoc is automatically in a paragraph IIRC.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4307,13 +4481,48 @@ public static String join(final int[] array, final char separator) {
...
+     */
+    public static String join(final int[] array, final String delimiter, final int startIndex, final int endIndex) {
         if (array == null) {
             return null;
         }
         if (endIndex - startIndex <= 0) {
             return EMPTY;
         }
-        final StringJoiner joiner = newStringJoiner(delimiter);
+        final StringJoiner joiner = new StringJoiner(toStringOrEmpty(delimiter));
```
garydgregory, line 4525: Why did you remove the private helper method?

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4858,6 +5127,38 @@ public static String join(final short[] array, final char delimiter, final int s
...
+        return (elements == null) ? null : join(elements, separator, 0, elements.length);
+    }
+
+    public static String join(long[] elements, String separator) {
+        return (elements == null) ? null : join(elements, separator, 0, elements.length);
+    }
+
+    public static String join(short[] elements, String separator) {
+        return (elements == null) ? null : join(elements, separator, 0, elements.length);
+    }
```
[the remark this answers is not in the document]
HubertWo: Done.

src/main/java/org/apache/commons/lang3/StringUtils.java
HubertWo: I've followed convention from other Javadoc in the file. Fixed.

src/main/java/org/apache/commons/lang3/StringUtils.java
HubertWo: Valid point. I brought it back.

src/main/java/org/apache/commons/lang3/StringUtils.java
HubertWo: Added.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3922,6 +3922,37 @@ public static String join(final boolean[] array, final char delimiter) {
      * @since 3.12.0
      */
     public static String join(final boolean[] array, final char delimiter, final int startIndex, final int endIndex) {
+        return join(array, String.valueOf(delimiter), startIndex, endIndex);
+    }
+
+    /**
+     * <p>Joins the elements of the provided array into a single String containing the provided list of elements.</p>
```
garydgregory, line 3929: The first sentence in a Javadoc comment is special, it is the summary and should not be additionally wrapped in a paragraph.

HubertWo: Thanks for pointing this out, I didn't know about such rule. 
Additional paragraph tags removed.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -4858,6 +5082,209 @@ public static String join(final short[] array, final char delimiter, final int s
...
+     *
+     * @param elements
+     *            the values to join together, may be null
+     * @param separator
+     *            the separator String to use
+     * @return the joined String, {@code null} if null array input
+     * @since 3.13.0
+     */
+    public static String join(final char[] elements, final String separator) {
+        return elements == null ? null : join(elements, separator, 0, elements.length);
```
garydgregory, line 5160: The null checks in these methods are superfluous as the join method called already checks for the same condition. Less boilerplate is better IMO.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -1355,15 +1372,6 @@ public void testJoin_Objectarray() {
         assertEquals("foo2", StringUtils.join(MIXED_TYPE_LIST));
     }
 
-    @Disabled
```
garydgregory, line 1358: Was this test removed because the code is already covered elsewhere? Did you confirm this by looking at the JaCoCo report?

garydgregory:
@HubertWo 
Thank you for your update. Please see my comments.

XenoAmess:
I just think it use too many `String.valueOf`
Is it really performance acceptable?
I do think it it need a jmh...

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -1226,6 +1241,8 @@ public void testJoin_ArrayOfShorts() {
         assertNull(StringUtils.join((short[]) null, SEPARATOR_CHAR, 0, 1));
         assertEquals(StringUtils.EMPTY, StringUtils.join(SHORT_PRIM_LIST, SEPARATOR_CHAR, 0, 0));
         assertEquals(StringUtils.EMPTY, StringUtils.join(SHORT_PRIM_LIST, SEPARATOR_CHAR, 1, 0));
+        assertEquals("1,2", StringUtils.join(SHORT_PRIM_LIST, SEPARATOR));
```
garydgregory, line 1244: @HubertWo 
The point of String vs char separators is that a String can be longer than one character, so you should test for String separators longer than one character to make sure the right work takes place under the covers.

HubertWo: Sorry for the late reply, I was on vac. 
I will take a look on your comments and try to apply changes later this week.

src/main/java/org/apache/commons/lang3/StringUtils.java
HubertWo: 1. The method is public - so users may pass  ```null```  as ```elements``` which will cause ```NPE``` in the same line ```elements.length```
```
assertNull(StringUtils.join((char[]) null, "-")); // throws NPE if null check is removed
```
2. This case was not tested, so I added additional assertions in tests to cover it.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
HubertWo: This is exactly the test that exposes this bug: ```testLang1593```. 
Please find the same assertion here ```testJoin_ArrayOfInt``` in ```StringUtilsTest```

``` assertEquals("1,2", StringUtils.join(INT_PRIM_LIST, SEPARATOR));```

When it comes to coverage, it looks like ```join``` methods have 100% coverage.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
HubertWo: The joining part is done via ```java.util.StringJoiner``` with accepts only```CharSequence``` in constructor, so it's made to use more than one character as separator by default. Don't you think that testing that would be redundant?

HubertWo:
@XenoAmess 
Could you please be more descriptive? The only ```String.valueOf``` I have added are in high level methods, main logic does not. Actually I removed ```String.valueOf``` from loops.  Sorry for answering in comment, but there's no way to add reply to your comment.

HubertWo:
@garydgregory looks like I've addressed all the questions. 
Could you please review again?
Hope this will be LGTM this time :)

HubertWo:
@garydgregory BUMP

XenoAmess:
> @XenoAmess Could you please be more descriptive? The only `String.valueOf` I have added are in high level methods, main logic does not. Actually I removed `String.valueOf` from loops. Sorry for answering in comment, but there's no way to add reply to your comment.

@HubertWo 

Hi.

I find some time today for this question (finally).

`The only `String.valueOf` I have added are in high level methods, main logic does not`

This statement be correct, but also be where I worried about performance.

Please look at the sources in JDK, where `java.lang.AbstractStringBuilder#append(boolean b)`

In short, for basic types, should not convert them to String, but use StringBuilder directly, will bring far better performance.

BUT I just foundout that you did not bring this issue, this issue is already in commons-lang before your pr.

So I will not stop you from merging this in, but still I will find time to refactor the whole `join` serial functions later.

HubertWo:
> > @XenoAmess Could you please be more descriptive? The only `String.valueOf` I have added are in high level methods, main logic does not. Actually I removed `String.valueOf` from loops. Sorry for answering in comment, but there's no way to add reply to your comment.
> 
> @HubertWo
> 
> Hi.
> 
> I find some time today for this question (finally).
> 
> `The only `String.valueOf` I have added are in high level methods, main logic does not`
> 
> This statement be correct, but also be where I worried about performance.
> 
> Please look at the sources in JDK, where `java.lang.AbstractStringBuilder#append(boolean b)`
> 
> In short, for basic types, should not convert them to String, but use StringBuilder directly, will bring far better performance.
> 
> BUT I just foundout that you did not bring this issue, this issue is already in commons-lang before your pr.
> 
> So I will not stop you from merging this in, but still I will find time to refactor the whole `join` serial functions later.

I was not aware of that.
I will be happy to introduce changes (with JMH tests) related to your suggestions in separate PR.  
I've crated new JIRA for it: https://issues.apache.org/jira/browse/LANG-1675 

Thank you for your time @XenoAmess .

PR #785, opened 2021-08-07
LANG-1663: Update documentation to list correct exception for null array parameters

greatmastermario:
Updated documentation in NumberUtils to correctly state that a NullPointerException is thrown when the array is null, not an IllegalArgumentException.


HubertWo:
Regarding Javadoc - good catch! :) 
Did you consider to add assertThrows in tests for these methods?

greatmastermario:
I went to go write some test cases, and it looks like there are already test cases that use assertThrows with the correct NullPointerException. So we already have the test coverage, it just didn't reflect in the Javadoc.

garydgregory:
May you please rebase on master to pick up the recent Java 16 fix.

PR #786, opened 2021-08-07
Lang-1657: Diff Result Type Constraint

greatmastermario:
- Updated type parameter for DiffBuilder.append(String, DiffResult<T>) to accept DiffResults of any type in case nested field is not the same type as the containing class
- Added test case for nested Diffable types


garydgregory:
May you please rebase on master to pick up the recent Java 16 fix.

Osdel:
What happened to this pull request? This change is really needed. Having the `DiffBuilder<T>.append()` constraint on passing a `DiffResult<T>` is not ideal. Ideally you want to recursively call on nested attributes that implement the `Diffable` interface. I am not getting build errors but a lot of `Uncheck` warnings  which are unpleasant.

garydgregory:
Looking...

greatmastermario:
Since it has been a bit since I opened this PR, let me know if any edits
are required.

garydgregory:
@greatmastermario 
Hm, I'm having trouble apply the patch locally on master with `git apply`. Would you rebase on master please?

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
```diff
@@ -495,4 +519,16 @@ public void testTriviallyEqualTestEnabled() {
         assertThat(explicitTestAndEqual.build().getNumberOfDiffs(), equalToZero);
     }
 
+    @Test
+    public void testNestedDiffable() {
```
garydgregory, line 523: This test needs to be different: It compiles just fine AND without warnings without the change to `main`. What am I missing?

greatmastermario:
> @greatmastermario I'm having trouble applying the patch locally on master with `git apply`. Would you rebase on master, please?

Hi Gary, I am trying to do a git rebase, but it's considering every change from the last two years a "merge conflict" and making me manually apply all the changes. Should I close this PR, recreate the branch, and open a new PR?

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
greatmastermario: Let me validate again, but this should not be passing. The generic type T in master (did it get renamed to main? I don't see that in the repo) makes it where the passed diff needs to be the same type as the original builder. Making it ? allows other types of DiffBuilders to be passed. If this is not the behavior in master, I can rewrite the test.

garydgregory:
No, please don't open and close PRs and create noise. Just Google on how to use git and you'll get there, but for now, see my comment https://github.com/apache/commons-lang/pull/786#discussion_r1264740474

If you can't figure out git, don't worry, and we will deal with the PR as it is.

greatmastermario:
No problem, thought I would check. I typically use git pull to merge a
remote branch, so I'm not used to the rebase command. I can continue to
work through the rebate conflicts, was just trying to save some time.

garydgregory:
@greatmastermario 
Don't worry about git for now since GitHub does seem to see conflicts for a merge.

greatmastermario:
@garydgregory Apologies for taking so long to get that rebase done. Finally have everything worked out with the rebase. Let me know if there are further changes required.

src/test/java/org/apache/commons/lang3/builder/DiffBuilderTest.java
greatmastermario: After checking out main/master and copy-pasting the test without the code fix, I am getting a compilation error due to the generic types, which is what was asked to be fixed in the original issue. The test cannot compile without the NestedTypeTestClass, so I am unsure why main would be working. Screenshot proof prior to the fix is attached.
DiffBuilderGenericCompileFailure

garydgregory:
@greatmastermario 
Thank you for your PR. 
The solution in git master now includes your main change but with more tests. You are credited in `changes.xml`. 
Closing.

PR #796, opened 2021-08-28
Adopt is no more, long live Zulu!

geertjanw:


garydgregory:
We are using Eclipse Temurin.

carldea:
Do you still consider building on both distros Temurin and Zulu? ;-)

PR #797, opened 2021-08-30
Avoid multiple equivalent occurrences of the same expression.

arturobernalg:


src/main/java/org/apache/commons/lang3/ClassUtils.java
```diff
@@ -1365,12 +1365,10 @@ public static boolean isAssignable(Class<?> cls, final Class<?> toClass, final b
...
             }
-            if (Character.TYPE.equals(cls)) {
-                return Integer.TYPE.equals(toClass) || Long.TYPE.equals(toClass) || Float.TYPE.equals(toClass) || Double.TYPE.equals(toClass);
-            }
-            if (Short.TYPE.equals(cls)) {
+            if (Character.TYPE.equals(cls)  || Short.TYPE.equals(cls)) {
                 return Integer.TYPE.equals(toClass) || Long.TYPE.equals(toClass) || Float.TYPE.equals(toClass) || Double.TYPE.equals(toClass);
             }
+
             if (Byte.TYPE.equals(cls)) {
```
garydgregory, line 1372: Extra spacing not needed.

arturobernalg: HI. @garydgregory 
Fixed.
TY

PR #817, opened 2021-10-15
fix NumberFormatException in method org.apache.commons.lang3.math.Fra…

suoyi123wang:
the testcase fails for NumberFormatException
[
image
](url)

#### Reason:
when use
```
Double.parseDouble(str)
or
Integer.parseInt(strt)
```
don not throw NumberFormatException


garydgregory:
ping to @suoyi123wang

garydgregory:
Closing: No feedback.

PR #821, opened 2021-10-19
Betters names variables and  Fix typos

arturobernalg:


src/main/java/org/apache/commons/lang3/AnnotationUtils.java
```diff
@@ -69,9 +69,9 @@
          */
         @Override
         protected String getShortClassName(final Class<?> cls) {
-            for (final Class<?> iface : ClassUtils.getAllInterfaces(cls)) {
-                if (Annotation.class.isAssignableFrom(iface)) {
-                    return "@" + iface.getName();
+            for (final Class<?> iFace : ClassUtils.getAllInterfaces(cls)) {
```
garydgregory, line 72: -1: _iface_ is short for _interface_, not _interFace_. We can't use "interface" since its reserved. No change is needed IMO.

src/main/java/org/apache/commons/lang3/LocaleUtils.java
```diff
@@ -169,21 +169,21 @@ private static boolean isNumericAreaCode(final String str) {
         if (countryCode == null) {
             return Collections.emptyList();
         }
-        List<Locale> langs = cLanguagesByCountry.get(countryCode);
-        if (langs == null) {
-            langs = new ArrayList<>();
+        List<Locale> langByCountry = cLanguagesByCountry.get(countryCode);
```
garydgregory, line 172: Sorry, _langs_ was better IMO, mostly because it is _plural_ and is typed as a List. The new name is singular and therefore confusing (to me).

src/main/java/org/apache/commons/lang3/text/StrBuilder.java
```diff
@@ -3045,14 +3045,14 @@ public void write(final int c) {
 
         /** {@inheritDoc} */
         @Override
-        public void write(final char[] cbuf) {
-            StrBuilder.this.append(cbuf);
+        public void write(final char[] buffer) {
```
garydgregory, line 3048: -1: Better to keep the name matching the overridden method from java.io.Writer, it will be less of a surprise if you know the Writer API.

src/main/java/org/apache/commons/lang3/text/StrBuilder.java
```diff
@@ -3045,14 +3045,14 @@ public void write(final int c) {
...
+            StrBuilder.this.append(buffer);
         }
 
         /** {@inheritDoc} */
         @Override
-        public void write(final char[] cbuf, final int off, final int len) {
-            StrBuilder.this.append(cbuf, off, len);
+        public void write(final char[] buffer, final int off, final int len) {
```
garydgregory, line 3054: -1: Better to keep the name matching the overridden method from java.io.Writer, it will be less of a surprise if you know the Writer API.

src/main/java/org/apache/commons/lang3/text/WordUtils.java
```diff
@@ -402,8 +402,8 @@ public static String capitalize(final String str) {
      * @since 2.1
      */
     public static String capitalize(final String str, final char... delimiters) {
-        final int delimLen = delimiters == null ? -1 : delimiters.length;
-        if (StringUtils.isEmpty(str) || delimLen == 0) {
+        final int delimitersLength = delimiters == null ? -1 : delimiters.length;
```
garydgregory, line 405: -1: Subjective from me granted, not but better to read IMO.

src/main/java/org/apache/commons/lang3/text/WordUtils.java
```diff
@@ -525,8 +525,8 @@ public static String uncapitalize(final String str) {
      * @since 2.1
      */
     public static String uncapitalize(final String str, final char... delimiters) {
-        final int delimLen = delimiters == null ? -1 : delimiters.length;
-        if (StringUtils.isEmpty(str) || delimLen == 0) {
+        final int delimitersLength = delimiters == null ? -1 : delimiters.length;
```
garydgregory, line 528: -1: Subjective from me granted, not but better to read IMO.

garydgregory (changes requested):
@arturobernalg 
TY for your updates. Please see my comments.

PR #831, opened 2021-11-14
feat: list2tree method add

ejums:


garydgregory:
-1: No idea what this does as the doc is just one short sentence but it seems like some hyper specific collection processing method not generally useful. Feel free to elucidate. Also, instead of opening and closing PR after PR with the same code, please commit and optionally squash to the same PR.

ejums:
> -1: No idea what this does as the doc is just one short sentence but it seems like some hyper specific collection processing method not generally useful. Feel free to elucidate. Also, instead of opening and closing PR after PR with the same code, please commit and optionally squash to the same PR.

1. This method is a general method for converting the flattened list of any object type into a tree list, supporting operations such as computing hierarchy and judging whether it is the top and last level (refer to the corresponding test case for details of calling) 
2. As for the problem of multiple PR, it is because the jdoc format is incompatible and failed to pass the detection. I will pay attention to it later

garydgregory:
Closing: See my previous comment.

PR #838, opened 2021-11-30
LANG-1677 : Add ReflectionDiffBuilder.setExcludeFieldNames(...) and DiffExclude a…

debae:
Change for LANG-1677 to make it possible to exclude fields in ReflectionDiffBuilder. Similar solution like the ReflectionToStringBuilder has been applied.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -106,7 +106,7 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
     }
 
     /**
-     * Returns the field names that should be excluded from the diff
+     * Gets the field names that should be excluded from the diff
      * @return Returns the excludeFieldNames.
      * @since 3.13.0
      */
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -76,7 +76,7 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
     private final DiffBuilder<T> diffBuilder;
 
     /**
-     * Which field names to exclude from output. Intended for fields like {@code "password"} or {@code "lastModificationDate}.
+     * Which field names to exclude from output. Intended for fields like {@code "password"} or {@code "lastModificationDate"}.
      * @since 3.13.0
      */
     private String[] excludeFieldNames;
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -20,7 +20,10 @@ import static org.apache.commons.lang3.reflect.FieldUtils.readField;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 
+import org.apache.commons.lang3.ArraySorter;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
 
@@ -73,6 +76,12 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
     private final DiffBuilder<T> diffBuilder;
 
     /**
+     * Which field names to exclude from output. Intended for fields like {@code "password"} or {@code "lastModificationDate}.
+     * @since 3.13.0
+     */
+    private String[] excludeFieldNames;
+
+    /**
      * Constructs a builder for the specified objects with the specified style.
      *
      * <p>
@@ -94,6 +103,34 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
         this.left = lhs;
         this.right = rhs;
         diffBuilder = new DiffBuilder<>(lhs, rhs, style);
+    }
+
+    /**
+     * Returns the field names that should be excluded from the diff
+     * @return Returns the excludeFieldNames.
+     * @since 3.13.0
+     */
+    public String[] getExcludeFieldNames() {
+        return this.excludeFieldNames.clone();
+    }
+
+
+    /**
+     * Sets the field names to exclude.
+     *
+     * @param excludeFieldNamesParam
+     *            The field names to exclude from the diff or {@code null}.
+     * @return {@code this}
+     * @since 3.13.0
+     */
+    public ReflectionDiffBuilder setExcludeFieldNames(final String... excludeFieldNamesParam) {
+        if (excludeFieldNamesParam == null) {
+            this.excludeFieldNames = ArrayUtils.EMPTY_STRING_ARRAY;
+        } else {
+            // clone and remove nulls
+            this.excludeFieldNames = ArraySorter.sort(ReflectionToStringBuilder.toNoNullStringArray(excludeFieldNamesParam));
+        }
+        return this;
     }
 
     @Override
@@ -128,7 +165,15 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
         if (Modifier.isTransient(field.getModifiers())) {
             return false;
         }
-        return !Modifier.isStatic(field.getModifiers());
+        if (Modifier.isStatic(field.getModifiers())) {
+            return false;
+        }
+        if (this.excludeFieldNames != null
+                && Arrays.binarySearch(this.excludeFieldNames, field.getName()) >= 0) {
+            // Reject fields from the getExcludeFieldNames list.
+            return false;
+        }
+        return !field.isAnnotationPresent(DiffExclude.class);
     }
 
 }
--- a/src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
@@ -17,6 +17,7 @@
 package org.apache.commons.lang3.builder;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import org.apache.commons.lang3.AbstractLangTest;
 import org.junit.jupiter.api.Test;
@@ -48,10 +49,14 @@ public class ReflectionDiffBuilderTest extends AbstractLangTest {
         private final Object[] objectArrayField = {null};
         private static int staticField;
         private transient String transientField;
+        @DiffExclude
+        private String annotatedField = "a";
+        private String excludedField = "a";
+
 
         @Override
         public DiffResult diff(final TypeTestClass obj) {
-            return new ReflectionDiffBuilder(this, obj, style).build();
+            return new ReflectionDiffBuilder(this, obj, style).setExcludeFieldNames("excludedField").build();
         }
 
         @Override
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -73,6 +75,11 @@
     private final Object right;
     private final DiffBuilder<T> diffBuilder;
 
+    /**
+     * Which field names to exclude from output. Intended for fields like {@code "password"} or {@code "lastModificationDate}.
+     */
```
garydgregory, line 80: All _new_ `protected` and `public` elements new `@since 3.13.0`.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -99,6 +106,31 @@ public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style
         diffBuilder = new DiffBuilder<>(lhs, rhs, style);
     }
 
+    /**
+     * @return Returns the excludeFieldNames.
```
garydgregory, line 110: Javadoc comments start with a sentence.

garydgregory (changes requested):
@debae 
Thank you for your PR. Please see comments.

garydgregory:
Hi @debae 
Thank you for your patience. The code coverage for this new feature is not good enough IMO. Run `mvn clean site`, then browse the site in `target\site\index.html`, navigate to the JaCoCo report and you will see how your tests should be improved. For _new_ code, I'd like to get 100% code coverage which should be reasonable in this case.
TY!

garydgregory (changes requested):
See my previous comment.

debae:
> Hi @debae Thank you for your patience. The code coverage for this new feature is not good enough IMO. Run `mvn clean site`, then browse the site in `target\site\index.html`, navigate to the JaCoCo report and you will see how your tests should be improved. For _new_ code, I'd like to get 100% code coverage which should be reasonable in this case. TY!

No problem at all, some extra tests have been added.

debae:
> See my previous comment.

Hi @garydgregory , extra testcases have been added.

garydgregory:
I'll take a look over the weekend.

src/changes/changes.xml
```diff
@@ -66,6 +66,7 @@ The <action> type attribute can be add,update,fix,remove.
     <action                   type="fix" dev="ggregory" due-to="CiprianBodnarescu">Use Set instead of List for checking the contains() method #734.</action>
     <action                   type="fix" dev="kinow" due-to="Roland Kreuzer">Javadoc for StringUtils.substringBefore(String str, int separator) doesn't mention that the separator is an int.</action>
     <!-- ADD -->
+    <action issue="LANG-1677" type="add" dev="debae" due-to="Dennis Baerten">Add ReflectionDiffBuilder.setExcludeFieldNames(...) and DiffExclude annotation to support field exclusion from diff.</action>
```
garydgregory, line 69: Don't edit this file, it will usually clash depending on parallel changes.

debae: Removed the file again.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -99,6 +107,34 @@ public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style
...
+     * Sets the field names to exclude.
+     *
+     * @param excludeFieldNamesParam
+     *            The field names to exclude from the diff or {@code null}.
+     * @return {@code this}
+     * @since 3.13.0
+     */
+    public ReflectionDiffBuilder setExcludeFieldNames(final String... excludeFieldNamesParam) {
+        if (excludeFieldNamesParam == null) {
+            this.excludeFieldNames = new String[0];
```
garydgregory, line 130: Reuse `ArrayUtils.EMPTY_STRING_ARRAY`

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -99,6 +107,34 @@ public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style
...
+        } else {
+            //clone and remove nulls
```
garydgregory, line 132: Space after `//`

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -73,6 +75,12 @@
     private final Object right;
     private final DiffBuilder<T> diffBuilder;
 
+    /**
+     * Which field names to exclude from output. Intended for fields like {@code "password"} or {@code "lastModificationDate}.
+     * @since 3.13.0
+     */
+    protected String[] excludeFieldNames;
```
garydgregory, line 82: Make `private`.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -131,7 +167,15 @@ private boolean accept(final Field field) {
         if (Modifier.isTransient(field.getModifiers())) {
             return false;
         }
-        return !Modifier.isStatic(field.getModifiers());
+        if (Modifier.isStatic(field.getModifiers())) {
+            return false;
+        }
+        if (this.excludeFieldNames != null
```
garydgregory, line 173: If you make the default value the empty array and the setter does not allow excludeFieldNames to be null, then you do not need this check? It might be better this way though, not sure.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -92,7 +96,7 @@ public void test_primitive_difference() {
     @Test
     public void test_array_difference() {
         final TypeTestClass firstObject = new TypeTestClass();
-        firstObject.charArrayField = new char[] { 'c' };
+        firstObject.charArrayField = new char[]{'c'};
```
garydgregory, line 99: Leave formatting alone here.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -128,4 +132,43 @@ public void test_difference_in_inherited_field() {
...
+        firstObject.annotatedField = "b";
+        final TypeTestClass secondObject = new TypeTestClass();
+
+        final DiffResult list = firstObject.diff(secondObject);
+        assertEquals(0, list.getNumberOfDiffs());
+    }
+
+    @Test
+    public void testGetExcludeFieldNamesWithNullExcludedFieldNames() {
+        ReflectionDiffBuilder<TypeTestClass> reflectionDiffBuilder = new ReflectionDiffBuilder<>(new TypeTestClass(), new TypeTestChildClass(), SHORT_STYLE);
```
garydgregory, line 158: Use final where you can.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -128,4 +132,43 @@ public void test_difference_in_inherited_field() {
...
+        reflectionDiffBuilder.setExcludeFieldNames(null);
+        String[] excludeFieldNames = reflectionDiffBuilder.getExcludeFieldNames();
+        assertNotNull(excludeFieldNames);
+        assertEquals(0, excludeFieldNames.length);
+    }
+
+    @Test
+    public void testGetExcludeFieldNamesWithNullValuesInExcludedFieldNames() {
+        ReflectionDiffBuilder<TypeTestClass> reflectionDiffBuilder = new ReflectionDiffBuilder<>(new TypeTestClass(), new TypeTestChildClass(), SHORT_STYLE);
```
garydgregory, line 167: Use final where you can.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -48,10 +48,14 @@
         private final Object[] objectArrayField = {null};
         private static int staticField;
         private transient String transientField;
+        @DiffExclude
+        private String annotatedField = "a";
+        private String excludedField = "a";
```
garydgregory, line 53: You should test a non-transient null field.

garydgregory (changes requested):
Hi @debae 
Please see my comments.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
debae: There is already a testcase `test_transient_field_difference` that validates this, there is no specific or new logic for transient fields that get excluded.

debae:
@garydgregory friendly reminder on this PR , are requested changes are tackled.

garydgregory:
I'll look over the weekend.

debae:
@garydgregory did you have the time too look into this PR

garydgregory:
I will be busy this weekend with Commons BCEL and Commons Net, so probably after that.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -128,4 +133,44 @@ public void test_difference_in_inherited_field() {
...
+
+    @Test
+    public void test_no_differences_diff_exclude_annotated_field() {
+        final TypeTestClass firstObject = new TypeTestClass();
+        firstObject.annotatedField = "b";
+        final TypeTestClass secondObject = new TypeTestClass();
+
+        final DiffResult list = firstObject.diff(secondObject);
+        assertEquals(0, list.getNumberOfDiffs());
+    }
```
kinow, line 155: Maybe add a test that does both `annotatedField` and `excludedField` being excluded.

kinow (commented):
Looks good to me. Left a couple of comments but nothing too important nor a blocker. It needs to be squashed to a single commit, preferably with the LANG-1677: issue ID as prefix (as other commits).

Had a look at other builders, like `EqualsBuilder`, and it has a very similar feature already. Only difference I noted is that in the Javadocs of that builder, there are several links in `@see EqualsExcclude`, I think to indicate to the developers that the code in context is not applicable to annotated members. Not sure if we need that here, nor it other builders also have that (if so would be good to maintain consistency).

+1

Thanks!
-Bruno

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
debae: If you do invoke the method setExcludeFieldNames() it will have a default value of null, and in that cause a NPE here.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
debae: Extra testcase added

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
kinow: Hmm, I think I didn't understand the whole discussion here.

@garydgregory pointed that instead of having a default of `null`, we could have something like `private String[] excludeFieldNames = ArrayUtils.EMPTY_STRING_ARRAY;`

>If you do invoke the method setExcludeFieldNames() it will have a default value of null, and in that cause a NPE here.

That sounds incorrect, I think (it's late, so I could be wrong). If you call `setExcludeFieldNames`, there's a check if the value passed is `null`, and `this.excludeFieldNames = ArrayUtils.EMPTY_STRING_ARRAY;`. Thus, `excludeFieldNames` cannot be `null`.

So if you did as @garydgregory suggested, and if we keep the defensive check for `null` in `setExcludeFieldNames`, then I believe what @garydgregory initially suggested holds here? 

>If you make the default value the empty array and the setter does not allow excludeFieldNames to be null, then you do not need this check? It might be better this way though, not sure.

Then, I think the question would be whether it's worth having the `binarySearch` knowing it's an empty array (maybe we could skip if if we know it's an empty array).

debae:
@kinow and @garydgregory Can you guys have another look at this PR, all remarks have been processed.

garydgregory (approved):

garydgregory:
https://issues.apache.org/jira/browse/LANG-1667

PR #848, opened 2022-01-18
LANG-1682 Adding StringUtils.startsWithAnyIgnoreCase method

Enigo:
Adding `StringUtils.startsWithAnyIgnoreCase` to have more flexibility.
Based on the existing `startsWith` method

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8129,11 +8129,52 @@ public class StringUtils {
      * @since 3.0 Changed signature from startsWithAny(String, String[]) to startsWithAny(CharSequence, CharSequence...)
      */
     public static boolean startsWithAny(final CharSequence sequence, final CharSequence... searchStrings) {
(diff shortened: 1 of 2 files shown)


LarsBodewig:
Concurrent open PR for LANG-1682: #852

Enigo:
Hey @garydgregory 
any chance for this PR to be reviewed and merged?
thanks!

garydgregory:
Needs further thought: I don't think it makes sense to overcrowd this class with more "ignore case" versions of methods and to also have "any" or "all" versions of methods. Instead, we should consider another design, maybe having a string util class with a subclass for case sensitive and another for cases insensitive matches, for example.

Enigo:
I'm not sure if that would be user-friendly, as most users are used to one single class for all operations with strings, numbers, lists, sets etc.
That would be quite a drastic change
But I do agree that this class is quite crowded already

garydgregory:
@Enigo @LarsBodewig 
Closing: Please try your use case with the Strings class in git master and snapshot builds.

PR #850, opened 2022-01-27
improve DateUtils by adding calenda.month in comparison in considerin…

SophieHYe:
Modified DateUtils method isSameLocalTime: adding the same month in consideration when comparing the same local time :)


aherbert:
Note that this calendar comparison currently uses DAY_OF_YEAR. If we use DAY_OF_MONTH then it would require a check of the month too.

The DAY_OF_YEAR check will compare all 365 (366) possible day values. This makes the DAY_OF_MONTH check is redundant.

garydgregory:
Hello @SophieHYe and thank you for your interest in Apache Commons Lang ! :-)
Without looking too deeply at the specifics of this change I can tell you that this kind of update or fix should be accompanied by a change in Javadoc (if it matters which may not be the case here) and a test that fails without the main change to show what the PR does. Now, specifically, if you cannot cause a test to fail with the current behavior, then you've not fixed anything, so what are you trying to fix? How can the months mismatch when the day of year match? What am I missing?

garydgregory:
Ping?

garydgregory:
Closing, zero feedback from OP.

PR #851, opened 2022-02-09
LANG-1671: AtomicSafeInitializerTest and LazyInitializerTest modifications using Mockito

JKLedzion:
PR contains proposed changes to Jira issue 1671: 
* replacing AtomicSafeInitializerTestImpl  with a spy using Mockito,
* stubbing a method initialize(),
* adding verification that method initialize() is called only once.

Similar modification I’ve done in class LazyInitializerTest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -588,6 +588,18 @@
       <version>3.0.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${commons.mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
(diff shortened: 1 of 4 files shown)


garydgregory:
A PR without a description, even if it is only a URL to a ticket, is harder to follow than it has to be.

JKLedzion:
Hi @garydgregory. Good point. I’ve added description to PR also.

garydgregory:
I've looked at this again and decided that while it would be OK to have this type of change in principle, I want to use the tests as a simple example in this case. I've made adjustments to the Javadoc to reflect this intent. 

The mocking here is fine (again, in principle) but it makes the code much harder to follow and presents an additional hurdle to people wanting to understand the library. Personally, I prefer to use mocking when it is the only solution for testing or the normal way to test requires too much code. I do not feel this is the case here.

Thank you for taking the time to provide this PR though.

PR #852, opened 2022-02-09
startsWithAnyIgnoreCase method

Kaammill:
Hi, this is my first commit therefore all kind of feedback is more the welcome :)

I added the  startsWithAnyIgnoreCase  described in this topic:
https://issues.apache.org/jira/projects/LANG/issues/LANG-1682?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8147,6 +8147,41 @@ public class StringUtils {
      * references are considered to be equal. The comparison is case insensitive.</p>
      *
      * <pre>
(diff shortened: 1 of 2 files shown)


LarsBodewig:
Concurrent open PR for LANG-1682: #848

garydgregory:
-1: I do not think we want to have an "ignore case" version of this and that, or all APIs. Looking ahead, I see a `Strings` class with configurable attributes like the case sensitivity, charset, and maybe a few other items, then you can say `Strings.CASE_INSENSITIVE.startsWith(...)` and `Strings.CASE_SENSITIVE.startsWith(...)`

PR #855, opened 2022-02-18
Add countBlanks() function

avandal:
Count how many blank CharSequences are empty (""), null or whitespace only.


garydgregory:
1) LANG-123 is NOT about this issue, will edit
2) this does not seem like it belongs in a low level library. For example I would expect String to implement such a method. Maybe in Commons Text since this is more about interpreting text.

garydgregory:
Zero feedback from OP, closing.

PR #857, opened 2022-02-23
Add method defaultIfEmpty with lazy default string computation

Mikhail2048:
I have encountered a following use case: we have a java object in the local cache, that might have or might have not a `String` field. In case this String field is empty we need to do AWS S3 lookup for that object and fetch this String field from pulled object. Problem is that `defaultIfEmpty(final T source, final T default)` method that already present in `StringUtils` will do S3 lookup in any case, even if source String is present. I think it would be a great idea to create almost the same method but with lazy computation of default string, exactly for that kind of use cases.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -1153,6 +1153,20 @@ public class ObjectUtils {
     }
 
     /**
+     * Get passed {@code object} in case it is not null, otherwise return the value produced by supplier
+     *
+     * Consider to utilize this method if lazy computation of default value makes sense in your case
+     *
+     * @param object to get if not null else to apply passed {@link Supplier}
+     * @param supplier {@link Supplier} to utilize in case provided {@code object} is null
+     * @param <T> the type of the object
+     * @return passed {@code object} in case it is not null, otherwise return the value produced by supplier
+     */
+    public static <T> T defaultIfNull(T object, Supplier<T> supplier) {
+        return object == null ? supplier.get() : object;
+    }
+
+    /**
      * <p>Null safe comparison of Comparables.</p>
      * <p>TODO Move to ComparableUtils.</p>
      *
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1525,6 +1525,26 @@ public class StringUtils {
 
     /**
      * <p>Returns either the passed in CharSequence, or if the CharSequence is
+     * whitespace, empty ("") or {@code null}, the value produced by {@code supplier}.</p>
+     *
+     * Consider to utilize this method instead of {@link #defaultIfBlank(CharSequence, CharSequence)}
+     * in case of lazy computation of default string make sense
+     *
+     * <p>Whitespace is defined by {@link Character#isWhitespace(char)}.</p>
+     *
+     * @param source - CharSequence to check, may be null
+     * @param supplier represents the supplier, that should return instance
+     *                 of type {@literal T} in case source string is empty
+     * @param <T> represents the specific kind of CharSequence
+     * @return the passed source instance of CharSequence in case it is not blank,
+     *                 otherwise the value returned by {@link Supplier}
+     */
+    public static <T extends CharSequence> T defaultIfEmpty(final T source, final Supplier<T> supplier) {
+        return isBlank(source) ? supplier.get() : source;
+    }
+
+    /**
+     * <p>Returns either the passed in CharSequence, or if the CharSequence is
      * empty or {@code null}, the value of {@code defaultStr}.</p>
      *
      * <pre>
--- a/src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ObjectUtilsTest.java
@@ -17,6 +17,7 @@
 package org.apache.commons.lang3;
 
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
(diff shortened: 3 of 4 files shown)


ecki:
Looks like a useful extension. I think this would be acceptable trivial as a submission, but just in case do you know if you are covered by individual or corporate contributors Agreement?

Mikhail2048:
> Looks like a useful extension. I think this would be acceptable trivial as a submission, but just in case do you know if you are covered by individual or corporate contributors Agreement?

Individual, I suppose) I am not covered by any corporate contribution Agreement, it is purely my own initiative to contribute)

Mikhail2048:
> > > Hello @Mikhail2048 & @ecki The new method name does not match its internal test, one or the other needs fixing. Since the method works on any Object, not a String, it likely belongs next to `ObjectUtils.toString(Object, Supplier<String>)`. Please adjust this PR.
> > 
> > 
> > I think it is not entirely true. We cannot test any given object for emptiness, so, if we have `ObjectUtils.toString(Object, Supplier<String>)`, then I do not understand how should we test passed object for emptiness. We can, of course, test if passed object is `null` and then apply `Supplier`, but empty string and null string are not the same.
> > I think the appropriate solution would look like this: in general is a great idea to implement something like `ObjectUtils.applyIfNull(T, Supplier<T>)`, but I think we should left `StringUtils.defaultIfEmpty(final T source, final Supplier<T> default)`, maybe with slightly different name, for example `StringUtils.applyIfNull(final T source, final Supplier<T> default)`
> 
> We have ObjectUtils#isEmpty() but we could have a String specific version I suppose...

Well, yes, but if we should have a String specific version of isEmpty(), then I think we really need to think twice on why we are placing this in ObjectUtils class. It is, from my opinion, a bit controversial - we are placing `String` specific logic into ObjectUtils.

I think that the best solution would be - to create general purpose utility method in `ObjectUtils`, like `ObjectUtils#defaultIfNull(T object, Supplier<T> supplier)`, but specifically for strings we can have `StringUtils#defaultIfEmpty(String object, Supplier<String> supplier)` - because logic of how we verify emptiness of String/CharSequence is different from any other object. 

@garydgregory I think it will be a great solution, is not it?)

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -1152,6 +1152,20 @@ public static boolean isNotEmpty(final Object object) {
...
+     *
+     * Consider to utilize this method if lazy computation of default value makes sense in your case
+     *
+     * @param object to get if not null else to apply passed {@link Supplier}
+     * @param supplier {@link Supplier} to utilize in case provided {@code object} is null
+     * @param <T> the type of the object
+     * @return passed {@code object} in case it is not null, otherwise return the value produced by supplier
+     */
+    public static <T> T defaultIfNull(T object, Supplier<T> supplier) {
+        return object == null ? supplier.get() : object;
```
garydgregory, line 1166: This duplicates `ObjectUtils.getIfNull(T, Supplier<T>)`

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -1523,6 +1523,26 @@ public static int countMatches(final CharSequence str, final CharSequence sub) {
...
+     *
+     * @param source - CharSequence to check, may be null
+     * @param supplier represents the supplier, that should return instance
+     *                 of type {@literal T} in case source string is empty
+     * @param <T> represents the specific kind of CharSequence
+     * @return the passed source instance of CharSequence in case it is not blank,
+     *                 otherwise the value returned by {@link Supplier}
+     */
+    public static <T extends CharSequence> T defaultIfEmpty(final T source, final Supplier<T> supplier) {
+        return isBlank(source) ? supplier.get() : source;
```
garydgregory, line 1543: This duplicates:
- `StringUtils.getIfBlank(T, Supplier<T>)`
- `StringUtils.getIfEmpty(T, Supplier<T>)`

garydgregory:
> Looks like a useful extension. I think this would be acceptable trivial as a submission, but just in case do you know if you are covered by individual or corporate contributors Agreement?

This PR duplicates existing functionality. See my other comments in the code.

src/main/java/org/apache/commons/lang3/ObjectUtils.java
garydgregory: @Mikhail2048  right?

src/main/java/org/apache/commons/lang3/StringUtils.java
garydgregory: @Mikhail2048  right?

src/main/java/org/apache/commons/lang3/ObjectUtils.java
Mikhail2048: I  am sorry, but I have not found any method in `ObjectUtils` with such signature.  There is `ObjectUtils.getIfNull(T, T)`  method, but it is not a lazy implementation

garydgregory: See https://github.com/apache/commons-lang/blob/02f3bee7bce4bebb4e1bfe27b613a46adedccef6/src/main/java/org/apache/commons/lang3/ObjectUtils.java#L744

Mikhail2048: @garydgregory Yeah, thats true)

src/main/java/org/apache/commons/lang3/StringUtils.java
Mikhail2048: @garydgregory correct)

src/main/java/org/apache/commons/lang3/ObjectUtils.java
Mikhail2048: Yeah we can close this PR, it seems that this functionality is already present, thank you!

PR #862, opened 2022-03-06
LANG-1668 Update getMatchingMethod documentation.

mbuiakova:
Adds a mention of IllegalStateException in case no unique method is found.


garydgregory:
@mbuiakova 
This change does not appear to be accurate: see the case "candidates.isEmpty()" compared to the message of the exception. Or, am I missing something?

mbuiakova:
> @mbuiakova This change does not appear to be accurate: see the case "candidates.isEmpty()" compared to the message of the exception. Or, am I missing something?

Yes, you are right. I changed it back.

PR #863, opened 2022-03-06
[LANG-1462] Use TimeZone from calendar in DateFormatUtils.

mbuiakova:
Uses timezone from the calendar object when no timezone is specified.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DateFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateFormatUtils.java
@@ -271,7 +271,8 @@ public class DateFormatUtils {
     }
 
     /**
-     * <p>Formats a calendar into a specific pattern.</p>
+     * <p>Formats a calendar into a specific pattern. Timezone from the calendar
+     * will be used for formatting.</p>
      *
      * @param calendar  the calendar to format, not null
      * @param pattern  the pattern to use to format the calendar, not null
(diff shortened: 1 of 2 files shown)


garydgregory:
Would you please explain why this is needed?

Also:
- there are no tests
- the code has superfluous parentheses

mbuiakova:
> Would you please explain why this is needed?
> 
> Also:
> 
> * there are no tests
> * the code has superfluous parentheses

I added a test and removed extra parentheses. 
The reason for these changes is that Calendar has timeZone, so it is expected by a user of the `format(final Calendar calendar, final String pattern)`-like methods that method gets the timezone from a calendar object. More detailed explanation [here](https://issues.apache.org/jira/browse/LANG-1462).

Can you help please with this error, how to deal with it?

```
Error:  High: Null passed for non-null parameter of format(Calendar, String, TimeZone, Locale) in org.apache.commons.lang3.time.DateFormatUtils.format(Calendar, String) [org.apache.commons.lang3.time.DateFormatUtils, org.apache.commons.lang3.time.DateFormatUtils] Method invoked at DateFormatUtils.java:[line 283]Known null at DateFormatUtils.java:[line 283] NP_NULL_PARAM_DEREF
```

garydgregory:
@mbuiakova 
You'll need to update the Javadoc to reflect the change in behavior.
For the error, I'd start with Google ;-)

mbuiakova:
> @mbuiakova You'll need to update the Javadoc to reflect the change in behavior. For the error, I'd start with Google ;-)

Thank you for your feedback, I updated the Javadoc.
I think this error is false-positive because I can't see where the parameter is said to be non-null and we check for the null value everywhere. Maybe we should suppress it?

garydgregory:
https://issues.apache.org/jira/browse/LANG-1462

garydgregory:
Fixed slightly differently in git master but you are still credited in changes.xml. TY!

PR #865, opened 2022-03-25
Refactoring the files.

shivambhojani:
Hello,

I am working on refactoring the project and found that few files could be refactored and used more efficiently.
Refactoring is done in such a way that all the existing Unit test cases are passing without any issue.


shivambhojani:
Will be raising new one after one more commit

PR #866, opened 2022-03-25
Refactoring the code

shivambhojani:
Hello,

I am working on refactoring the project and found that few files could be refactored and used more efficiently.
Refactoring is done in such a way that all the existing Unit test cases are passing without any issue.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BooleanUtils.java
+++ b/src/main/java/org/apache/commons/lang3/BooleanUtils.java
@@ -622,17 +622,7 @@ public class BooleanUtils {
 
     public static Boolean toBooleanObject(final Integer value, final Integer trueValue, final Integer falseValue, final Integer nullValue) {
         if (value == null) {
-
             return checkTrueFalseNullValues(trueValue,falseValue,nullValue);
-//            if (trueValue == null) {
-//                return Boolean.TRUE;
-//            }
-//            if (falseValue == null) {
-//                return Boolean.FALSE;
-//            }
-//            if (nullValue == null) {
-//                return null;
-//            }
         } else if (value.equals(trueValue)) {
             return Boolean.TRUE;
         } else if (value.equals(falseValue)) {
--- a/src/main/java/org/apache/commons/lang3/BooleanUtils.java
+++ b/src/main/java/org/apache/commons/lang3/BooleanUtils.java
@@ -599,24 +599,40 @@ public class BooleanUtils {
      *   BooleanUtils.toBooleanObject(Integer.valueOf(3), Integer.valueOf(1), Integer.valueOf(2), Integer.valueOf(3)) = null
      * </pre>
      *
-     * @param value  the Integer to convert
      * @param trueValue  the value to match for {@code true}, may be {@code null}
      * @param falseValue  the value to match for {@code false}, may be {@code null}
      * @param nullValue  the value to to match for {@code null}, may be {@code null}
      * @return Boolean.TRUE, Boolean.FALSE, or {@code null}
      * @throws IllegalArgumentException if no match
      */
+
+    public static Boolean checkTrueFalseNullValues(Integer trueValue, Integer falseValue, Integer nullValue){
+
+        if (trueValue==null){
+            return Boolean.TRUE;
+        }
+        else if (falseValue == null){
+            return Boolean.FALSE;
+        }
+        else if(nullValue == null) {
+            return null;
+        }
+        throw new IllegalArgumentException("The Integer did not match any specified value");
+    }
+
     public static Boolean toBooleanObject(final Integer value, final Integer trueValue, final Integer falseValue, final Integer nullValue) {
         if (value == null) {
-            if (trueValue == null) {
-                return Boolean.TRUE;
-            }
-            if (falseValue == null) {
-                return Boolean.FALSE;
-            }
-            if (nullValue == null) {
-                return null;
-            }
+
+            return checkTrueFalseNullValues(trueValue,falseValue,nullValue);
+//            if (trueValue == null) {
+//                return Boolean.TRUE;
+//            }
+//            if (falseValue == null) {
+//                return Boolean.FALSE;
+//            }
+//            if (nullValue == null) {
+//                return null;
+//            }
         } else if (value.equals(trueValue)) {
             return Boolean.TRUE;
         } else if (value.equals(falseValue)) {
@@ -784,6 +800,7 @@ public class BooleanUtils {
      */
     public static Boolean toBooleanObject(final String str, final String trueString, final String falseString, final String nullString) {
         if (str == null) {
+
             if (trueString == null) {
                 return Boolean.TRUE;
             }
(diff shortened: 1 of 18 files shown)


src/main/java/org/apache/commons/lang3/CSVEscapeUtils.java
```diff
@@ -0,0 +1,44 @@
+package org.apache.commons.lang3;
```
kinow, line 1: Every file in Lang must include a license header.

src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -681,13 +681,13 @@ public static Boolean toBooleanObject(final String str) {
             case 1: {
                 final char ch0 = str.charAt(0);
                 if (ch0 == 'y' || ch0 == 'Y' ||
-                    ch0 == 't' || ch0 == 'T' ||
-                    ch0 == '1') {
+                        ch0 == 't' || ch0 == 'T' ||
+                        ch0 == '1') {
```
kinow, line 685: I think this is just cosmetic, not really fixing an issue. Changes like this make things harder to review. It's better to discuss with the developers of the project before changing code style.

src/main/java/org/apache/commons/lang3/BooleanUtils.java
```diff
@@ -1132,4 +1132,4 @@ public static Boolean xor(final Boolean... array) {
     public BooleanUtils() {
     }
 
-}
+}
```
kinow, line 1135: Missing newline.

src/main/java/org/apache/commons/lang3/time/GmtTimeZone.java
```diff
@@ -31,6 +36,32 @@
     private static final int MINUTES_PER_HOUR = 60;
     private static final int HOURS_PER_DAY = 24;
 
+    private static final TimeZone timezone = new GmtTimeZone(false,0,0);
+    private static final Pattern GMT_PATTERN1 = Pattern.compile("^(?:(?i)GMT)?([+-])?(\\d\\d?)?(:?(\\d\\d?))?$");
+
+    public static TimeZone fetchGmtTimeZone(final String pattern) {
```
kinow, line 42: Other methods in this class follow the get/set pattern. This `fetchGmt...` method is being called from `FastTimeZone#getGmtTimeZone`. So it could probably stay as `getGmt...`, I think.

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -29,14 +29,14 @@
...
-    private static final TimeZone GREENWICH = new GmtTimeZone(false, 0, 0);
+    //private static TimeZone GREENWICH =
 
     /**
      * Gets the GMT TimeZone.
      * @return A TimeZone with a raw offset of zero.
      */
     public static TimeZone getGmtTimeZone() {
-        return GREENWICH;
+        return new GmtTimeZone(false, 0, 0);
```
kinow, line 39: I believe the reason for the static object was to avoid instantiating a new object every time this method was called.

src/main/java/org/apache/commons/lang3/time/FastTimeZone.java
```diff
@@ -29,14 +29,14 @@
 
     private static final Pattern GMT_PATTERN = Pattern.compile("^(?:(?i)GMT)?([+-])?(\\d\\d?)?(:?(\\d\\d?))?$");
 
-    private static final TimeZone GREENWICH = new GmtTimeZone(false, 0, 0);
+    //private static TimeZone GREENWICH =
```
kinow, line 32: We tend to only leave comments that are intended to help other developers, or things that are almost-ready to be finished, for instance. This one could probably be removed.

src/main/java/org/apache/commons/lang3/exception/ContextedRuntimeException.java
```diff
@@ -214,13 +214,13 @@ public Object getFirstContextValue(final String label) {
...
-    public Set<String> getContextLabels() {
-        return exceptionContext.getContextLabels();
-    }
+//    /**
+//     * {@inheritDoc}
+//     */
+//    @Override
+//    public Set<String> getContextLabels() {
+//        return exceptionContext.getContextLabels();
+//    }
```
kinow, line 223: It's hard to understand why this was commented out. This is also a public method. We cannot remove it without a major release - https://semver.org/

kinow (changes requested):
Hi @shivambhojani 

I think it's better to discuss your plans of refactoring in the dev mailing list, before working on such a large change like this one.

I added a few comments in this PR just to point to things that normally are asked in review in Commons. I think the main issues here are that you have multiple things in this single PR (changing styles, removing code, adding new code, extracting new methods from existing code to simplify code, etc.)

This complicates things for reviewers. But also complicate for you, as as we ask for changes in your pull request, the rest of the code may move forward, forcing you to rebase and fix conflicts in your code.

I suggest to create a JIRA issue to discuss if you would like to refactor the code. If the change as purely cosmetic (formatting, fix IDE issues) they might be postponed or never-done. If you have multiple changes in one issue, developers may ask you to split it into smaller issues.

One of the most important things to remember too, when working on Apache Commons components, is to maintain binary backward compatibility - https://garygregory.wordpress.com/tag/binary/

Hope that helps you to stay motivated and keep thinking in how to fix issues and improve Commons Lang and other components :)

Cheers
Bruno

src/main/java/org/apache/commons/lang3/BooleanUtils.java
garydgregory: I agree with @kinow

src/main/java/org/apache/commons/lang3/CSVEscapeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.apache.commons.lang3.text.translate.CharSequenceTranslator;
+
+import java.io.IOException;
+import java.io.Writer;
+
+public class CSVEscapeUtils extends CharSequenceTranslator {
```
garydgregory, line 8: This is confusing IMO, we already have an Apache Commons CSV component, we should not mix in public CSV code in Lang IMO. If this is absolutely needed it should be package-private but I am doubtful.

src/main/java/org/apache/commons/lang3/CSVEscapeUtils.java
garydgregory: This is confusing IMO, we already have an Apache Commons CSV component, we should not mix in public CSV code in Lang IMO. If this is absolutely needed it should be package-private but I am doubtful.

src/main/java/org/apache/commons/lang3/StringEscapeUtils.java
```diff
@@ -372,49 +333,6 @@ public int translate(final CharSequence input, final int index, final Writer out
...
-     *
-     * While {@link #unescapeCsv(String)} is the expected method of use, this
-     * object allows the CSV unescaping functionality to be used
-     * as the foundation for a custom translator.
-     *
-     * @since 3.0
-     */
-    public static final CharSequenceTranslator UNESCAPE_CSV = new CsvUnescaper();
-
-    static class CsvUnescaper extends CharSequenceTranslator {
```
garydgregory, line 386: This is a big -1. You cannot break binary compatibility. Run 'mvn' and watch for build failures.

garydgregory (changes requested):
I started to review but there are too many problems. One good aspect of this PR is that we should consider deprecate any CSV code in Commons Lang in favor of Commons CSV. This may require adding code to Commons CSV.

garydgregory:
Closing, no feedback.

PR #870, opened 2022-03-26
updating javadoc for NullPointerException when Validate.notNull() is called

dvmarcilio:
Hello! I updated some javadocs that mentioned `IllegalArgumentException` when an argument is `null` instead of `NullPointerException`. Tests were already checking for NPE, so I didn't write any.
This PR is similar to https://github.com/apache/commons-lang/pull/785.

This is reminiscent of https://github.com/apache/commons-lang/commit/ba607f525b842661d40195d0d4778528e2384e70, when lots of calls to `Validate.isTrue(arg != null)` (throws IAE) were changed to `Validate.notNull(arg)` (throws NPE).

The same wrong javadocs are also still present for `DateUtils` and `FieldUtils`.
I'll be glad to provide PRs for those if you feel it's needed. However, the diffs there will be much larger.
Thanks!

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharRange.java
+++ b/src/main/java/org/apache/commons/lang3/CharRange.java
@@ -184,7 +184,7 @@ final class CharRange implements Iterable<Character>, Serializable {
      *
      * @param range  the range to check against
      * @return {@code true} if this range entirely contains the input range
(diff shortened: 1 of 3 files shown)


garydgregory:
@dvmarcilio Thank you so much. More PRs are welcome.

PR #874, opened 2022-04-05
Add early null check MethodUtils.

arturobernalg:


src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -327,6 +327,7 @@ public static Object invokeExactMethod(final Object object, final String methodN
     public static Object invokeExactMethod(final Object object, final String methodName, Object[] args, Class<?>[] parameterTypes)
         throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
         Objects.requireNonNull(object, "object");
+        Objects.requireNonNull(object, "methodName");
```
aherbert, line 330: The null check should be on `methodName`

garydgregory: -1 You lose the better exception message below, specifically, you lose the class name. Follow the code path and you'll see that `getAccessibleMethod()` returns null if the `methodName` is null.

src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -422,13 +426,16 @@ public static Object invokeStaticMethod(final Class<?> cls, final String methodN
...
+     * @throws NullPointerException if the specified method or the cls is {@code null}.
      */
     public static Object invokeStaticMethod(final Class<?> cls, final String methodName,
             Object[] args, Class<?>[] parameterTypes)
             throws NoSuchMethodException, IllegalAccessException,
             InvocationTargetException {
         args = ArrayUtils.nullToEmpty(args);
         parameterTypes = ArrayUtils.nullToEmpty(parameterTypes);
+        Objects.requireNonNull(cls, "cls");
+        Objects.requireNonNull(methodName, "methodName");
```
garydgregory, line 438: -1 Same comment as above.

garydgregory (commented):
Hello @arturobernalg 
Thank you for your PR but...
-1 See comments.

PR #875, opened 2022-04-06
simplify operations and avoid extra calculations

arturobernalg:


src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
```diff
@@ -710,7 +710,7 @@ public static Method getMatchingAccessibleMethod(final Class<?> cls,
             final String parameterTypeName = lastParameterType==null ? null : lastParameterType.getName();
             final String parameterTypeSuperClassName = lastParameterType==null ? null : lastParameterType.getSuperclass().getName();
 
-            if (parameterTypeName!= null && parameterTypeSuperClassName != null && !methodParameterComponentTypeName.equals(parameterTypeName)
+            if (parameterTypeName!= null && !methodParameterComponentTypeName.equals(parameterTypeName)
```
garydgregory, line 713: Hi @arturobernalg 
Do you even need to check that `parameterTypeName` is not null here?

arturobernalg: HI @garydgregory 
No needed. we can safely remove 
TY

PR #878, opened 2022-04-07
Improvement conditions.

arturobernalg:
* Simplify conditions. --> `type` its always `!=null` otherwise NPE would be throw
* Fix javadoc.


arturobernalg:
HI @garydgregory 
Look god. TY

PR #887, opened 2022-04-30
Remove unnecessary code

o-x-e:
Remove unnecessary boxing, unboxing and casts.


garydgregory:
Hi @o-x-e 
-1: We want to see boxing and unboxing to make it obvious where NPEs may occur. We also want to show that there are method invocations taking place which should be considered when looking at the code from the performance angle.

PR #889, opened 2022-05-01
Added method to convert Object to String that support null values.

EACUAMBA:
I added a method to help people convert any object to String or return an empty String.

I added this methods because I was facing a problem with convert Wrap classes like Integer, Double to String.
I needed something to return empty string or a default value.

This is my contribute.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1589,6 +1589,59 @@ public class StringUtils {
     }
 
     /**
+     * <p>Returns either the passed in String,
+     * or if the String is {@code null}, an empty String ("").</p>
+     *
+     * <pre>
+     * StringUtils.defaultString(null)  = ""
+     * StringUtils.defaultString("")    = ""
+     * StringUtils.defaultString("bat") = "bat"
+     * StringUtils.defaultString(1) = "1"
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -1588,6 +1588,59 @@ public static String defaultString(final String str, final String nullDefault) {
...
+     *  if the input is {@code null}, may be null
+     * @return the passed in String, or the default if it was {@code null}
+     * @deprecated Use {@link Objects#toString(Object, String)}
+     */
+    @Deprecated
+    public static String defaultString(final Object object, final String nullDefault) {
+        if(object == null)
+            return defaultString(nullDefault);
+
+        return Objects.toString(object);
```
kinow, line 1641: Hi @EACUAMBA ,

I think instead of this new method, users could just use the Java 8+ `Objects.toString(Object o, String nullDefault)` - https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-

kinow (changes requested):

src/main/java/org/apache/commons/lang3/StringUtils.java
EACUAMBA: Hi @kinow,

You are right, the toString method  in Objects can work too.  Thanks.

PR #892, opened 2022-05-12
LANG-1680 Add support for standalone month formats

stevebosman-oc:

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -22,6 +22,7 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.Locale;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
@@ -1829,6 +1830,42 @@ public class DateUtils {
         public void remove() {
             throw new UnsupportedOperationException();
         }
+    }
+
+    /**
+     * Gets full standalone month names as used in "LLLL" date formatting.
+     * @param locale Locale
+     * @return Long names of months
+     */
+    static String[] getStandaloneLongMonths(final Locale locale) {
+        return getMonthNames(locale, Calendar.LONG_STANDALONE);
+    }
+
+    /**
+     * Gets short standalone month names as used in "LLLL" date formatting.
+     * @param locale Locale
+     * @return Short names of months
+     */
+    static String[] getStandaloneShortMonths(final Locale locale) {
+        return getMonthNames(locale, Calendar.SHORT_STANDALONE);
+    }
+
+    /**
+     * Gets month names in the requested style.
+     * @param locale Locale
+     * @param style Must be a valid {@link Calendar#getDisplayNames(int, int, Locale)} month style.
+     * @return Styled names of months
+     */
+    private static String[] getMonthNames(final Locale locale, final int style) {
+        // Unfortunately standalone month names are not available in DateFormatSymbols,
+        // so we have to extract them.
+        final Calendar calendar = Calendar.getInstance(locale);
+        Map<String, Integer> displayNames = calendar.getDisplayNames(Calendar.MONTH, style, locale);
+        final String[] monthNames = new String[displayNames.size()];
+        for (Map.Entry<String, Integer> entry: displayNames.entrySet()) {
+          monthNames[entry.getValue()] = entry.getKey();
+        }
+        return monthNames;
     }
 
 }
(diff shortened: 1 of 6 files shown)


stevebosman-oc:
Interesting. two unicode issues - both easy to fix, and one issue that seems to work on Java 11, but 8

stevebosman-oc:
After investigating I have determined the Java 8, 11, and 17 all return different standalone short months for September in German which makes my tests unstable

garydgregory:
Why not use code like `java.time.Month.APRIL.getDisplayName(TextStyle.FULL, Locale.GERMAN)`?

stevebosman-oc:
Is it safe to that only locales with 12 months are supported? In theory some locales can have thirteen months, although I am not sure there are any supported by Locales in the standard Java library.

garydgregory:
Safe? I'm not sure how this is a security issue but an enum seems "safer" than open ended code...

stevebosman-oc:
Unfortunately there appears to be a bug in Java 8 (I have 1.8_0332) around getDisplayName of standalone values
This 

    Month.MARCH.getDisplayName(TextStyle.FULL_STANDALONE, Locale.GERMAN)

returns "3" using Java 8, but using Java 11 it returns "März"

stevebosman-oc:
Odd, I seem to have encountered a version of https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8146356 but that claims to have been fixed in an earlier version of the 1.8 JDK than the one I'm running

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1831,4 +1831,40 @@ public void remove() {
...
+        // so we have to extract them.
+        final Calendar calendar = Calendar.getInstance(locale);
+        final int monthCount = calendar.getMaximum(Calendar.MONTH) + 1;
+        final String[] monthNames = new String[monthCount];
+        for (int i = 0; i < monthCount; i++) {
+            calendar.set(Calendar.MONTH, i);
+            monthNames[i] = calendar.getDisplayName(Calendar.MONTH, Calendar.SHORT_STANDALONE, locale);
+        }
+        return monthNames;
+    }
```
garydgregory, line 1868: These two methods are identical except for Calendar.SHORT_STANDALONE vs. Calendar.LONG_STANDALONE? Why duplicate? Javadoc: "Obtain..." -> "Gets..."

stevebosman-oc: sorry, I really should have noticed that before submitting. I'll refactor

stevebosman-oc: refactored into one method and changed javadoc. While doing this I looked at the Calendar source and Calendar#getDisplayNames seems to be more efficient than using the Calendar#getDisplayName method multiple times, so I am now using that.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1831,4 +1832,40 @@ public void remove() {
...
+     * Gets month names in the requested style.
+     * @param locale Locale
+     * @param style Must be a valid {@link Calendar#getDisplayNames(int, int, Locale)} month style.
+     * @return Styled names of months
+     */
+    private static String[] getMonthNames(final Locale locale, final int style) {
+        // Unfortunately standalone month names are not available in DateFormatSymbols,
+        // so we have to extract them.
+        final Calendar calendar = Calendar.getInstance(locale);
+        Map<String, Integer> displayNames = calendar.getDisplayNames(Calendar.MONTH, style, locale);
```
arturobernalg, line 1863: add final

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1831,4 +1832,40 @@ public void remove() {
...
+        final String[] monthNames = new String[displayNames.size()];
+        for (Map.Entry<String, Integer> entry: displayNames.entrySet()) {
```
arturobernalg, line 1865: `final Map.Entry<String, Integer> entry`

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
```diff
@@ -1694,5 +1694,43 @@ public void testWeekIterator() {
             now.add(Calendar.DATE, 1);
         }
     }
+
+    @Test
+    public void testGetStandaloneLongMonthNames() {
+        Locale testLocale = Locale.GERMAN;
+        String[] standaloneShortMonths = DateUtils.getStandaloneLongMonths(testLocale);
```
garydgregory, line 1701: Copy pasta typo: standaloneShortMonths -> standaloneLongMonths?

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
```diff
@@ -1694,5 +1694,43 @@ public void testWeekIterator() {
             now.add(Calendar.DATE, 1);
         }
     }
+
+    @Test
+    public void testGetStandaloneLongMonthNames() {
+        Locale testLocale = Locale.GERMAN;
```
garydgregory, line 1700: We don't need a local variable for the locale IMO.

PR #898, opened 2022-05-26
Increase test coverage of ComparableUtils from 71% to 100%

stevebosman-oc:
Static methods in ComparableUtils had 0 coverage this fixes that

Contributed on behalf of the @opencastsoftware open source team

PR #901, opened 2022-05-27
Fix unstable coverage of CharSequenceUtils tests noticed during merge of PRs 898 and 899

stevebosman-oc:
Line 36 in CharSequenceUtils usually returns true for at least one test in a test run, but occasionally (for instance in PR 898) it doesn't. This is because the majority of the tests rely on randomly generated strings.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/CharSequenceUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/CharSequenceUtilsTest.java
@@ -246,6 +246,10 @@ public class CharSequenceUtilsTest {
         testNewLastIndexOfSingle("apache", "x");
         testNewLastIndexOfSingle("oraoraoraora", "r");
         testNewLastIndexOfSingle("mudamudamudamuda", "d");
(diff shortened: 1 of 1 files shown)


garydgregory:
Hi @stevebosman-oc 

Thank you for your PR.

This report says there is no difference in code coverage: https://app.codecov.io/gh/apache/commons-lang/commit/618400c7d870308b14a09f2555270eb3193f194c

How did you notice the difference?

garydgregory:
Ah... I noticed your comment on random strings. Yeah, random tests are not great :-(

stevebosman-oc:
Yes, randomness can be a problem sometimes. 

I noticed in #898 the coverage dropped unexpectedly and then for #899 it came back. I then repeated the run a few times in my IDE and analysed the differences in the strings used on the runs that had slightly better coverage compared to the ones that didn't. Since making this change I have done several runs and the coverage has not dropped back down again so I have convinced myself that it is fixed.

PR #903, opened 2022-05-28
Add toCharacterObject(String,String) with default value.

arturobernalg:
Create a new method  that Converts the String to a Character using the first character, returning a
  default value if null  or empty is passed in.


garydgregory:
Hello @arturobernalg 
What is a use case for this?

arturobernalg:
> Hello @arturobernalg
> What is a use case for this?

HI @garydgregory 
for exactly the same as the ` public static Character toCharacterObject(final String str)` function but avoiding null return values,

src/main/java/org/apache/commons/lang3/CharUtils.java
```diff
@@ -115,6 +115,33 @@ public static Character toCharacterObject(final String str) {
...
+     * <p>Converts the String to a Character using the first character, returning a
+     *  default value if {@code null} or empty is passed in.</p>
+     *
+     * <p>For ASCII 7 bit characters, this uses a cache that will return the
+     * same Character object each time.</p>
+     *
+     * <pre>
+     *   CharUtils.toCharacterObject(null, 'A') = 'A'
+     *   CharUtils.toCharacterObject("", 'A')   = 'A'
+     *   CharUtils.toCharacterObject("A")  = 'A'
```
aherbert, line 128: These docs require updating too

arturobernalg: thank you @aherbert 
Fixed.

class101:
I was wondering why lang3 has a release per year. Really not motivating in participating this project, 110 pull request my god….

garydgregory:
You can also look up the word "volunteer"... ;-)

class101:
> You can also look up the word "volunteer"... ;-)

Volunteers to clear the pull request queue ?

kinow (changes requested):
I think rebasing will fix the conflict and also re-trigger GH actions that should now pass :crossed_fingers:

src/main/java/org/apache/commons/lang3/CharUtils.java
```diff
@@ -114,6 +114,33 @@ public static Character toCharacterObject(final String str) {
...
+     * @return the Character value of the first letter of the String or the default if null.
+     * @since 3.13.0
+     */
+    public static Character toCharacterObject(final String str, final Character defaultValue) {
+      final Character character = toCharacterObject(str);
+      if (character == null){
+          return defaultValue;
+      }
+      return character;
+    }
```
kinow, line 142: I think the rest of the code is using 4 spaces (you used 4 spaces in the test).

kinow (changes requested):
Thanks for rebasing it so quickly @arturobernalg ! :wave: 

One more review feedback. This also needs a JIRA issue. I think given the class has `toChar(ch, default)`, `toInt(ch, default)`, maybe it makes sense adding this method with a default value. The alternatives would be, I think, `Optional`, JDK 9+ `Objects.requireNonNullElse`, or using lang's `ObjectUtils.getIfNull` (e.g. `ObjectUtils.getIfNull(CharUtils.toCharacterObject(""), () -> 'B');`).

Cheers
-Bruno

src/main/java/org/apache/commons/lang3/CharUtils.java
```diff
@@ -114,6 +114,33 @@ public static Character toCharacterObject(final String str) {
...
+     * same Character object each time.</p>
+     *
+     * <pre>
+     *   CharUtils.toCharacterObject(null, 'A') = 'A'
+     *   CharUtils.toCharacterObject("", 'A')   = 'A'
+     *   CharUtils.toCharacterObject("A", 'D')  = 'A'
+     *   CharUtils.toCharacterObject("BA", 'D') = 'B'
+     * </pre>
+     *
+     * @param str  the character to convert
```
kinow, line 131: I think `str` is not a character too, sorry forgot to review the text in the javadocs.

kinow: Also worth checking if the docs are ending sentences with dot or without it. I think this `@param` doesn't use that, but the next ones does.

src/main/java/org/apache/commons/lang3/CharUtils.java
garydgregory: Better like this IMO: `return StringUtils.isEmpty(str) ? defaultValue : Character.valueOf(str.charAt(0));`

garydgregory: > I think if you still uses ternary conditions in 2023, you are a really bad developer. For the love of god, understand ternaries are unreadable and are a very bad practice to show junior developers.

The ternary expression is an uncontroversial feature of the Java language as well as of most programming languages. This is the first time I've ever read anyone express dismay at its usage. For a great overview of best practices, I recommend Josh Block's Effective Java.

PR #907, opened 2022-06-06
Increase test coverage of functional interfaces

stevebosman-oc:
Primarily ensuring static interface methods are covered


kinow (commented):
@stevebosman-oc thanks a lot for this pull request!

I had a look where else the code in this pull request is used, and it looks like there is a [FailableFunctionsTest.java](https://github.com/apache/commons-lang/blob/03b84cb3492938980b31d8cf15fdc6aeb553d063/src/test/java/org/apache/commons/lang3/function/FailableFunctionsTest.java) file where it covers multiple classes in that package.

Looking at the tests you added, it looks to me like it could be simplified, trying to have a single function that is applied to all the classes changed here testing the `nop` method.

But it looks like `FailableFunctionsTest` is not using this pattern yet, and instead has some duplication…

WDYT? Do you think your new tests could become a sort of parametrized method and applied to all these classes?

Thanks!

stevebosman-oc:
There is repetition, but because there isn't a common interface for each nop, and each instance requires different parameter types in the apply call I can't see a good way of using a ParameterisedTest, e.g.

```
assertNull(FailableBiFunction.nop().apply("Foo", "Bar"), "Expect NOP to return null");
assertNull(FailableDoubleFunction.nop().apply(Double.MAX_VALUE), "Expect NOP to return null");
assertNull(FailableIntFunction.nop().apply(Integer.MAX_VALUE), "Expect NOP to return null");
assertNull(FailableLongFunction.nop().apply(Long.MAX_VALUE), "Expect NOP to return null");
```

I could move the tests inside of FailableFunctionsTest either individually or bunched (similar to other tests in the class), e.g.

```
    @Test
    public void testFailablePrimitiveFunctionNops() throws Throwable {
        // NOP functions returning primitives should return default values and not throw exceptions
        assertEquals(0, FailableDoubleToIntFunction.nop().applyAsInt(Double.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableDoubleToLongFunction.nop().applyAsLong(Double.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableIntToDoubleFunction.nop().applyAsDouble(Integer.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableIntToLongFunction.nop().applyAsLong(Integer.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableLongToDoubleFunction.nop().applyAsDouble(Long.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableLongToIntFunction.nop().applyAsInt(Long.MAX_VALUE), "Expect NOP to return 0");
        assertEquals(0, FailableToIntFunction.nop().applyAsInt("Foo"), "Expect NOP to return 0");
        assertEquals(0, FailableToIntBiFunction.nop().applyAsInt("Foo", "Bar"), "Expect NOP to return 0");
        assertEquals(0, FailableToLongFunction.nop().applyAsLong("Foo"), "Expect NOP to return 0");
        assertEquals(0, FailableToLongBiFunction.nop().applyAsLong("Foo", "Bar"), "Expect NOP to return 0");
        assertEquals(0, FailableToDoubleFunction.nop().applyAsDouble("Foo"), "Expect NOP to return 0");
        assertEquals(0, FailableToDoubleBiFunction.nop().applyAsDouble("Foo", "Bar"), "Expect NOP to return 0");
    }

    @Test
    public void testFailableFunctionNops() throws Throwable {
        // NOP functions returning objects should return null and not throw exceptions
        assertNull(FailableBiFunction.nop().apply("Foo", "Bar"), "Expect NOP to return null");
        assertNull(FailableDoubleFunction.nop().apply(Double.MAX_VALUE), "Expect NOP to return null");
        assertNull(FailableIntFunction.nop().apply(Integer.MAX_VALUE), "Expect NOP to return null");
        assertNull(FailableLongFunction.nop().apply(Long.MAX_VALUE), "Expect NOP to return null");
    }

    @Test
    public void testFailableConsumerNops() throws Throwable {
        // NOP Consumers should not throw exceptions
        FailableConsumer.nop().accept("Foo");
        FailableDoubleConsumer.nop().accept(Double.MAX_VALUE);
        FailableIntConsumer.nop().accept(Integer.MAX_VALUE);
        FailableLongConsumer.nop().accept(Long.MAX_VALUE);
        FailableObjDoubleConsumer.nop().accept("Foo", Double.MAX_VALUE);
        FailableObjIntConsumer.nop().accept("Foo", Integer.MAX_VALUE);
        FailableObjLongConsumer.nop().accept("Foo", Long.MAX_VALUE);
    }
```

stevebosman-oc:
I don't really like tests which cover many different things, so I'll move them into FailableFunctionsTest as one test per Nop

stevebosman-oc:
@kinow I can squash if you're happy with the new changes, but I'd rather wait and see what you think first

PR #908, opened 2022-06-07
[LANG-1687] fix java Doc

TaeYongPark96:
The example of java document is incorrect.
StringUtils.substringBetween("yabczyabcz", "y", "z")   = "abc" => 
StringUtils.substringBetween("yabczyabcz", "y", "z")   = "abczyabc" =>


kinow (changes requested):
Hi @TaeYongPark96 

Have you tested the method? I just tried it using your branch, and got the following:

```java
    public static void main(String[] args) {
        String s = StringUtils.substringBetween("yabczyabcz", "y", "z");
        System.out.println(s);
    }

// outputs abc
```

Also, reading the top part of the javadoc, it says

> Only the first match is returned.

So I think the example was intentionally created to demonstrate that part of the method contract.

garydgregory:
@TaeYongPark96 ?

garydgregory:
Closing: -1, see org.apache.commons.lang3.StringUtilsSubstringTest.testSubstringBetween_StringStringString().

PR #911, opened 2022-06-12
fix StringUtils.upperCase(String str) java doc

k-nikhil:
This MR fixes a typo with the StringUtils.upperCase(String str) javadoc
Updated #lowerCase to #upperCase in scenarios where Locale needs to be provided for case transformation


kinow (approved):
Good spot! Thanks!!!

PR #912, opened 2022-06-12
Update ClassUtils Javadoc with some missing throws NPE

shalk:
method `toCanonicalName` use method  Validate.notNull  , so it could throw npe.

Meanswhile  `getClass`  use method `toCanonicalName` , so these method also throw npe. 

 https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/ClassUtils.java#L559

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -533,6 +533,7 @@ public class ClassUtils {
      * @param classLoader the class loader to use to load the class
      * @param className the class name
      * @return the class represented by {@code className} using the {@code classLoader}
+     * @throws NullPointerException if the className is null
      * @throws ClassNotFoundException if the class is not found
      */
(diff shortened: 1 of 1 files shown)


kinow (approved):
Thanks! I think these are fine, but leaving it so another committer can review it too.

shalk:
> 

I found some apis return null if argument is null  and the other apis check them throw npe.  they are different behave for argument.  
For example, `ClassUtils.getAllInterfaces` document ` {@code null} if null input` , so i think append `ClassUtils.getClass` with   `@throws NullPointerException if the className is null` is  meaningful. 

I aslo found  ClassUtilsTest have cover the npe case. So document can also cover it  and these document will not break code.

May be we should check which kind of  npe should declare in document.

kinow:
When reviewing this I also looked at other methods and found some documenting the behavior about null's and illegal argument/other runtime exceptions. But not sure if we want to document every case or not :+1:

PR #913, opened 2022-06-19
remove some duplicate assertion in StringUtilsTest

shalk:
keep it clean

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -305,7 +305,6 @@ public class StringUtilsTest {
         assertAbbreviateWithOffset("...ijklmno", 8, 10);
         assertAbbreviateWithOffset("...ijklmno", 9, 10);
         assertAbbreviateWithOffset("...ijklmno", 10, 10);
-        assertAbbreviateWithOffset("...ijklmno", 10, 10);
         assertAbbreviateWithOffset("...ijklmno", 11, 10);
         assertAbbreviateWithOffset("...ijklmno", 12, 10);
         assertAbbreviateWithOffset("...ijklmno", 13, 10);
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -1623,13 +1622,9 @@ public void testReCapitalize() {
     public void testRemove_char() {
         // StringUtils.remove(null, *)       = null
         assertNull(StringUtils.remove(null, 'a'));
-        assertNull(StringUtils.remove(null, 'a'));
-        assertNull(StringUtils.remove(null, 'a'));
 
         // StringUtils.remove("", *)          = ""
         assertEquals("", StringUtils.remove("", 'a'));
-        assertEquals("", StringUtils.remove("", 'a'));
-        assertEquals("", StringUtils.remove("", 'a'));
```
kinow, line 1632: This method, `testRemove_char`, appears to have been based off the `testRemove_String`. You are correct these three statements are duplicated and can be removed. But in `testRemove_String` there's one test case that must be added here as well:

```
assertNull(StringUtils.remove(null, null));
&
assertEquals("", StringUtils.remove("", null));
```

Can you add them, please, @shalk ?

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -2774,7 +2769,6 @@ public void testSplitPreserveAllTokens_StringChar() {
         assertEquals("b", res[1]);
         assertEquals("c", res[2]);
         assertEquals("", res[3]);
-        assertEquals("", res[3]);
```
kinow, line 2777: I believe this was supposed to be `assertEquals("", res[4]);`. The other tests have a similar case, where they verify the indexes from `0` to the last index available (`4` in this case). Can you change it, please?

kinow (changes requested):
There are a couple of places where I think the tests were supposed to have other values. Take a look and if you agree and have spare time, can you update them, please?

The other ones look correct to me. Great catch @shalk . Thanks for your pull request.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
shalk: done

kinow:
Merged. Thanks @shalk !!

PR #924, opened 2022-07-16
LANG-1638: Added docs regarding week year support

greatmastermario:
Added documentation indicating that DateFormatUtils does not support week year format YYYY like in java.util.Calendar, and it will default to yyyy current year format.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DateFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateFormatUtils.java
@@ -31,7 +31,8 @@ import java.util.TimeZone;
  * override any previously called Calendar.clear() calls. See LANG-755.</p>
  *
  * <p>Note that when using capital YYYY instead of lowercase yyyy, the formatter
- * will assume current year as week year is not supported.</p>
+ * will assume current year as week year is not supported. See {@link java.util.GregorianCalendar}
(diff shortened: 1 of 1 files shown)


garydgregory:
https://issues.apache.org/jira/browse/LANG-1638

garydgregory:
Hello community. Would someone review this, please?

kinow (commented):
From what I understood after reading the JIRA, this PR tries to avoid confusion by documenting the current behavior (much easier than trying to come up with a solution either backward-compatible or wait for a major release to change it).

It looks OK to me, but the text could be improved, I think. Maybe add an example of what's an YYYY and an yyyy expected to match against?

Thanks for the pull request @greatmastermario !

garydgregory:
Should the comment be at the class level instead of duplicated all over? A link to the class comment could also be used.

greatmastermario:
> It looks OK to me, but the text could be improved, I think. Maybe add an example of what's an YYYY and an yyyy expected to match against?

Thank you for the review. I have added a link tag to the Javadoc of GregorianCalendar which contains an explanation of week year. Hopefully that should work. I will add more if required.

> Should the comment be at the class level instead of duplicated all over? A link to the class comment could also be used.

Thank you for the suggestion. I agree, and I have moved all the method level docs up to the class level.

PR #930, opened 2022-08-05
Fixed incorrect documentation

Arav1229:
Methods throw null pointer exception but the javadocs did not reflect it. Fixed javadocs to address the issue.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
@@ -100,6 +100,7 @@ public class DiffBuilder<T> implements Builder<DiffResult<T>> {
      *            test is enabled and returns true.  The result of this test
      *            is never changed throughout the life of this {@link DiffBuilder}.
      * @throws IllegalArgumentException
+     * @throws NullPointerException
      *             if {@code lhs} or {@code rhs} is {@code null}
      * @since 3.4
      */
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/lang3/builder/DiffBuilder.java
```diff
@@ -100,6 +100,7 @@
      *            test is enabled and returns true.  The result of this test
      *            is never changed throughout the life of this {@link DiffBuilder}.
      * @throws IllegalArgumentException
+     * @throws NullPointerException
```
garydgregory, line 103: Indentation does not seem to match what's above.

src/main/java/org/apache/commons/lang3/builder/HashCodeBuilder.java
```diff
@@ -352,7 +352,8 @@ public static int reflectionHashCode(final int initialNonZeroOddNumber, final in
      *             if the Object is {@code null}
      * @throws IllegalArgumentException
      *             if the number is zero or even
-     *
+     * @throws NullPointerException
+     *			   if the Object is {@code null}
```
garydgregory, line 356: Indentation does not seem to match what's above. Name the parameter, not a generic type like Object.

garydgregory (changes requested):
@Arav1229 Thank you for your PR, please see comments.

garydgregory:
I'm not sure we should document this, otherwise, most if not all APIs would need to say they can throw NPE.

garydgregory:
No reply, closing.

PR #933, opened 2022-08-10
Lang 1689 add optional to objectutils isempty

hendrixjoseph:
* Original JIRA ticket (not by me): https://issues.apache.org/jira/projects/LANG/issues/LANG-1689
* Method `isEmpty` in `ObjectUtils` was modified to return the negation of `Optional.isPresent` (i.e. `!optional.isPresent()`) when the passed in object is an Optional.
* The object contained within the Optional is not checked if it is also empty. This gives the caller the option to check if it's present (or not) on their own.
  * For instance, the caller might want to do something like the following:
    `ObjectUtils.isEmpty(optional) && ObjectUtils.isNotEmpty(optional.get())`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -1020,6 +1020,7 @@ public class ObjectUtils {
      * <li>{@link Array}: Considered empty if its length is zero.</li>
      * <li>{@link Collection}: Considered empty if it has zero elements.</li>
      * <li>{@link Map}: Considered empty if it has zero key-value mappings.</li>
+     * <li>{@link Optional}: Considered empty if {@link Optional#isPresent} returns false, regardless of the "emptiness" of the contents.</li>
      * </ul>
      *
      * <pre>
@@ -1029,6 +1030,9 @@ public class ObjectUtils {
      * ObjectUtils.isEmpty(new int[]{})      = true
      * ObjectUtils.isEmpty(new int[]{1,2,3}) = false
      * ObjectUtils.isEmpty(1234)             = false
+     * ObjectUtils.isEmpty(1234)             = false
+     * ObjectUtils.isEmpty(Optional.of(""))  = false
+     * ObjectUtils.isEmpty(Optional.empty()) = true
      * </pre>
      *
      * @param object  the {@link Object} to test, may be {@code null}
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -1051,6 +1052,10 @@ public static boolean isEmpty(final Object object) {
         if (object instanceof Map<?, ?>) {
             return ((Map<?, ?>) object).isEmpty();
         }
+        if (object instanceof Optional<?>) {
+        	return !((Optional<?>) object).isPresent();
+        }
+
```
garydgregory, line 1058: What if it's an Optional with an empty string in it?

hendrixjoseph: The optional would be non-empty. Take the following code snippet:
```java
Optional<String> optional = Optional.of("");
boolean empty = ObjectUtils.isEmpty(optional);
// empty is false
```
In the case of Optionals, we're concerned whether the optional is empty, not if the contents of the optional is empty.

This is similar to having an array or collection of length 1 that only contains an empty string. In that case, ObjectUtils.isEmpty would still return false.

I can update the JavaDoc to make it more clear.

(And unrelated to this conversation, I notices I accidently slipped a couple tabs instead of spaces in. I'll fix that shortly.)

garydgregory:
Hm, at first I thought this would be OK but now I'm not sure and would like community feedback:
- There are APIs im the JRE called `isEmpty()` on various classes that usually return true when `size() == 0`; for example `Collection`, `String`, and `Map`.
- There _isn't_ an `Optional` API called `isEmpty`, instead, there is a different concept represented by `isPresent()`; which returns true when the wrapped value is non-null.

So it feels like this PR conflates the two concepts.

hendrixjoseph:
I create another PR (#934) that uses the approach mentioned in my previous comment (using the non-recursive way) instead of looking at the optional itself.

I'm unsure which is the "better" or "preferred" way and I agree more community input is needed.

garydgregory:
Hi All:

The addition of Optional#isEmpty() in Java 11 seems to be an argument in favor of supporting Optional in our isEmpty() method, I think we need to consider if it makes sense, as opposed to just being doable. IOW, is a wrapper of a null value "empty"? Is a wrapper of an empty String empty? Is a wrapper of an empty Collection empty?

Using recursion seems to be a recipe for confusion: Why recurse into one kind of object and not another? IOW, why recuse into Optional and not into Collections and Maps?

What do others think? 

I do not think we need to consider an implementation vs. another until we can give answers and provide a consistent picture.

PR #934, opened 2022-08-16
Lang 1689 add optional to objectutils isempty with unpacking the optional

hendrixjoseph:
Note: This PR is related to #933.

Each is a different approach to adding an Optional to ObjectUtils.isEmpty. At most one should me merged.

This approach "unwraps" the optional so that the contents are checked if they are empty or not.


garydgregory:
Merged #933

PR #935, opened 2022-08-20
LANG-1603: Deprecate Fraction class

rednikeeg:
jira issue: https://issues.apache.org/jira/browse/LANG-1603

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -31,8 +31,10 @@ import org.apache.commons.lang3.Validate;
  * based and thus suffers from various overflow issues. For a BigInteger based
  * equivalent, please see the Commons Math BigFraction class. </p>
  *
+ * @deprecated for fraction implementation from commons-numbers
  * @since 2.0
  */
+@Deprecated
 public final class Fraction extends Number implements Comparable<Fraction> {
 
     /**


src/main/java/org/apache/commons/lang3/math/Fraction.java
```diff
@@ -31,8 +31,10 @@
  * based and thus suffers from various overflow issues. For a BigInteger based
  * equivalent, please see the Commons Math BigFraction class. </p>
  *
+ * @deprecated for fraction implementation from commons-numbers
```
garydgregory, line 34: Please add a URL to the Javadoc for the favored implementation.

src/main/java/org/apache/commons/lang3/math/Fraction.java
```diff
@@ -31,7 +31,8 @@
  * based and thus suffers from various overflow issues. For a BigInteger based
  * equivalent, please see the Commons Math BigFraction class. </p>
  *
- * @deprecated for fraction implementation from commons-numbers
+ * @deprecated for <a href="https://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/index.html">Fraction</a>
+ * from commons-numbers
```
garydgregory, line 35: Still not quite right: You talk about Commons Numbers but link to Commons Math as a whole. You want the URL to the Commons Numbers Fraction package or class.

src/main/java/org/apache/commons/lang3/math/Fraction.java
rednikeeg: Added reference to the favored implementation

src/main/java/org/apache/commons/lang3/math/Fraction.java
rednikeeg: Aw, sorry, resolved that moment already

garydgregory:
The discussion in https://issues.apache.org/jira/browse/LANG-1603 does not make it clear (to me) if the proposed replacement is convenient enough.

src/main/java/org/apache/commons/lang3/math/Fraction.java
```diff
@@ -31,8 +31,11 @@
  * based and thus suffers from various overflow issues. For a BigInteger based
  * equivalent, please see the Commons Math BigFraction class. </p>
  *
+ * @deprecated for <a href="https://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/fraction/Fraction.html">Fraction</a>
```
garydgregory, line 34: We can't just deprecate the class without documenting how to port your current code to the other API IMO. For example, what do you do with `Fraction#invert()`? Each method must be documented, otherwise, it's a disservice, and everyone using the class is left having to do their own investigation and mapping.

garydgregory (changes requested):

PR #936, opened 2022-08-20
LANG-1604: Deprecate RandomUtils class

rednikeeg:
jira issue - https://issues.apache.org/jira/browse/LANG-1604

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomUtils.java
@@ -30,8 +30,10 @@ import java.util.concurrent.ThreadLocalRandom;
  * a better choice for applications with more stringent requirements
  * (performance and/or correctness).</p>
  *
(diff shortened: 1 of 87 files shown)


src/main/java/org/apache/commons/lang3/RandomUtils.java
```diff
@@ -30,8 +30,10 @@
  * a better choice for applications with more stringent requirements
  * (performance and/or correctness).</p>
  *
+ * @deprecated for more optimized implementation from commons-rng
```
garydgregory, line 33: Please add a URL to the Javadoc for the favored implementation.

rednikeeg: Sorry, I don't understand. What should I do?

rednikeeg: Added reference

PR #938, opened 2022-08-20
Fix broken javadoc.

arturobernalg:


kinow (approved):

src/main/java/org/apache/commons/lang3/math/Fraction.java
```diff
@@ -517,7 +517,7 @@ public Fraction abs() {
      * <p>The returned fraction is in reduced form.</p>
      *
      * @param power  the power to raise the fraction to
-     * @return {@code this} if the power is one, {@link ONE} if the power
+     * @return {@code this} if the power is one, {@link Fraction#ONE} if the power
```
garydgregory, line 520: You can just say `{@link #ONE}` IIRC, you don't need to refer to your own class name.

garydgregory (changes requested):
Hi @arturobernalg 
Thank you for your PR. Please check my comment.

src/main/java/org/apache/commons/lang3/math/Fraction.java
arturobernalg: true. Changed.

PR #949, opened 2022-09-15
LANG-1691: ClassUtils.getShortCanonicalName - using canonicalName rat…

thiyagu-7:
JIRA issue - https://issues.apache.org/jira/browse/LANG-1691

The `getShortCanonicalName` is using class name rather than the canonical name. Fixing it by using the canonicalName.
Also updated the unit tests because of this change. For some of the tests, the canonical name will be null and hence changed the expected value to be an empty string. For inner class, the output will not have the outer class name.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
@@ -517,13 +517,12 @@ public class ClassUtilsTest extends AbstractLangTest {
         class Named {
             // empty
         }
-        // WARNING: this is fragile, implementation may change, naming is not guaranteed
-        assertEquals("ClassUtilsTest.8", ClassUtils.getShortCanonicalName(new Object() {
+        assertEquals("", ClassUtils.getShortCanonicalName(new Object() {
             // empty
         }.getClass()));
         // WARNING: this is fragile, implementation may change, naming is not guaranteed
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/ClassUtilsTest.java
```diff
@@ -540,12 +539,11 @@ public void test_getShortCanonicalName_Object() {
...
             // empty
         }
-        // WARNING: this is fragile, implementation may change, naming is not guaranteed
-        assertEquals("ClassUtilsTest.9", ClassUtils.getShortCanonicalName(new Object() {
+        assertEquals("", ClassUtils.getShortCanonicalName(new Object() {
             // empty
         }, "<null>"));
-        assertEquals("ClassUtilsTest.9Named", ClassUtils.getShortCanonicalName(new Named(), "<null>"));
-        assertEquals("ClassUtilsTest.Inner", ClassUtils.getShortCanonicalName(new Inner(), "<null>"));
+        assertEquals("", ClassUtils.getShortCanonicalName(new Named(), "<null>"));
```
garydgregory, line 545: This seems obviously no good to me because "Named" is a real concrete class defined in this very file, it's name is not "". What am I missing?

garydgregory: @thiyagu-7 
Thank you for the clarification!

PR #951, opened 2022-09-16
[LANG-1692] Cast FieldUtils.readField result to the recipient type

tisonkun:
This patch should be easy to understand.

Generally, it won't affect user cases as long as the type can be inferred by the recipient or JDK rules. There's a case as shown in `ReflectionDiffBuilder.java` that if the type inference results in ambiguity, it may cause compile error.

Thus, I'm unsure whether this is proper to be included with a minor version bump. I don't think adding another series of methods helps - it causes further confusion.

This patch doesn't cause extra class cast exception because if the recipient is `Object` the cast should always succeed. And if the recipient already cast it into another type, there will be an exception if the cast cannot be performed.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -110,8 +110,9 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
         for (final Field field : FieldUtils.getAllFields(clazz)) {
             if (accept(field)) {
                 try {
-                    diffBuilder.append(field.getName(), readField(field, left, true),
-                            readField(field, right, true));
+                    Object leftObject = readField(field, left, true);
+                    Object rightObject = readField(field, right, true);
+                    diffBuilder.append(field.getName(), leftObject, rightObject);
                 } catch (final IllegalAccessException ex) {
                     //this can't happen. Would get a Security exception instead
                     //throw a runtime exception in case the impossible happens.
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -110,9 +110,8 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
         for (final Field field : FieldUtils.getAllFields(clazz)) {
             if (accept(field)) {
                 try {
-                    Object leftObject = readField(field, left, true);
-                    Object rightObject = readField(field, right, true);
-                    diffBuilder.append(field.getName(), leftObject, rightObject);
+                    diffBuilder.append(field.getName(), readField(field, left, true),
+                            readField(field, right, true));
                 } catch (final IllegalAccessException ex) {
                     //this can't happen. Would get a Security exception instead
                     //throw a runtime exception in case the impossible happens.
--- a/src/main/java/org/apache/commons/lang3/reflect/FieldUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/FieldUtils.java
@@ -254,8 +254,8 @@ public class FieldUtils {
     /**
      * Reads an accessible {@code static} {@link Field}.
      *
-     * @param field
-     *            to read
+     * @param <T> the recipient type
+     * @param field to read
      * @return the field value
      * @throws IllegalArgumentException
      *             if the field is {@code null}, or not {@code static}
@@ -269,8 +269,8 @@ public class FieldUtils {
     /**
      * Reads a static {@link Field}.
      *
-     * @param field
-     *            to read
+     * @param <T> the recipient type
+     * @param field to read
      * @param forceAccess
      *            whether to break scope restrictions using the
      *            {@link java.lang.reflect.AccessibleObject#setAccessible(boolean)} method.
@@ -289,6 +289,7 @@ public class FieldUtils {
     /**
      * Reads the named {@code public static} {@link Field}. Superclasses will be considered.
      *
+     * @param <T> the recipient type
      * @param cls
      *            the {@link Class} to reflect, must not be {@code null}
      * @param fieldName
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -110,8 +110,9 @@ private void appendFields(final Class<?> clazz) {
         for (final Field field : FieldUtils.getAllFields(clazz)) {
             if (accept(field)) {
                 try {
-                    diffBuilder.append(field.getName(), readField(field, left, true),
-                            readField(field, right, true));
+                    Object leftObject = readField(field, left, true);
```
garydgregory, line 113: Is this change needed for the type changes? Please revert it otherwise.

garydgregory:
This PR has no matching tests.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
tisonkun: Please read the description:

> There's a case as shown in `ReflectionDiffBuilder.java` that if the type inference results in ambiguity, it may cause compile error.

tisonkun:
> This PR has no matching tests.

Please read the description:

> * it won't affect user cases as long as the type can be inferred by the recipient or JDK rules
> * This patch doesn't cause extra class cast exception ...

No existing tests break is fine. I don't know how to add a new effective test.

tisonkun:
I may write something like:

```diff
diff --git a/src/test/java/org/apache/commons/lang3/reflect/FieldUtilsTest.java b/src/test/java/org/apache/commons/lang3/reflect/FieldUtilsTest.java
index b68a994f3..dcecbca06 100644
--- a/src/test/java/org/apache/commons/lang3/reflect/FieldUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/reflect/FieldUtilsTest.java
@@ -480,7 +480,8 @@ public void testReadDeclaredNamedStaticFieldForceAccess() throws Exception {
     @Test
     public void testReadField() throws Exception {
         final Field parentS = FieldUtils.getDeclaredField(parentClass, "s");
-        assertEquals("s", FieldUtils.readField(parentS, publicChild));
+        String s = FieldUtils.readField(parentS, publicChild);
+        assertEquals("s", s);
         assertEquals("s", FieldUtils.readField(parentS, publiclyShadowedChild));
         assertEquals("s", FieldUtils.readField(parentS, privatelyShadowedChild));
         final Field parentB = FieldUtils.getDeclaredField(parentClass, "b", true);
```

to prove this change works. But it seems useless.

garydgregory:
You need to think about how to avoid regressions in case the change is reverted. IOW, in this case, you should get a compilation error.

tisonkun:
@garydgregory Thanks for your comments. Will push a followup later this week.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
garydgregory: @tisonkun 
I still don't get it, why is this needed in this PR as it relates to the generics change?

src/main/java/org/apache/commons/lang3/reflect/FieldUtils.java
```diff
@@ -254,23 +254,23 @@ public static List<Field> getFieldsListWithAnnotation(final Class<?> cls, final
     /**
      * Reads an accessible {@code static} {@link Field}.
      *
-     * @param field
```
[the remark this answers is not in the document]
tisonkun: Reverting...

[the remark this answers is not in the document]
tisonkun: Oops. I think I should add the `@param <T>` line so adjusting neighbor lines should be OK. Said you're updating the comments, and adjusting a bit the layout should not be a big deal. I don't touch unrelated comments.

tisonkun:
Tests added.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
tisonkun: image

tisonkun:
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/chenzili/Brittani/commons-lang/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java:[113,32] reference to append is ambiguous
  both method append(java.lang.String,short[],short[]) in org.apache.commons.lang3.builder.DiffBuilder and method append(java.lang.String,java.lang.Object[],java.lang.Object[]) in org.apache.commons.lang3.builder.DiffBuilder match
```

tisonkun:
@garydgregory is there any remaining blocker for this patch?

garydgregory:
I'll take a look this weekend...

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
garydgregory: Revert this change, please.

tisonkun: Revert this change will cause compile to fail. I can do it and you see what happens.

tisonkun: Reverted at bc4e02229407a80b1315c8c5f00cad859724aa4e.

Now the compilation failed: https://github.com/apache/commons-lang/actions/runs/3210472599/jobs/5247938888.

garydgregory: Does that mean that existing call sites in users' applications will also fail?

tisonkun: If you think it breaks the existing code and should not do it, you can reject the PR. But I'm a bit confused since reverting it obviously causes compile to fail.

tisonkun:
Closed as no response. I'll add wrappers for this changes downstream when necessary.

PR #953, opened 2022-09-21
Handle error messages with special characters.

stawirej:
Currently Validate.isTrue(false, MESSAGE) when MESSAGE  contains special characters, which are part of the message and should not be interpreted as the placeholders for parameters used in String.format() it throws e.g. UnknownFormatConversionException when desired behaviour should be to throw IllegalArgumentException with message containing special characters.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Validate.java
+++ b/src/main/java/org/apache/commons/lang3/Validate.java
@@ -103,7 +103,7 @@ public class Validate {
      */
     public static void isTrue(final boolean expression, final String message, final long value) {
         if (!expression) {
-            throw new IllegalArgumentException(String.format(message, Long.valueOf(value)));
+            throw new IllegalArgumentException(getMessage(message, Long.valueOf(value)));
         }
     }
 
@@ -128,7 +128,7 @@ public class Validate {
      */
     public static void isTrue(final boolean expression, final String message, final double value) {
         if (!expression) {
-            throw new IllegalArgumentException(String.format(message, Double.valueOf(value)));
+            throw new IllegalArgumentException(getMessage(message, Double.valueOf(value)));
         }
     }
 
@@ -139,8 +139,7 @@ public class Validate {
      * primitive number or using your own custom validation expression.
      *
      * <pre>
-     * Validate.isTrue(i &gt;= min &amp;&amp; i &lt;= max, "The value must be between &#37;d and &#37;d", min, max);
-     * Validate.isTrue(myObject.isOk(), "The object is not okay");</pre>
+     * Validate.isTrue(i &gt;= min &amp;&amp; i &lt;= max, "The value must be between &#37;d and &#37;d", min, max);</pre>
      *
      * @param expression  the boolean expression to check
      * @param message  the {@link String#format(String, Object...)} exception message if invalid, not null
@@ -152,7 +151,7 @@ public class Validate {
      */
     public static void isTrue(final boolean expression, final String message, final Object... values) {
         if (!expression) {
-            throw new IllegalArgumentException(String.format(message, values));
+            throw new IllegalArgumentException(getMessage(message, values));
         }
     }
 
@@ -217,7 +216,7 @@ public class Validate {
      * @see Objects#requireNonNull(Object)
      */
     public static <T> T notNull(final T object, final String message, final Object... values) {
-        return Objects.requireNonNull(object, () -> String.format(message, values));
+        return Objects.requireNonNull(object, () -> getMessage(message, values));
     }
 
     /**
@@ -237,9 +236,9 @@ public class Validate {
      * @see #notEmpty(Object[])
      */
     public static <T> T[] notEmpty(final T[] array, final String message, final Object... values) {
-        Objects.requireNonNull(array, () -> String.format(message, values));
+        Objects.requireNonNull(array, () -> getMessage(message, values));
         if (array.length == 0) {
-            throw new IllegalArgumentException(String.format(message, values));
+            throw new IllegalArgumentException(getMessage(message, values));
         }
         return array;
     }
@@ -281,9 +280,9 @@ public class Validate {
      * @see #notEmpty(Object[])
      */
     public static <T extends Collection<?>> T notEmpty(final T collection, final String message, final Object... values) {
-        Objects.requireNonNull(collection, () -> String.format(message, values));
+        Objects.requireNonNull(collection, () -> getMessage(message, values));
         if (collection.isEmpty()) {
-            throw new IllegalArgumentException(String.format(message, values));
+            throw new IllegalArgumentException(getMessage(message, values));
         }
         return collection;
     }
@@ -325,9 +324,9 @@ public class Validate {
      * @see #notEmpty(Object[])
      */
     public static <T extends Map<?, ?>> T notEmpty(final T map, final String message, final Object... values) {
-        Objects.requireNonNull(map, () -> String.format(message, values));
+        Objects.requireNonNull(map, () -> getMessage(message, values));
         if (map.isEmpty()) {
-            throw new IllegalArgumentException(String.format(message, values));
+            throw new IllegalArgumentException(getMessage(message, values));
         }
         return map;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -156,6 +156,30 @@ public static void isTrue(final boolean expression, final String message, final
...
+     *
+     * @param expression  the boolean expression to check
+     * @throws IllegalArgumentException if expression is {@code false}
+     * @see #isTrue(boolean, String, long)
+     * @see #isTrue(boolean, String, double)
+     * @see #isTrue(boolean, String, Object...)
+     */
+    public static void isTrue(final boolean expression, final String message) {
+        if (!expression) {
+            throw new IllegalArgumentException(String.valueOf(message));
```
aherbert, line 179: `String.valueOf(String)` is not required. Note that String.valueOf(null) will return null and that is handled fine by the IllegalArgumentException constructor. So just pass through the String message argument (which can be null).

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -156,6 +156,30 @@ public static void isTrue(final boolean expression, final String message, final
...
+     * to an arbitrary boolean expression, such as validating a
+     * primitive number or using your own custom validation expression.
+     *
+     * <pre>
+     * Validate.isTrue(i &gt; 0, "The value must be greater than 0");
+     * Validate.isTrue(myObject.isOk(), "The object is not okay");</pre>
+     *
+     * <p>The exception message will be converted to a string using String.valueOf(Object).</p>
+     *
+     * @param expression  the boolean expression to check
```
aherbert, line 171: There is no `@param` tag for `message`.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -156,6 +156,30 @@ public static void isTrue(final boolean expression, final String message, final
...
+     * Validate that the argument condition is {@code true}; otherwise
+     * throwing an exception. This method is useful when validating according
+     * to an arbitrary boolean expression, such as validating a
+     * primitive number or using your own custom validation expression.
+     *
+     * <pre>
+     * Validate.isTrue(i &gt; 0, "The value must be greater than 0");
+     * Validate.isTrue(myObject.isOk(), "The object is not okay");</pre>
+     *
+     * <p>The exception message will be converted to a string using String.valueOf(Object).</p>
```
aherbert, line 169: This is not required.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -156,6 +156,30 @@ public static void isTrue(final boolean expression, final String message, final
         }
     }
 
+    /**
+     * Validate that the argument condition is {@code true}; otherwise
+     * throwing an exception. This method is useful when validating according
```
aherbert, line 161: `an exception with the specified message`

Note: This change makes the code example on line 143 invalid. Please update the javadoc in the preceding method to remove:
```Validate.isTrue(myObject.isOk(), "The object is not okay");</pre>```

**Review by aherbert — CHANGES_REQUESTED** [review]
Adding this method will change source compatibility to use this method in place of the varargs method when there are no supplied args. This should be reflected in a javadoc change to that method.

File: src/main/java/org/apache/commons/lang3/Validate.java (hunk 1 above)
Comment (stawirej) [reply src/main/java/org/apache/commons/lang3/Validate.java]: Done.

File: src/main/java/org/apache/commons/lang3/Validate.java (hunk 2 above)
Comment (stawirej) [reply src/main/java/org/apache/commons/lang3/Validate.java:169]: Done.

File: src/main/java/org/apache/commons/lang3/Validate.java (hunk 3 above)
Comment (stawirej) [reply src/main/java/org/apache/commons/lang3/Validate.java]: Done.

File: src/main/java/org/apache/commons/lang3/Validate.java (hunk 4 above)
Comment (stawirej) [reply src/main/java/org/apache/commons/lang3/Validate.java:160]: Done.

**Comment by stawirej** [comment]
> Adding this method will change source compatibility to use this method in place of the varargs method when there are no supplied args. This should be reflected in a javadoc change to that method.

Done. Could you please review if javadoc is updated as you had in mind?

**Comment by garydgregory** [comment]
New public and protected APIs must have Javadoc since tags.

File: src/main/java/org/apache/commons/lang3/Validate.java [hunk 5]
```diff
@@ -156,6 +155,29 @@ public static void isTrue(final boolean expression, final String message, final
...
+     * Validate.isTrue(i &gt; 0, "The value must be greater than 0");
+     * Validate.isTrue(myObject.isOk(), "The object is not okay");</pre>
+     *
+     * @param expression  the boolean expression to check
+     * @param message  the exception message if invalid
+     * @throws IllegalArgumentException if expression is {@code false}
+     * @see #isTrue(boolean, String, long)
+     * @see #isTrue(boolean, String, double)
+     * @see #isTrue(boolean, String, Object...)
+     */
```
aherbert, line 174: Add a `@Since 3.13.0`

garydgregory:
Why just this one method and not all of the message/vargs methods in the Validate class? 
Alternatively, we could deprecate the class and move it to Commons Validator and complete the task there.

aherbert:
IIUC the fix should remove the new public method and then add a private function:
```Java
private static String getMessage(final String message, final Object... values) {
    return values.length == 0 ? message : String.format(message, values);
}
```
for use as:
```Java
public static void isTrue(final boolean expression, final String message, final Object... values) {
    if (!expression) {
        throw new IllegalArgumentException(getMessage(message, values));
    }
}
```
This can be used in all call sites that have a varargs.

garydgregory:
> IIUC the fix should remove the new public method and then add a private function:
> 
> ```java
> private static String getMessage(final String message, final Object... values) {
>     return values.length == 0 ? message : String.format(message, values);
> }
> ```
> 
> for use as:
> 
> ```java
> public static void isTrue(final boolean expression, final String message, final Object... values) {
>     if (!expression) {
>         throw new IllegalArgumentException(getMessage(message, values));
>     }
> }
> ```
> 
> This can be used in all call sites that have a varargs.

Clever! 👍😀
And call it from the other vararg methods.

stawirej:
> > IIUC the fix should remove the new public method and then add a private function:
> > ```java
> > private static String getMessage(final String message, final Object... values) {
> >     return values.length == 0 ? message : String.format(message, values);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for use as:
> > ```java
> > public static void isTrue(final boolean expression, final String message, final Object... values) {
> >     if (!expression) {
> >         throw new IllegalArgumentException(getMessage(message, values));
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This can be used in all call sites that have a varargs.
> 
> Clever! 👍😀 And call it from the other vararg methods.

Done. Please review.

src/main/java/org/apache/commons/lang3/Validate.java
```diff
@@ -1265,7 +1264,11 @@ public static void isAssignableFrom(final Class<?> superType, final Class<?> typ
     public static void isAssignableFrom(final Class<?> superType, final Class<?> type, final String message, final Object... values) {
         // TODO when breaking BC, consider returning type
         if (!superType.isAssignableFrom(type)) {
-            throw new IllegalArgumentException(String.format(message, values));
+            throw new IllegalArgumentException(getMessage(message, values));
         }
     }
+
+    private static String getMessage(final String message, final Object... values) {
```
aherbert, line 1271: This all looks fine. Please add a comment to the private method, e.g. 
```
Get the message using {@link String#format(String, Object...) String.format(message, values)}
if the values are not empty, otherwise return the message unformatted.
This method exists to allow validation methods declaring a String message and varargs parameters
to be used without any message parameters when the message contains special characters,
e.g. {@code Validate.isTrue(false, "%Failed%")}.
@param ...
@return ...
```

stawirej: Done.

aherbert:
@stawirej Thanks for the contribution

stawirej:
@aherbert  Thank you for your help! When do you plan to release it?

aherbert:
@stawirej I am not sure of the release schedule. There are a lot of fixes in lang 3.13-SNAPSHOT so a release may be done within a few months.

garydgregory:
Right, before the end of the year for certain, much before that I hope. I have a couple of other releases to push through first.

php-coder:
This change was backward incompatible by the way. Previously, in order to have `%` character in a message, we doubled it (`%%`). After update, both percentage are shown, so the (trivial) changes in code are required.

PR #956, opened 2022-09-24
Improve JavaDocs 

arturobernalg:
Jus add a bunch of javadoc improvement in order to make it more readable when the user search information.


arturobernalg:
BTW sorry the size of the PR. But IMO it was necessary

src/main/java/org/apache/commons/lang3/CharSetUtils.java
```diff
@@ -45,8 +45,8 @@ public class CharSetUtils {
      * </pre>
      *
      * @see CharSet#getInstance(String...) for set-syntax.
-     * @param str  String to look for characters in, may be null
-     * @param set  String[] set of characters to identify, may be null
+     * @param str  String to look for characters in, may be {@code null}
```
garydgregory, line 48: Is there a reason why "String" is sometimes linked and sometimes not?

garydgregory:
You really need to build the project locally before you create the PR, see the Javadoc errors.

Aside from that, I'm starting to think that this is really making the code harder to read and that it would be better handled by a Doclet.

src/main/java/org/apache/commons/lang3/CharSetUtils.java
arturobernalg: No Reason at all.

arturobernalg:
HI @garydgregory 

> You really need to build the project locally before you create the PR, see the Javadoc errors. 
> 
> Aside from that, I'm starting to think that this is really making the code harder to read and that it would be better handled by a Doclet.

I lose track of what apache jenkings does. I do not have access

`[INFO] Configuration changed, re-generating javadoc.
[WARNING] Javadoc Warnings
[WARNING] warning: URL https://docs.oracle.com/javase/8/docs/api/element-list was redirected to https://docs.oracle.com/en/java/javase/19/ -- Update the command-line options to suppress this warning.
[WARNING] 1 warning
[INFO] BUILD SUCCESS
[INFO] Total time:  03:28 min
[INFO] Finished at: 2022-09-24T21:01:26+02:00
 abernal@Arturo-MacBook-Pro  ~/IdeaProyect/commons-lang   feature/better_javadoc `

> You really need to build the project locally before you create the PR, see the Javadoc errors.

None. 

> 
> Aside from that, I'm starting to think that this is really making the code harder to read and that it would be better handled by a Doclet.

garydgregory:
I'm not sure what Jenkins has to do with this. In any case, you can see the failure here:

https://github.com/apache/commons-lang/actions/runs/3119069043/jobs/5058792919

aherbert:
FYI: This javadoc warning:
```
[WARNING] Javadoc Warnings
[WARNING] warning: URL https://docs.oracle.com/javase/8/docs/api/element-list was
redirected to https://docs.oracle.com/en/java/javase/19/
 -- Update the command-line options to suppress this warning.
```
Is due to the `<links>` added to the `maven-javadoc-plugin` in commons-parent for JDK classes. This seems to not be required when building on recent JDK LTS versions (tested on 8, 11, and 17). Only JDK 17 puts out this warning as it attempts to look up an element-list and not a package-list for the javadoc at the provided URL.

Note: To test locally you simply run maven with the default goal: `mvn`. Most of the commons GH actions builds are configured to use the default goal.

arturobernalg:
So do I have to change something? should I close the pr?

garydgregory:
97 files changed... I should be able to take a look over the weekend.

src/main/java/org/apache/commons/lang3/BitField.java
```diff
@@ -56,15 +56,15 @@
  * <p>Flags and data can be retrieved from the integer:</p>
  *
  *<pre>
- *    // Prints true if red, green or blue is non-zero
+ *    // Prints {@code true} if red, green or blue is non-zero
```
garydgregory, line 59: This is a code comment, not a Javadoc comment, so using Javadoc decorations is wrong. Please review all the files and revert this type of change please.

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/BitField.java
arturobernalg: done. Looks like only was this mistake.

garydgregory: @arturobernalg 
You need to update the PR...

arturobernalg: ups. Sorry @garydgregory 
it's done.

src/main/java/org/apache/commons/lang3/BitField.java
```diff
@@ -281,7 +281,7 @@ public byte setByte(final byte holder) {
     }
 
     /**
-     * Sets a boolean BitField.
+     * Sets a {@code boolean} BitField.
```
garydgregory, line 284: Hi @arturobernalg 
In this case, the bit field is a `short`, not a `boolean`, right? It's the bits in the bit field that represent booleans. I think If we are going to touch Javadocs we should make sure they are correct first ;-) See the other setters as well.

PR #965, opened 2022-10-08
Remove unnecessary conditions. It's no need it after assigns tmp to duration

arturobernalg:
We don't need that conditions. Its always false and will never reach it.  And just because in the line 192 we assigns tmp to duration.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -190,9 +190,6 @@ public class DurationFormatUtils {
                     duration = tmp;
                     tmp = StringUtils.replaceOnce(duration, " 0 minutes", StringUtils.EMPTY);
                     duration = tmp;
-                    if (tmp.length() != duration.length()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -190,9 +190,6 @@ public static String formatDurationWords(
                     duration = tmp;
                     tmp = StringUtils.replaceOnce(duration, " 0 minutes", StringUtils.EMPTY);
                     duration = tmp;
-                    if (tmp.length() != duration.length()) {
-                        duration = StringUtils.replaceOnce(tmp, " 0 seconds", StringUtils.EMPTY);
-                    }
```
kinow, line 195: Or is it a bug? Maybe a copypasta bug where accidentally lines 190 and 192 were duplicated? Because there are `replaceOne` for "hour" and "minute", and this block that was removed has the part for "seconds".

arturobernalg: Hi @kinow 
Yes, could be a bug as well.  In any case we don't need that conditions.
TY

PR #967, opened 2022-10-09
add StringUtils.ubstringBeforeLast(String, int)

mmariotti:
Hello,
This method with int param is missing :)


arturobernalg:
HI @mmariotti 
Thank you for your PR. Can you add test cases that cover all the possible path?

TY

Serwios:
@mmariotti Don't you mind if I take it and finish with tests?

garydgregory:
I don't think this is needed or that we should cover every combination of before/after/char/int/any/not and so on.

garydgregory:
Closing: Won't do.

PR #968, opened 2022-10-09
 Add gets the current day of year.

arturobernalg:
Add method to gets the current day of year given a calendar.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
@@ -78,6 +78,16 @@ public class CalendarUtils {
     }
 
     /**
+     * Gets the current day of year.
+     *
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/time/CalendarUtils.java
```diff
@@ -77,6 +77,16 @@ public int getDayOfMonth() {
         return calendar.get(Calendar.DAY_OF_MONTH);
     }
 
+    /**
+     * Gets the current day of year.
+     *
+     * @return the current day of year.
+     * @since 3.12.1
```
garydgregory, line 84: Wrong version.

arturobernalg: done. I ser 3.13.0

PR #969, opened 2022-10-09
Add and use PAD character constant for ' '.

arturobernalg:


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -147,6 +147,13 @@ public class StringUtils {
      */
     public static final String SPACE = " ";
 
+    /**
+     * A space character for pad {@code ' '}.
+     *
+     * @since 3.13.0
+     */
+    public static final char PAD = ' ';
```
garydgregory, line 155: Does not need to be public.

arturobernalg: done

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -147,6 +147,12 @@ public class StringUtils {
      */
     public static final String SPACE = " ";
 
+    /**
+     * A space character for pad {@code ' '}.
+     *
```
garydgregory, line 152: No need for a blank Javadoc line.

arturobernalg: done

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -3808,7 +3813,7 @@ public static boolean isNumericSpace(final CharSequence cs) {
         final int sz = cs.length();
         for (int i = 0; i < sz; i++) {
             final char nowChar = cs.charAt(i);
-            if (nowChar != ' ' && !Character.isDigit(nowChar)) {
+            if (nowChar != PAD && !Character.isDigit(nowChar)) {
```
garydgregory, line 3816: The concept here is not "padding", it's "spacing", so it does not feel appropriate to use the name "PAD".  Note that we have a SPACE constant already but it's a String, not a char, so maybe we need a private space char constant. I'm not sure.

arturobernalg: HI @garydgregory .You're right., I have the same doubt. For the moment I removed that change

PR #976, opened 2022-10-20
Use java stream in ArrayUtils.

arturobernalg:


aherbert:
This seems like de-optimisation to me. The object creation overhead of the stream verses the simple for-loop constructs. I do not think it aids readability or maintainability.

arturobernalg:
> This seems like de-optimisation to me. The object creation overhead of the stream verses the simple for-loop constructs. I do not think it aids readability or maintainability.

It seems that is what the API is tending towards.

aherbert:
Using `Arrays.setAll` is not using a stream. It is offloading the for-loop into `Arrays` to work with a provided lambda function. This is not the same type of change.

arturobernalg:
> Using `Arrays.setAll` is not using a stream. It is offloading the for-loop into `Arrays` to work with a provided lambda function. This is not the same type of change.

Should/must/have close the pr?

PR #983, opened 2022-10-22
Simplify streams.

arturobernalg:


src/test/java/org/apache/commons/lang3/stream/StreamsTest.java
```diff
@@ -136,25 +136,25 @@ public Stream<DynamicTest> simpleStreamForEachFailing() {
 
     @Test
     public void testInstanceOfStream() {
-        assertEquals(2, Streams.instancesOf(String.class, Arrays.asList("A", "B")).collect(Collectors.toList()).size());
-        assertEquals(2, Streams.instancesOf(String.class, Arrays.asList(null, "A", null, "B", null)).collect(Collectors.toList()).size());
-        assertEquals(0, Streams.instancesOf(String.class, Arrays.asList(null, null)).collect(Collectors.toList()).size());
+        assertEquals(2, (int) Streams.instancesOf(String.class, Arrays.asList("A", "B")).count());
```
garydgregory, line 139: Why clutter up the code with all these type-casts? Just let the compiler pick up the `long` version of the JUnit API IMO.

garydgregory (changes requested):
Hi @arturobernalg:  See my inline comment.

PR #984, opened 2022-10-22
Use IntStream

arturobernalg:


src/main/java/org/apache/commons/lang3/text/StrBuilder.java
```diff
@@ -1844,12 +1845,7 @@ public StrBuilder deleteAll(final char ch) {
      * @return this, to enable chaining
      */
     public StrBuilder deleteFirst(final char ch) {
-        for (int i = 0; i < size; i++) {
-            if (buffer[i] == ch) {
-                deleteImpl(i, i + 1, 1);
-                break;
-            }
-        }
+        IntStream.range(0, size).filter(i -> buffer[i] == ch).findFirst().ifPresent(i -> deleteImpl(i, i + 1, 1));
```
DamnedElric, line 1848: The previous implementation was a lot easier to read and created less garbage. This one introduces a bunch of unnecessary objects at seemingly no benefit?

PR #992, opened 2022-11-29
NumberUtils check if Number is null or zero

PiMaDaum:


PiMaDaum:
Sorry, I have forget to comment the new method and clean unused imports.

PR #1005, opened 2023-01-02
formatted java.lang package.

esivakumar18:
formatted java.lang package.


kinow (commented):
Thanks for the PR. I think this file is generated automatically by the ... build or the release plug-in. This text comes from `pom.xml` and will be replaced when the file is re-generated (not sure if the syntax could be updated in `pom.xml` either).

garydgregory:
Correct, this is a generated file, so -1.

PR #1007, opened 2023-01-02
Added new examples for rightPad method

esivakumar18:
Added new examples for `rightPad` method.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -7097,6 +7097,8 @@ public class StringUtils {
      * <pre>
      * StringUtils.rightPad(null, *)   = null
      * StringUtils.rightPad("", 3)     = "   "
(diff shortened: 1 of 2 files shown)


garydgregory:
This does not help, it just makes things bigger without helping users IMO. The additional tests don't even increase code coverage, so there is really little value in even the new assertions IMO.

esivakumar18:
> This does not help, it just makes things bigger without helping users IMO. The additional tests don't even increase code coverage, so there is really little value in even the new assertions IMO.

Hi @garydgregory  thank you for the comment. I have added this example to clear some confusion i.e., if we try to do padding on space, it will yield the same result without any padding. Kindly review again.

esivakumar18:
@kinow could you please review this ?

ecki:
I am also not sure if we need the samples, but i could see some use in explaining that blank is not special, in that case (“    “,3) (similar to („abc“,2) but with blanks is missing.

the sample and tests differ, i would use ,3 in both.

PR #1008, opened 2023-01-02
Make java.desktop module optional

Sineaggi:
As an alternative to https://github.com/apache/commons-lang/pull/275, this allows existing code to continue to work, while providing a migration path for other users. We can choose to either deprecate or keep the methods that take `PropertyChangeListener`s.

By guarding the construction of the `PropertyChangeSupport` with a try/catch, we can continue to run when the `java.desktop` module is absent.

Relates to [LANG-1339](https://issues.apache.org/jira/browse/LANG-1339)


garydgregory:
Broken build, no updates, closing.

PR #1015, opened 2023-01-09
Explicitly checks string for emptiness to prevent the exception logic overhead

fenik17:
Adding the fail fast checks for empty string into `NumberUtils` methods (like `toInt`, `toLong`, etc.) to prevent a big overhead due to generating exception. 

Checking for an empty string is cheap, and a case with an empty string is quite likely.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -100,7 +100,7 @@ public class NumberUtils {
      * Convert a {@link String} to an {@code int}, returning
      * {@code zero} if the conversion fails.
      *
-     * <p>If the string is {@code null}, {@code zero} is returned.</p>
+     * <p>If the string is empty or {@code null}, {@code zero} is returned.</p>
      *
      * <pre>
      *   NumberUtils.toInt(null) = 0
@@ -121,7 +121,7 @@ public class NumberUtils {
      * Convert a {@link String} to an {@code int}, returning a
      * default value if the conversion fails.
      *
-     * <p>If the string is {@code null}, the default value is returned.</p>
+     * <p>If the string is empty or {@code null}, the default value is returned.</p>
      *
      * <pre>
      *   NumberUtils.toInt(null, 1) = 1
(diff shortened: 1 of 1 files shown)


garydgregory:
How much is the overhead and how did you measure it?

fenik17:
@garydgregory it is minimum 1000 times difference. [Benchmark here](https://github.com/fenik17/commons-lang/blob/to_int_benchmark/src/test/java/org/apache/commons/lang3/ToIntBenchmark.java).

Results on my laptop:
```
Benchmark                            Mode  Cnt     Score    Error  Units
ToIntBenchmark.toIntNew_emptyString  avgt    5     0,382 ±  0,016  ns/op
ToIntBenchmark.toIntNew_notNumber    avgt    5  1449,161 ± 17,574  ns/op
ToIntBenchmark.toIntNew_number       avgt    5     1,682 ±  0,021  ns/op
ToIntBenchmark.toIntOld_emptyString  avgt    5  1456,267 ± 19,870  ns/op
ToIntBenchmark.toIntOld_notNumber    avgt    5  1443,685 ± 15,124  ns/op
ToIntBenchmark.toIntOld_number       avgt    5     1,683 ±  0,012  ns/op
```

Exceptions are [very expensive](https://www.baeldung.com/java-exceptions-performance).

garydgregory:
So... there is one case out of three where the PR helps and the others two make things possibly worse due to the error margins?

fenik17:
@garydgregory if an empty string is coming (it is often case in my practice), this change can drastically improve performance. In other cases degradation is not noticeable.

garydgregory:
That's not what your benchmark shows...

fenik17:
@garydgregory could you explain, please? I see the difference is within the margin of error:

```
Benchmark                            Mode  Cnt     Score    Error  Units
ToIntBenchmark.toIntNew_number       avgt    5     1,682 ±  0,021  ns/op
ToIntBenchmark.toIntOld_number       avgt    5     1,683 ±  0,012  ns/op
```

I understand that additional check can worth some CPU time. But in this case, it's literally fractions of nanoseconds. And this is incommensurable with the possible profit.

PR #1026, opened 2023-02-03
Use (<p>) to split the text to different paragraphs in Javadoc comments.

arturobernalg:
See https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -984,7 +984,7 @@ public class ObjectUtils {
 
     /**
      * Checks if an Object is empty or null.
-     *
+     * <p>
      * The following types are supported:
      * <ul>
      * <li>{@link CharSequence}: Considered empty if its length is zero.</li>
@@ -993,7 +993,7 @@ public class ObjectUtils {
      * <li>{@link Map}: Considered empty if it has zero key-value mappings.</li>
      * <li>{@link Optional}: Considered empty if {@link Optional#isPresent} returns false, regardless of the "emptiness" of the contents.</li>
      * </ul>
-     *
+     * </p>
      * <pre>
      * ObjectUtils.isEmpty(null)             = true
      * ObjectUtils.isEmpty("")               = true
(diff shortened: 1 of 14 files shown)


garydgregory:
@arturobernalg Close HTML tags please.

arturobernalg:
> @arturobernalg Close HTML tags please.

why? 
its just a paragraph separator.

garydgregory:
It is cleaner to follow the same element tag rules everywhere and close all tags.

arturobernalg:
> It is cleaner to follow the same element tag rules everywhere and close all tags.

ok, so I do something like ```<p></p>```??

garydgregory:
> > It is cleaner to follow the same element tag rules everywhere and close all tags.
> 
> ok, so I do something like `<p></p>`??

Yes, just like any other tag like em, code, pre, and so on.

arturobernalg:
> > > It is cleaner to follow the same element tag rules everywhere and close all tags.
> > 
> > 
> > ok, so I do something like `<p></p>`??
> 
> Yes, just like any other tag like em, code, pre, and so on.

done

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -984,7 +984,7 @@ public static boolean isArray(final Object object) {
 
     /**
      * Checks if an Object is empty or null.
-     *
+     * <p></p>
```
garydgregory, line 987: Uh?

garydgregory: `<p>`This is a paragraph.`</p>`

arturobernalg: done

garydgregory: Did you not see my example?

garydgregory: Are you not trying a local build before you push to git?

garydgregory: @arturobernalg 
`<p>`This is a paragraph.`</p>`, you can't embed lists per the errors in the build. Again, run a 'mvn' build before you push and you will discover this type of issue without forcing reviewers to ping you for broken builds. TY.

garydgregory (changes requested):
@arturobernalg Please check GHA builds after you push, notice the failures.

PR #1029, opened 2023-02-14
Added empty string example for appendIfMissing method

esivakumar18:
Added empty string example for `appendIfMissing` method


garydgregory:
-1 no need to clutter up the docs IMO with an example that is the same as the one 2 lines below `StringUtils.appendIfMissing("abc", "xyz") = "abcxyz"`

garydgregory:
No feedback, closing.

PR #1033, opened 2023-02-23
[LANG-1694] MethodUtils.getMatchingMethod() fails with "Found multiple candidates"

SeasonPanPan:
JIRA: [https://issues.apache.org/jira/browse/LANG-1694](https://issues.apache.org/jira/browse/LANG-1694)

MethodUtils.getMatchingMethod() fails with "Found multiple candidates for method" message if the method is **an override of an abstract method** and parameter types do not exactly match the declared types.

The behavior of this method has changed between versions 3.11 and 3.12.0. Previously, the method would pick the first method that matches the signature and replace it only if a method with a more specific signature is found. Maybe it's a compatibility issue.

Fix: Keep only one candidate to return directly, if there are more than two candidates, continue to compare whether the declaration classes of the first two methods are the same, if not, also consider the first method the best match(Class itself).

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
@@ -751,8 +751,11 @@ public class MethodUtils {
         }
 
         final List<Method> bestCandidates = candidates.values().iterator().next();
-        if (bestCandidates.size() == 1) {
-            return bestCandidates.get(0);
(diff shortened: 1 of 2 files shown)


SeasonPanPan:
@garydgregory  @esivakumar18 @kinow Do you have time to review my code?

garydgregory:
I'll take a look this weekend.

SeasonPanPan:
> I'll take a look this weekend.

Thank u.

PR #1036, opened 2023-03-02
Typo in EnumUtils class

jblaski1:
Fixed a small typo in the EnumUtils documentation, where the wording does not match that of other methods in the class.


kinow (approved):
Thanks!

PR #1038, opened 2023-03-14
[LANG-1687] Propose update from substringBetween doc and new method

laurentschoelens:
Propose an update to documentation of substringBetween (treat LANG-1687) and add new method to have behaviour wanted by the user

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8778,17 +8778,59 @@ public class StringUtils {
      *
      * @param str  the String containing the substring, may be null
      * @param open  the String before the substring, may be null
-     * @param close  the String after the substring, may be null
+     * @param close  the String after the substring, may be null - search first match from left to right, after the <code>open</code> String found
      * @return the substring, {@code null} if no match
      * @since 2.0
      */
     public static String substringBetween(final String str, final String open, final String close) {
+        return substringBetween(str, open, close, false);
+    }
+    /**
+     * Gets the String that is nested in between two Strings.
+     * Only the first match is returned.
+     *
+     * <p>A {@code null} input String returns {@code null}.
+     * A {@code null} open/close returns {@code null} (no match).
+     * An empty ("") open and close returns an empty string.</p>
+     *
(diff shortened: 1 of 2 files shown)


garydgregory:
I am -1 for this new feature which feels like a super niche use case which overcomplicates the code base. I'm curious to hear what others think.

laurentschoelens:
HI @garydgregory thanks for the review and comment.
Someone thought the stringBetween works like taking the substring between two string, with last argument taken as far as possible, and maybe there is some use-cases, and the documentation is not clear on that.

If not adding the new function, maybe updating the documentation to make it clear what it does should be enough :)

laurentschoelens:
seen build failed due to checkstyle violation (not having checkstyle activated on my IDE), added a fix for this

laurentschoelens:
@garydgregory : got an exemple for you of usage
`5 * ((6 * 7 + 2) * 3)`
Actual method will return `(6 * 7 + 2` where we would lire to have `(6 * 7 + 2) * 3` with params `"("), ")"`

garydgregory:
> @garydgregory : got an exemple for you of usage `5 * ((6 * 7 + 2) * 3)` Actual method will return `(6 * 7 + 2` where we would like to have `(6 * 7 + 2) * 3` with params `"(", ")"`

Hello @laurentschoelens 
I hope that's a random made up example because it's not a good one ;-) Expression parsing should not be hand-coded IMO, that's what Antlr and JavaCC are for...

laurentschoelens:
> > @garydgregory : got an exemple for you of usage `5 * ((6 * 7 + 2) * 3)` Actual method will return `(6 * 7 + 2` where we would like to have `(6 * 7 + 2) * 3` with params `"(", ")"`
> 
> Hello @laurentschoelens I hope that's a random made up example because it's not a good one ;-) Expression parsing should not be hand-coded IMO, that's what Antlr and JavaCC are for...

Yes that is hand made but it's only to point that there could be some usage 😀

trumpstermother (commented):
src/main/java/org/apache/commons/lang3/StringUtils.java

garydgregory:
Closing, discussion stopped months ago, won't do.

PR #1039, opened 2023-03-14
LANG-1682 : adding new startsWithAnyIgnoreCase method and tests cases

laurentschoelens:
Adding new method `startsWithAnyIgnoreCase` method for case-insensitive version of `startsWithAny`


laurentschoelens:
Closing as non-wanted commits in PR

PR #1040, opened 2023-03-14
[LANG-1682] Adding new startsWithAnyIgnoreCase method and tests cases

laurentschoelens:
Adding new method startsWithAnyIgnoreCase method for case-insensitive version of startsWithAny

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8057,6 +8057,39 @@ public class StringUtils {
     }
 
     /**
+     * Case insensitive check if a CharSequence starts with any of the provided prefixes.
+     *
+     * <pre>
+     * StringUtils.startsWithAny(null, null)      = false
+     * StringUtils.startsWithAny(null, new String[] {"abc"})  = false
+     * StringUtils.startsWithAny("abcxyz", null)     = false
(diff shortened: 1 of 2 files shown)


garydgregory:
My preference is to stop adding APIs that are variations of other APIs based on the casing (and encoding and so on). In the longer term, I can see adding a new String utility class that tracks such an attribute on its instance such that you'd say something like `Strings.caseSensitive().someOperation(...)` and `Strings.caseInsensitive().someOperation(...)`.

laurentschoelens:
Thanks for the reply.
Is there any plans for this ? It's huge refactoring.

garydgregory:
It's something I'd like to do and might take a swing at when I am on vacation in April...

laurentschoelens:
Like you expose the calling to this new Strings class, it looks like you're creating new instance (like a builder) and that should lead to memory impact and performances issues (as I see it)

laurentschoelens:
> Like you expose the calling to this new Strings class, it looks like you're creating new instance (like a builder) and that should lead to memory impact and performances issues (as I see it)

If not, this could lead to thread-safety issues. But again I don't know what you want to do

garydgregory:
@laurentschoelens 
Closing: Please try your use case with the Strings class in git master and snapshot builds.

laurentschoelens:
I've seen your commit regarding the Strings class. I'm pretty sure that will do the job.
Thanks

PR #1041, opened 2023-03-14
[LANG-1643] Construct ArrayList with better default size

laurentschoelens:
As pointed by LANG-1643, the `ArrayList<>()` constructor creates an empty List which will always grow at first add, which is done in the next line of code in `EqualsBuilder`


garydgregory:
https://issues.apache.org/jira/browse/LANG-1643

PR #1048, opened 2023-04-05
Refactored to make use of EMPTY constant

esivakumar18:


garydgregory (changes requested):
We only use static imports for unit test assertion methods.

garydgregory (changes requested):
We only use static imports for assertion methods.

garydgregory:
Closing: No reply in months.

PR #1052, opened 2023-04-14
refactor: Only use static imports to import assert methods in tests

remeio:
Only use `static` imports to import `assert methods` in tests.


kinow (commented):
I approved the GH actions to run, but I don't have a preference on this. I just follow whatever the project is using. So either way is fine for me.

remeio:
: ), I just keep the same code style with this commit https://github.com/apache/commons-lang/commit/c7d7a207671d407f18e55bfcefb448cda0240fcd

PR #1053, opened 2023-04-14
Refactor BooleanUtils#toBooleanObject for readability

remeio:
- Remove the unnecessary unboxing(**removed**)
- Use special boolean values array instead of hard code:
```java
      // The left array (index%2==0) is the Boolean.TRUE strings, the right (index%2==1) is the Boolean.FALSE strings.
      // The strings have the same length (index%2==0?index:index+1) for each pair.
      final String[][] booleanValues = new String[][]{
              {}, {},
              {"y", "t", "1"}, {"n", "f", "0"},
              {ON}, {NO},
              {YES}, {OFF},
              {TRUE}, {},
              {}, {FALSE},
      };
``` 
It has two advantages:

1. For `OCP`, if we add new boolean values, we just need adjust the array
2. Keep the same performance with old code

It is good for short values such as `T`, `OK`, `NOT`, but it is not good for too long values such as `truthful`.
```java
{
  {}, {},
  {}, {},
  {}, {},
  {}, {},
  {}, {},
  {}, {},
  {}, {},  // empty array for seize a seat, :(
  {}, {},
  { "truthful" }, {},
 
}
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BooleanUtils.java
+++ b/src/main/java/org/apache/commons/lang3/BooleanUtils.java
@@ -743,82 +743,28 @@ public class BooleanUtils {
         if (str == null) {
             return null;
         }
-        switch (str.length()) {
-            case 1: {
-                final char ch0 = str.charAt(0);
-                if (ch0 == 'y' || ch0 == 'Y' ||
-                    ch0 == 't' || ch0 == 'T' ||
-                    ch0 == '1') {
-                    return Boolean.TRUE;
-                }
-                if (ch0 == 'n' || ch0 == 'N' ||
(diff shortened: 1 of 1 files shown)


garydgregory:
In general, don't change (un)boxing, this is a low level library and we want to see where (un)boxing happens to watch out for NPEs and performance gotchas.

remeio:
Thanks. And do you thinks the refactor of `BooleanUtils#toBooleanObject` is necessary?
I think the old code maybe likes hard code.

garydgregory:
-1, does not need fixing. The original author obviously coded this method for performance, not readability.

PR #1057, opened 2023-04-17
Feature add NOP for TriFunction

remeio:
NOP for TriFunction is necessary?

### DIFF
--- a/src/main/java/org/apache/commons/lang3/function/TriFunction.java
+++ b/src/main/java/org/apache/commons/lang3/function/TriFunction.java
@@ -39,6 +39,19 @@ import java.util.function.Function;
 @FunctionalInterface
 public interface TriFunction<T, U, V, R> {
 
+    /** NOP singleton */
+    @SuppressWarnings("rawtypes")
+    TriFunction NOP = (t, u, v) -> null;
+
+    /**
+     * Returns The NOP singleton.
+     *
+     * @return The NOP singleton.
+     */
+    static <T, U, V, R> TriFunction<T, U, V, R> nop() {
+        return NOP;
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


aherbert:
If `TriFunction` is the three-arity specialization of `Function` then this interface should support the methods in the JDK and not add more. `java.util.function.BiFunction` does not have a `nop()` method in the latest LTS (JDK 17).

A no-op makes sense for a consumer with a void return. When returning a result, even specifying `null` is an operation.

remeio:
I agree the  ` When returning a result, even specifying null is an operation.` But in package `org.apache.commons.lang3.function`, there seems to be some kind of agreement.

- NOP for Function: (t, u) -> null;
- NOP for Consumer: (t, u) -> {/* NOP */};

Such as For Function:
```java
public interface FailableBiFunction<T, U, R, E extends Throwable> {

    /** NOP singleton */
    @SuppressWarnings("rawtypes")
    FailableBiFunction NOP = (t, u) -> null;
}
``` 
For Consumer: 
```java
@FunctionalInterface
public interface FailableBiConsumer<T, U, E extends Throwable> {

    /** NOP singleton */
    @SuppressWarnings("rawtypes")
    FailableBiConsumer NOP = (t, u) -> {/* NOP */};
}
```

garydgregory:
This PR breaks the build.

remeio:
I had repaired it.

aherbert:
The `noop` was added for [LANG-1568](https://issues.apache.org/jira/browse/LANG-1568) which was to add failable equivalents of JDK functional interfaces. There is not a `noop` in the JDK. Why the method was added is not clear and it has no use case in the library.

remeio:
I get it. : )

PR #1062, opened 2023-06-08
Syntax for optional tokens in DurationFormatUtils

theshoeshiner:
Relatively simple low-impact implementation of optional tokens within DurationFormatUtils. Useful when you need more human readable duration strings.

Brackets [] are used to denote optional tokens. Optional tokens are only printed if non-zero. Optional literals are only printed if the preceding non-literal was non-zero.

e.g. A single format:  `[d'd 'H'h 'm'm ']s's'`
Can produce outputs like:
```
1d 2h 3m 14s
2h 3m 14s
3m 14s
14s
0s
```

2 unit tests added.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -64,7 +64,7 @@ import org.apache.commons.lang3.Validate;
  * <br>
  * <table border="1">
  * <caption>Example Output</caption>
- * <tr><th>pattern</th><th>Duration.ofDays(1)</th><th>Duration.ofHours(1)</th><th>Duration.ofMinutes(1)</th><th>Duration.ofMillis(0)</th></tr>
+ * <tr><th>pattern</th><th>Duration.ofDays(1)</th><th>Duration.ofHours(1)</th><th>Duration.ofMinutes(1)</th><th>Duration.ZERO</th></tr>
  * <tr><td>d'd'H'h'm'm's's'</td><td>1d0h0m0s</td><td>0d1h0m0s</td><td>0d0h1m0s</td><td>0d0h0m0s</td></tr>
  * <tr><td>d'd'[H'h'm'm']s's'</td><td>1d0s</td><td>0d1h0s</td><td>0d1m0s</td><td>0d0s</td></tr>
  * <tr><td>[d'd'H'h'm'm']s's'</td><td>1d0s</td><td>1h0s</td><td>1m0s</td><td>0s</td></tr>
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -526,9 +526,11 @@ public class DurationFormatUtils {
             switch (ch) {
             // TODO: Need to handle escaping of '
             case '[':
+                if(inOptional) throw new IllegalArgumentException("Nested optional block at index: "+i);
                 inOptional = true;
                 break;
             case ']':
+                if(!inOptional) throw new IllegalArgumentException("Attempting to close unopened optional block at index: "+i);
                 inOptional = false;
                 break;
             case '\'':
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -686,7 +686,7 @@ public class DurationFormatUtils {
          * @param optionalIndex the index of the optional token within the pattern
          */
         Token(final Object value, final boolean optional, final int optionalIndex) {
-            this.value = value;
+            this.value = Objects.requireNonNull(value, "value");
             this.count = 1;
             if (optional) {
                 this.optionalIndex = optionalIndex;
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -429,48 +429,48 @@ public class DurationFormatUtils {
             final int count = token.getCount();
             optional = token.optional;
             if (value instanceof StringBuilder) {
-            	if(!optional || !lastOutputZero) buffer.append(value.toString());
+                if(!optional || !lastOutputZero) buffer.append(value.toString());
             } else if (value.equals(y)) {
                 lastOutputSeconds = false;
                 lastOutputZero = years == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(years, padWithZeros, count));
+                    buffer.append(paddedValue(years, padWithZeros, count));
             } else if (value.equals(M)) {
                 lastOutputSeconds = false;
                 lastOutputZero = months == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(months, padWithZeros, count));
+                    buffer.append(paddedValue(months, padWithZeros, count));
             } else if (value.equals(d)) {
                 lastOutputSeconds = false;
                 lastOutputZero = days == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(days, padWithZeros, count));
+                    buffer.append(paddedValue(days, padWithZeros, count));
             } else if (value.equals(H)) {
                 lastOutputSeconds = false;
                 lastOutputZero = hours == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(hours, padWithZeros, count));
+                    buffer.append(paddedValue(hours, padWithZeros, count));
             } else if (value.equals(m)) {
                 lastOutputSeconds = false;
                 lastOutputZero = minutes == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(minutes, padWithZeros, count));
+                    buffer.append(paddedValue(minutes, padWithZeros, count));
             } else if (value.equals(s)) {
                 lastOutputSeconds = true;
                 lastOutputZero = seconds == 0;
                 if(!optional || !lastOutputZero) 
-                	buffer.append(paddedValue(seconds, padWithZeros, count));
+                    buffer.append(paddedValue(seconds, padWithZeros, count));
             } else if (value.equals(S)) {
-            	lastOutputZero = milliseconds == 0;
-            	if(!optional || !lastOutputZero) {
+                lastOutputZero = milliseconds == 0;
+                if(!optional || !lastOutputZero) {
                     if (lastOutputSeconds) {
 	                        // ensure at least 3 digits are displayed even if padding is not selected
-	                    	final int width = padWithZeros ? Math.max(3, count) : 3;
+	                        final int width = padWithZeros ? Math.max(3, count) : 3;
 	                        buffer.append(paddedValue(milliseconds, true, width));
                 } else {
-                    	buffer.append(paddedValue(milliseconds, padWithZeros, count));
+                        buffer.append(paddedValue(milliseconds, padWithZeros, count));
                     }
-            	}
+                }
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -632,4 +632,31 @@ public void testLowDurations() {
         }
     }
 
+    @Test
+    public void testUnmatchedOptionalToken() {
+        assertThrows(IllegalArgumentException.class, () -> DurationFormatUtils.formatDuration(1, "[s"));
```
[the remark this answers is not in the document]
theshoeshiner: Nesting has little value that I can see, so simplest solution will be to catch any nesting and throw an IllegalArgumentException. That will prevent unbalancing issues since the current optional state is tracked with a boolean, similar to literals, and we can easily detect attempts to nest.

Multiple optional blocks are supported by default (though realistic use cases may be few). If you wanted to always show minutes, but omit zero-value hours and seconds, then the format `[H'h ']m'm'[s's ']` will produce values like:
```
254h16m
15m15s 
0m9s
```
Will add a commit to address the nesting issue.

garydgregory:
I'm looking for community feedback. CC: @kinow @aherbert @chtompki ?

theshoeshiner:
I will add tests for `formatPeriod` as well, and check on my end for other library methods invoking these.

In regards to the javadocs, the method level `formatDuration` and `formatPeriod` docs not go into any other details about the syntax of the format string, as those details are covered in the class level doc. To follow that precedent I can put a paragraph at the class level that explains the optional syntax (unless anyone feels like this change deserves mentioning in both areas).

garydgregory:
Committers:
Run `mvn` (just that) locally before you push to avoid these types of build failures.

aherbert (changes requested):
I think this is a quirk: processing of literal values within an optional block will print the first literal:
```java
String format = "['{'dd'}']HH:mm:ss";
LongStream.of(0,
              Duration.ofSeconds(1).toMillis(),
              Duration.ofMinutes(2).toMillis(),
              Duration.ofHours(3).toMillis(),
              Duration.ofDays(4).toMillis())
          .mapToObj(millis -> DurationFormatUtils.formatDuration(millis, format))
          .forEach(System.out::println);
```
prints:
```
{00:00:00
{00:00:01
{00:02:00
{03:00:00
{04}00:00:00
```
The question is how to process the first literal in an optional block if it comes before a format token. Should the literal attach to the previous token that was before the optional block, or should it attach to the first token in the same optional block. IMO since it is inside the optional block then its presence should be mandated by the first format token inside the same optional block. So the above example should be:
```
00:00:00
00:00:01
00:02:00
03:00:00
{04}00:00:00
```
Looking at the code this change may require labelling of the optional regions with numbers and not a boolean flag to allow processing is groups. Or addition of another marker for the token to indicate {not-optional; optional; first in optional}.

Note: I tested using '[' as a literal inside an optional. This works. I think it should be added as a unit test to prevent future regressions, e.g.
```java
assertEquals("[3m]1s", DurationFormatUtils.formatDuration((3 * 60 + 1) * 1000, "['['m'm]']s's'"));
```
This cannot test the optional without the minutes present as it will print `"[1s"` due to the issue with leading optional literals.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -48,6 +48,13 @@
  * Token values are printed using decimal digits.
  * A token character can be repeated to ensure that the field occupies a certain minimum
  * size. Values will be left-padded with 0 unless padding is disabled in the method invocation.
+ * <br>
+ * Tokens can be marked as optional by surrounding them with brackets [ ]. These tokens will 
+ * only be printed if the token value is non-zero. Any literals within optional blocks will only be 
+ * printed if the nearest prior non-literal token value was non-zero.
```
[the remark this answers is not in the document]
theshoeshiner: Originally I could not come up with too many scenarios where a leading literal would make sense, but surrounding a token with brackets does seem like a reasonable use. I agree the simplest/best solution would be to use the first token.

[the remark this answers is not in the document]
theshoeshiner: Two commits to address this. Tried to keep it simple, but still required minor change to the Token class and multiple tracking variables in the format method.

Now it tracks leading literals in optional blocks, and when encountering a non-literal token we remove the prior literal from the buffer if the token is zero. Optional blocks are differentiated by tracking an optional "index" in the token class - as opposed to the prior boolean field. Literals following tokens are handled same as before.

Also added an example table to highlight the likely usage and potential pitfalls.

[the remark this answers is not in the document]
theshoeshiner: Also - not sure how you guys prefer to handle constructor collisions. But this change required a Token constructor method which collides with the Token(int,int) constructor that is used by unit tests. The addition of a boolean parameter to the new constructor resolves this. The parameter purpose is still clear and functional, but code could be refactored to remove it (were it not for the collision).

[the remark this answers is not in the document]
garydgregory: I'll repeat:

> Run mvn (just that) locally before you push to avoid these types of build failures.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -50,12 +50,27 @@
...
+ * only be printed if the token value is non-zero. Literals within optional blocks will only be
+ * printed if the preceding non-literal token is non-zero. Leading optional literals will only
+ * be printed if the following non-literal is non-zero.
+ * Multiple optional blocks can be used to group literals with the desired token. Note that
+ * multiple optional tokens without literals can result in impossible to understand output.
+ * (See examples below)
  * <br>
+ * <table border="1">
+ * <caption>Example Output</caption>
+ * <tr><th>pattern</th><th>Duration.ofDays(1)</th><th>Duration.ofHours(1)</th><th>Duration.ofMinutes(1)</th></tr>
```
aherbert, line 63: This is nice. For the entry with all optional formatting `[d'd'H'h'm'm's's']` it would be useful to state what happens when the duration is 0. E.g. If the entire format is optional and the duration is zero the result is the empty string.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -621,22 +672,23 @@ static boolean containsTokenWithValue(final Token[] tokens, final Object value)
...
         }
-        
+
         /**
          * Wraps a token around a value. A value would be something like a 'Y'.
          *
          * @param value to wrap, non-null.
          */
-        Token(final Object value, final boolean optional) {
+        Token(final Object value, final boolean optional, final int optionalIndex) {
```
aherbert, line 686: You should javadoc the other parameters and why they are used.

Note: It seems odd to have the other two constructors. These are only used in the tests. If you concur I think it sensible to remove them and add a helper method in the test itself to create these objects using the constructor that your code uses. IIUC the helper method can pass in the desired Object value then call increment a number of times as appropriate.

aherbert (changes requested):
This looks good. Would you like to remove the other constructors for Token and change to use a helper method to construct the objects for the tests? IMO this would make it clearer when reading the code that the only way a Token is used during lexx is through the constructor added in this PR. Elsewhere there are just a few simple javadoc extras to add.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
theshoeshiner: Agree. Added a column for Duration.ofMillis(0) so that empty string results are noticeable, as well as a specific Note.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
theshoeshiner: I found the presence of the test-only constructors odd as well. I went ahead and removed them and replaced with a helper method on the test class.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
garydgregory: > Agree. Added a column for Duration.ofMillis(0) so that empty string results are noticeable, as well as a specific Note.

Is the 0 important or the Millis part? It might be more obvious to refer to Duration.ZERO.

theshoeshiner: You are correct. I hadn't checked for that static field. Will update.

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -632,4 +633,154 @@ public void testLowDurations() {
...
+
+    @Test
+    public void testLiteralPrefixOptionalToken() {
+      assertEquals(
+          DurationFormatUtils.formatDuration(10000L, "s's'"),
+          DurationFormatUtils.formatDuration(10000L, "['['d']']['<'H'>']['{'m'}']s's'"));
+      assertEquals(
+          DurationFormatUtils.formatDuration(10000L, "s's'"),
+          DurationFormatUtils.formatDuration(10000L, "['{'m'}']s's'")
+          );
```
aherbert, line 695: Can be on previous line

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -632,4 +633,154 @@ public void testLowDurations() {
...
+
+    @Test
+    public void testEmptyOptionals() {
+      assertEquals(
+          "",
+          DurationFormatUtils.formatDuration(0L, "[d'd'][H'h'][m'm'][s's']"));
+      assertEquals(
+          "",
+          DurationFormatUtils.formatDuration(0L, "['d''h''m''s's]")
+          );
```
aherbert, line 706: Can be on previous line

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -632,4 +633,154 @@ public void testLowDurations() {
...
+    public void testMultipleOptionalBlocks() {
+
+      assertEquals(
+          DurationFormatUtils.formatDuration(Duration.ofHours(1).toMillis(), "'[['H']]'"),
+          DurationFormatUtils.formatDuration(Duration.ofHours(1).toMillis(), "['{'d'}']['[['H']]']"));
+
+      assertEquals(
+          DurationFormatUtils.formatDuration(Duration.ofDays(1).toMillis(), "['{'d'}']"),
+          DurationFormatUtils.formatDuration(Duration.ofDays(1).toMillis(), "['{'d'}']['['H']']"));
+
```
aherbert, line 719: Remove empty line

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -632,4 +633,154 @@ public void testLowDurations() {
...
+        //make sure we can handle omitting multiple literals after a token
+
+        assertEquals(
+                DurationFormatUtils.formatPeriod(915361L, 9153610L, "H'h''h2'm'm's's'"),
+                DurationFormatUtils.formatPeriod(915361L, 9153610L, "[d'd''d2'H'h''h2'm'm']s's'"));
+    }
+
+    private DurationFormatUtils.Token createTokenWithCount(final Object value, final int count) {
+        DurationFormatUtils.Token token = new DurationFormatUtils.Token(value, false, -1);
+        for (int i=1; i<count; i++) {
```
aherbert, line 781: Whitespace padding: `for (int i = 1; i < count; i++) {`

aherbert (changes requested):
This looks good to me. Just a few formatting niggles.

PR #1068, opened 2023-06-28
calculate the left-difference between 2 arrays

orionlibs:
Created methods that calculate the left-difference between 2 arrays. For example for {1, 2, 3, 4}-{1, 6, 8, 3}={2, 4}.
These methods could be useful in several scenarios:
1--Data Analysis: When dealing with large datasets, you may need to compare two arrays of data to identify the differences between them. This method can help you find the elements that exist in one dataset but not in the other, enabling you to analyze the discrepancies and make informed decisions based on the results.
2--Database Operations: In database operations, you might have two arrays representing different sets of data, such as database records or query results. By using this method, you can identify the elements that are present in one array but missing in the other, helping you detect anomalies, perform data synchronization, or handle data migrations effectively.
3--Set Operations: The method can be used to perform set operations, such as finding the relative complement of two sets. It helps you determine the elements that belong to the first set but not to the second set, providing you with a new array containing the unique elements.
4--Filtering: In some cases, you may have a large dataset and want to filter out specific elements based on a comparison with another dataset. This method can help you identify and extract the elements that are present in the first array but not in the second, allowing you to create a new filtered array.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -20,18 +20,22 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Random;
+import java.util.Set;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.IntFunction;
 import java.util.function.Supplier;
-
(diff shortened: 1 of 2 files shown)


garydgregory:
-1: Set operations belong, IMO in Apache Commons Numbers or Apache Commons Math. Otherwise, this opens the door for all set operations. What do others think?

orionlibs:
@garydgregory you spotted it immediately. That is exactly what I was planning. I have a list of about 60 array-related operations. This one I just did for this PR was just the first. I think I should put these in commons math. I am closing this PR

aherbert:
Note that this is not returning distinct elements from array 1 so is not a strict set operation. E.g. The provided code will compute:
```
{1, 2, 3, 4, 4} - {1, 6, 8, 3} = {2, 4, 4}
```

orionlibs:
@aherbert you are right. It was supposed to be an array operation and not a set operation. But it is OK. I closed this PR. I will implement set theory and vector stuff in commons math/numbers slowly in time. Thank you for the feedback anyway

PR #1069, opened 2023-06-29
LANG-1647

orionlibs:
--created method in ExceptionUtils that checks if a given Throwable represents a checked exception.
--created method in ExceptionUtils that checks if a given Throwable represents an unchecked exception.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/exception/ExceptionUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/exception/ExceptionUtilsTest.java
@@ -852,6 +852,11 @@ public class ExceptionUtilsTest extends AbstractLangTest {
     }
 
     @Test
+    public void testIsChecked_null() {
+        assertFalse(ExceptionUtils.isChecked(null));
+    }
+
+    @Test
     public void testIsChecked_unchecked() {
         assertFalse(ExceptionUtils.isChecked(new IllegalArgumentException()));
     }
--- a/src/test/java/org/apache/commons/lang3/exception/ExceptionUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/exception/ExceptionUtilsTest.java
@@ -850,4 +850,40 @@ public class ExceptionUtilsTest extends AbstractLangTest {
         final Throwable t = assertThrows(Throwable.class, () -> ExceptionUtils.wrapAndThrow(new TestThrowable()));
         assertTrue(ExceptionUtils.hasCause(t, TestThrowable.class));
     }
+
+    @Test
+    public void testIsChecked_unchecked() {
+        final boolean result = ExceptionUtils.isChecked(new IllegalArgumentException());
+        assertFalse(result);
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


garydgregory:
-1 as is, in https://issues.apache.org/jira/browse/LANG-1647 the description reads:
```
The idea it's have a function that check if a given throwable is a checked exception. There are some similar function in ConcurrentUtils, but have package visibility and return the same exception. 

Seem logic have this verification in the class that have this responsibility - ExceptionUtils
```
`ConcurrentUtils` and anywhere else in Commons Lang should reuse these new APIs, otherwise, the PR just creates duplication of existing code patterns.

src/test/java/org/apache/commons/lang3/exception/ExceptionUtilsTest.java
```diff
@@ -850,4 +850,40 @@ public void testWrapAndUnwrapThrowable() {
         final Throwable t = assertThrows(Throwable.class, () -> ExceptionUtils.wrapAndThrow(new TestThrowable()));
         assertTrue(ExceptionUtils.hasCause(t, TestThrowable.class));
     }
+
+    @Test
+    public void testIsChecked_unchecked() {
+        final boolean result = ExceptionUtils.isChecked(new IllegalArgumentException());
```
garydgregory, line 856: @orionlibs 
There is no need for a single-use local variable here and in other new tests.

orionlibs: @garydgregory so, should I create new local variables for each case?

orionlibs: @garydgregory what should I do for this PR to be successful?

garydgregory: > @garydgregory so, should I create new local variables for each case?

You've got it backward.

orionlibs: @garydgregory  so, I don't need to create local variables if the assertion looks like this:
assertTrue(ExceptionUtils.isChecked(new IllegalArgumentException()));
Is that what you mean?

garydgregory: 👍😉

orionlibs: done!

garydgregory: Not quite, see https://github.com/apache/commons-lang/pull/1069#issuecomment-1616686083

orionlibs: @garydgregory you are right. I made ConcurrentUtils to use the new method. I hope this completes this PR. Thank you for pointing that comment out. I focused on other parts and I missed that part of the comment. Distractions are bad

PR #1070, opened 2023-06-29
LANG-1644

orionlibs:
--created method that checkes if a given string represents a hexadecimal number without creating a number

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -1856,9 +1856,9 @@ public class NumberUtils {
      * and {@code 1} if it is.
      * @since 3.13
      */
-    public static boolean isHexadecimal(String str) {
+    public static boolean isHexadecimal(CharSequence str) {
         if (str != null) {
-            return str.matches("^(0x|0X)?[0-9a-fA-F]+$");
+            return str.toString().matches("^(0x|0X)?[0-9a-fA-F]+$");
         }
 
         return false;
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -1847,4 +1847,20 @@ public class NumberUtils {
     public static int compare(final byte x, final byte y) {
         return x - y;
     }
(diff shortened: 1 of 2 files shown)


elharo (commented):
Any limits on size of hexadecimal int? Should there be or is this a big integer?

garydgregory:
I wonder if this kind of text processing belongs here instead of Commons Text, especially since there is no use case for Lang itself.

elharo:
It could perhaps be both, but personally I'd probably look here first. I wouldn't expect a text processing package to have much to do with numbers. Similarly I'd look in a datetime package for processing of string to dates, a JSON package for processing of strings to JSON, and so on. commons-text shouldn't need to know much about non-text data types.

aherbert:
Note that the PR author is a new contributor working on existing tickets in Commons that they did not create.

This PR is related to a ticket created by Arturo Bernal. In most cases of such tickets there is a corresponding PR with a discussion on why the original contribution was rejected. These should be considered when updating a contribution.

Please see my comments on this functionality in PR #727. I think the method has no use cases, i.e. why detect if a number is a hexidecimal if you are not then going to create one using the available parsing methods in NumberUtils. This PR currently only handles unsigned integer numbers. IMO it should be fixed to be compatible with NumberUtils such that it is:

```
"a subset of isCreateable that will return true only if the Number created by createNumber is a mathematical integer"
```

Note also the recently merged #1069 by @orionlibs adds functionality that was never merged from another Arturo ticket #729. The reason is not in the PR (which was recently closed probably due to stagnation) and the corresponding Jira ticket 1647 was deleted. The mailing list had some discussion on the merits of the method and it was inconclusive. I would be cautious about working on long standing tickets without first viewing associated PRs. In most cases the tickets are left open but should have been marked with a reason for the PR rejection if applicable so it can be addressed by further work.

Note that this is not a rejection of the PR or new contributions but a wish to see the full history of the functionality considered so we can merge code with full consideration of the use cases.

elharo:
I do think these methods have a purpose in Java, though they're somewhat pedantic ones. It avoids the need to catch NumberFormatExceptions in normal control flow. I.e. it prevents NumberFormatExceptions from being thrown, which is aguably preferable from throwing and catching.

PR #1071, opened 2023-06-29
LANG-1695: NumberUtils::isParseable does not recognise numbers ending in dot

orionlibs:
[LANG-1695](https://issues.apache.org/jira/browse/LANG-1695) allowed for a number (positive or negative) that has only one decimal point and it is at the end of the input, is treated as a valid number

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -1754,9 +1754,6 @@ public class NumberUtils {
         if (StringUtils.isEmpty(str)) {
             return false;
         }
-        if (str.charAt(str.length() - 1) == '.') {
-            return false;
-        }
         if (str.charAt(0) == '-') {
             if (str.length() == 1) {
                 return false;
@@ -1776,7 +1773,7 @@ public class NumberUtils {
             if (decimalPoints > 1) {
                 return false;
             }
-            if (!isDecimalPoint && !Character.isDigit(str.charAt(i))) {
+            if (!isDecimalPoint && !Character.isDigit(str.charAt(i)) && !(i == str.length() - 1 && str.charAt(i) == '.')) {
                 return false;
             }
         }
--- a/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
@@ -918,7 +918,8 @@ public class NumberUtilsTest extends AbstractLangTest {
         assertFalse(NumberUtils.isParsable("pendro"));
         assertFalse(NumberUtils.isParsable("64, 2"));
         assertFalse(NumberUtils.isParsable("64.2.2"));
-        assertFalse(NumberUtils.isParsable("64."));
+        assertTrue(NumberUtils.isParsable("64."));
+        assertTrue(NumberUtils.isParsable("-64."));
         assertFalse(NumberUtils.isParsable("64L"));
         assertFalse(NumberUtils.isParsable("-"));
         assertFalse(NumberUtils.isParsable("--2"));
--- a/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
@@ -919,6 +919,7 @@ public class NumberUtilsTest extends AbstractLangTest {
         assertFalse(NumberUtils.isParsable("64, 2"));
         assertFalse(NumberUtils.isParsable("64.2.2"));
         assertTrue(NumberUtils.isParsable("64."));
+        assertFalse(NumberUtils.isParsable("64.."));
         assertTrue(NumberUtils.isParsable("-64."));
         assertFalse(NumberUtils.isParsable("64L"));
         assertFalse(NumberUtils.isParsable("-"));


garydgregory:
@orionlibs 
Thank you for the PR. Would you please reply in the Jira ticket to the comment https://issues.apache.org/jira/browse/LANG-1695?focusedCommentId=17709278&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17709278

We really need to make sure we are not creating trouble for ourselves ;-)

orionlibs:
@garydgregory I am new to open source contribution and a week ago I started with Apache. I applied for a JIRA account, but I was denied one (with no reason given). So, I cannot login to JIRA

orionlibs:
@garydgregory also there are many many open tickets in JIRA for many projects that have actually been completed, but they are still open, with no assignees or comments or anything.

garydgregory:
> @garydgregory I am new to open source contribution and a week ago I started with Apache. I applied for a JIRA account, but I was denied one (with no reason given). So, I cannot login to JIRA. Here is the email from apache: We regret to inform you that, upon reviewing your request for a new Jira account connected with The Apache Software Foundation, the commons project has chosen to deny the request. We therefore will not create the Jira account.
> 
> The following reason was given: No reason given.
> 
> If you wish to appeal this decision, you may do so either directly with the commons project, or by contacting the ASF Infrastructure team at: [users@infra.apache.org](mailto:users@infra.apache.org)

What user name and real name did you use in the application? I'll see if I can find the application and rejection in the archives.

orionlibs:
> What user name and real name did you use in the application? I'll see if I can find the application and rejection in the archives.

@garydgregory I used this page
https://selfserve.apache.org/jira-account.html
and the info I entered was:
email address = efthymiou.dimitrios1@gmail.com
username = dimitrios.efthymiou
real public name = Dimitrios Efthymiou
tell us the intent for this account = I don't need the account now, but I don't know when I will. It is like when you join a company. They give you access to some basic services. (This was probably the reason, but I didn't know then that I needed access to JIRA in order to update the tickets I worked on. I worked on a few tickets that were unassigned and did not have PR links in them. With a JIRA account I can assign myself to the tickets I work on and I can add my PR links in there. I guess that was the reason they rejected me)

garydgregory:
> @garydgregory I am new to open source contribution and a week ago I started with Apache. I applied for a JIRA account, but I was denied one (with no reason given). So, I cannot login to JIRA. Here is the email from apache: We regret to inform you that, upon reviewing your request for a new Jira account connected with The Apache Software Foundation, the commons project has chosen to deny the request. We therefore will not create the Jira account.
> 
> The following reason was given: No reason given.
> 
> If you wish to appeal this decision, you may do so either directly with the commons project, or by contacting the ASF Infrastructure team at: [users@infra.apache.org](mailto:users@infra.apache.org)

@orionlibs 
Please apply again for a Jira account, I'll watch for it.

orionlibs:
thank you Gary. I just registered again for JIRA with the
email = ***@***.***
username = dimitrios.efthymiou

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
```diff
@@ -918,7 +918,8 @@ public void testIsParsable() {
         assertFalse(NumberUtils.isParsable("pendro"));
         assertFalse(NumberUtils.isParsable("64, 2"));
         assertFalse(NumberUtils.isParsable("64.2.2"));
-        assertFalse(NumberUtils.isParsable("64."));
+        assertTrue(NumberUtils.isParsable("64."));
```
elharo, line 921: please add a test that 64.. is false

elharo (changes requested):

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
orionlibs: @elharo sorry. My understanding of the ticket is that 64. is valid

elharo: 64. is valid but 64.. (two periods) is not

orionlibs: you are right. I added a test for it

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1776,7 +1773,7 @@ private static boolean withDecimalsParsing(final String str, final int beginIdx)
             if (decimalPoints > 1) {
                 return false;
             }
-            if (!isDecimalPoint && !Character.isDigit(str.charAt(i))) {
+            if (!isDecimalPoint && !Character.isDigit(str.charAt(i)) && !(i == str.length() - 1 && str.charAt(i) == '.')) {
```
garydgregory, line 1776: What happens in countries that flip the meaning of the comma and period?

orionlibs: perhaps there could be another method that takes a Locale object. Otherwise, I don't think the current parser should concern itself with it

elharo: Per current Javadoc "Parsable numbers include those Strings understood by Integer#parseInt(String), Long#parseLong(String), Float#parseFloat(String)} or Double#parseDouble(String)."

If those methods don't worry about countries that flip the meaning of the comma and period, we shouldn't do it here. That would be a breaking behavior change.

As to another method that uses Locales, that's in java.util so outside the advertised scope for commons-lang.

elharo (approved):

garydgregory:
To make things more complicated, in the French locale a non-breaking space can be used to separate some items. I can't recall the details now but it's yet another thing to lookout for.

elharo:
The more I think about this, the more I'm convinced we should not account for localization. This method is *not* for the purpose of determining whether a string is plausibly a number. Its purpose is to return true or false depending on whether the string can be parsed by one of the java.lang.Number classes. Those classes don't consider localization, so this one shouldn't either.

garydgregory:
@elharo 
I agree with you on the purpose of the API.
I'd like to revisit the implementation though: Why are we half-baked-with-bugs parsing the input instead of just try-catching what we claim is input that will or not throw an exception when we a user actually calls on the number APIs?

elharo:
One purpose of this method is precisely to avoid using exceptions for flow control.

orionlibs:
this PR is still open after half a year

PR #1072, opened 2023-06-29
LANG-1658

orionlibs:
created method that extracts numbers from a given string


garydgregory:
1: I really don't think https://issues.apache.org/jira/browse/LANG-1658 belongs in Commons Lang, maybe in Commons Number, or Commons Text, but not here. The best place to discuss this is on the dev mailing list.

PR #1075, opened 2023-07-01
ArchUtils refactoring

orionlibs:
The logic to get the Processor for a given architecture string is spread across the init_* methods. A cleaner design would be to have a separate private method that maps the architecture string to a Processor, and calls that from the init_* methods.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArchUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArchUtils.java
@@ -41,48 +41,22 @@ public class ArchUtils {
     }
 
     private static void init() {
-        init_X86_32Bit();
-        init_X86_64Bit();
-        init_IA64_32Bit();
-        init_IA64_64Bit();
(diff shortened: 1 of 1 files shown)


garydgregory:
@orionlibs Thank you for your PR.

-1: IMO, this is a case where you have to consider making it evident to people reading the code what it is we support. One method per processor type makes it obvious as soon as you look at the class. I would say that this is a rare case, where a tiny amount of code duplication (calling the ctor) is worth it because: The code is only invoked once (per class loader) and the method names serve as "code as documentation". Each private method _could_ have a Javadoc added to make it even nicer, but I'll leave that as a TODO for anyone that wants to do that.

orionlibs:
@garydgregory thank you. So, should I close this PR?

garydgregory:
> @garydgregory thank you. So, should I close this PR?

I think so. You can check the master branch for the current state.

PR #1076, opened 2023-07-01
refactored an addAll method in ArrayUtils

orionlibs:
refactored an addAll method in ArrayUtils to reuse code and defer the array1 component type check to the exception catch block


garydgregory (changes requested):
-1: Why is this better? `array2` can't be null on line 1120, so there is no point in calling `getComponentType()`. Maybe I am missing something.

PR #1077, opened 2023-07-01
refactored ArrayUtils to reuse the getLength(Object array) method

orionlibs:
refactored ArrayUtils to reuse the getLength(Object array) method inside the class instead of calling Array.getLength()

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -655,7 +655,7 @@ public class ArrayUtils {
             Array.set(joinedArray, 0, element);
             return joinedArray;
         }
-        final int length = Array.getLength(array);
+        final int length = getLength(array);
         if (index > length || index < 0) {
             throw new IndexOutOfBoundsException("Index: " + index + ", Length: " + length);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -655,7 +655,7 @@ private static Object add(final Object array, final int index, final Object elem
             Array.set(joinedArray, 0, element);
             return joinedArray;
         }
-        final int length = Array.getLength(array);
+        final int length = getLength(array);
```
garydgregory, line 658: -1: Why? `array` can't be null here.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -1668,7 +1668,7 @@ public static boolean containsAny(final Object[] array, final Object... objectsT
      */
     private static Object copyArrayGrow1(final Object array, final Class<?> newArrayComponentType) {
         if (array != null) {
-            final int arrayLength = Array.getLength(array);
+            final int arrayLength = getLength(array);
```
garydgregory, line 1671: -1: Makes no sense since you're in a block where `array` can't be null.

garydgregory (changes requested):
I think this PR should be closed, see comments.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
orionlibs: just for code reuse. This class already has a getLength method. Why not reuse it inside this class? That was my thought. Not nulls. But it is OK. I will close it

PR #1078, opened 2023-07-01
Update Javadoc for the insert methods in ArrayUtils

orionlibs:
updated the JavaDoc for the insert methods in ArrayUtils to say the methods return also null if the input array is null


garydgregory:
@orionlibs Merged, TY for the PR!

PR #1079, opened 2023-07-03
LANG-1647-correction-of-isUnchecked-method

orionlibs:
--corrected the isUnchecked method to check for null
--created test for null case


garydgregory:
@orionlibs 
In the future, or even now, please add a link in the description to the Jira ticket if there is one.

garydgregory:
Also don't use whatever is this "--" notation. GH will render lists with "-".

PR #1082, opened 2023-07-20
Add capacity for ReflectionToStringBuilder to include methods annotated with @ToStringInclude

shashankrnr32:
**Note: Please feel free to close this PR if this was previously discussed before and was decided not to be added to this implementation.**

Currently, `ReflectiveToStringBuilder` allows users to exclude fields from the `toString` output by adding an annotation `@ToStringExclude` to the field but there is no support to include methods declared in the class to be included in the `toString` output. 

This PR introduces a new annotation `@ToStringInclude(String)` that can be added to the methods of a class with which, the methods in the toString output.

Some of the points which can be considered
1. Add this implementation to a completely different class instead of modifying the current implementation
2. Wrap this inside a flag, say `allowReflectingMethods` before going into this added implementation.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/Reflection.java
+++ b/src/main/java/org/apache/commons/lang3/builder/Reflection.java
@@ -18,6 +18,8 @@
 package org.apache.commons.lang3.builder;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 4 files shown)


garydgregory:
All fileds are included by default, so why do we need to specify which ones to inclde?

shashankrnr32:
Yes. But the methods declared in a class are not included in the `toString` output. Lets say I want to exclude a field in the output and include a modified value (masked, abbreviated) that is generated by a method, we can include with such a feature.

shashankrnr32:
Adding a small example for more clarity. In the below example, lets say the usecase is to print "******" in place of the actual username (to mask the personally identifable information), this will help include the outputs of the methods by calling reflection invoke on the methods that are annotated with `@ToStringInclude`

```java

package org.apache.commons.lang3.builder;

import org.apache.commons.lang3.StringUtils;

class TestFixture {
    @ToStringExclude
    private String username;

    @ToStringExclude
    private String phoneNumber;

    @ToStringInclude
    private String toStringUsername() {
        return StringUtils.repeat("*", StringUtils.length(username));
    }

    @ToStringInclude
    private String toStringPhoneNumber() {
        return StringUtils.repeat("*", StringUtils.length(phoneNumber));
    }

    @Override
    public String toString() {
        return new ReflectionToStringBuilder(this).build();
    }
}
```

garydgregory:
Note: Don't use code like this example in production, you're revealing the length of secrets. This might not be harmful in itself, but, combined with other leaks, could prove fatal. Perhaps this a good enough reason to avoid providing a potential foot gun like this one.

shashankrnr32:
Agreed. Just a crude example to understand the use case. Couple of other uses I can think of here

1. A method that calculates something based on the values present in the class which needs to be added to the toString output
2. Inferred attributes
3. Masking only certain characters when the entire String masking is not required. 
4. A method that's annotated with `@JsonProperty` which will get serialized as JSON but is not printed in toString output

PR #1086, opened 2023-07-24
Honor `getValue` method of ReflectionToStringBuilder & its subclasses to get the field value

shashankrnr32:
Noticed this change recently which removed the usage of the method `getValue` to get the field value. Custom implementations of the `getValue` would not be used anymore.

@garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
@@ -658,7 +658,7 @@ public class ReflectionToStringBuilder extends ToStringBuilder {
             final String fieldName = field.getName();
             if (this.accept(field)) {
                 // Warning: Field.get(Object) creates wrappers objects for primitive types.
-                final Object fieldValue = Reflection.getUnchecked(field, getObject());
+                final Object fieldValue = this.getValue(field);
                 if (!excludeNullValues || fieldValue != null) {
                     this.append(fieldName, fieldValue, !field.isAnnotationPresent(ToStringSummary.class));
                 }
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hi @shashankrnr32 
Thank you for your PR!
This PR missing a unit test to avoid a future regression.

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -703,13 +703,11 @@ public Class<?> getUpToClass() {
      *
      * @throws IllegalArgumentException
      *             see {@link java.lang.reflect.Field#get(Object)}
-     * @throws IllegalAccessException
-     *             see {@link java.lang.reflect.Field#get(Object)}
      *
      * @see java.lang.reflect.Field#get(Object)
      */
-    protected Object getValue(final Field field) throws IllegalAccessException {
```
garydgregory, line 711: Why are you removing `IllegalAccessException`?

shashankrnr32:
> Hi @shashankrnr32 Thank you for your PR! This PR missing a unit test to avoid a future regression.

@garydgregory Added `ReflectionToStringBuilderCustomImplementationTest` that overrides the `getValue` and modifies the generated field value by the base class. If this behaviour changes in the future, the test will break.

garydgregory:
https://issues.apache.org/jira/projects/LANG/issues/LANG-1710

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
garydgregory: Hello @shashankrnr32 
Thank you for your update.
This PR attempts to restore compatibility with the `getValue()` method but changes that method's definition and breaks compatibility differently. For example, if I have a subclass that overrides `getValue()` using the released signature, that subclass will no longer compile if this patch is applied. The test in this PR does not reflect that a subclass can throw `IllegalAccessException` in the `getValue()` method signature, so the test is not checking for full compatibility. See git master for a fix.

garydgregory:
Closing, see the comment.

PR #1087, opened 2023-07-27
Refactor rename the variable name from 'clss' to 'clazz'

remeio:
Avoid using reserved word `class`, so `clss`. 
`clazz` is better? :)


garydgregory:
'clazz' is commonly used.

PR #1095, opened 2023-08-14
NumberUtils countSignificantFigures method

theshoeshiner:
A util method in NumberUtils that counts the number of significant figures in a decimal strings of standard / exponential format. e.g.(101, 102.1, 5.1e2, 1.001, 0.2e4, etc...)

This method counts the number of "significant figures" according to standard mathematical rules (details and references are in the javadoc). It must operate on a string because the rules themselves inherently depend on the string representation and not the underlying numeric value. The rules are well documented and widely accepted, aside from zero values, which are not usually covered by the conventions and the choice as to how to interpret is documented in the javadoc.

My first instinct was that this belonged in commons-math or commons-numbers, as the concept of significant figures is usually necessary in scientific / mathematical fields, but thus far it doesnt seem to be a good fit. Decided Id create the PR here so that at least the implementation is out there for review and perhaps this discussion will find it a better home. I do feel like this is a valuable feature because any manipulation of numerical scientific data that needs to be accurate (e.g. clinical data) ultimately needs to respect significant figure counts.

There are 7 basic rules that the method follows, #1-5 being the core sigfig rules that you'll find on most external sources, #6 to explain how zero values are handled, and #7 to explain how exponents are handled (or excluded, rather). The overall format of the string is validated, with IAEs thrown for invalid characters or formatting.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -22,6 +22,7 @@ import java.math.BigInteger;
 import java.math.RoundingMode;
 import java.util.Objects;
 
+import org.apache.commons.lang3.CharUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 
@@ -32,6 +33,14 @@ import org.apache.commons.lang3.Validate;
  */
 public class NumberUtils {
 
+    /** Reusable char constant for zero. */
+    public static final char CHAR_ZERO = '0';
+    /** Reusable char constant for decimal separator character period. */
+    public static final char CHAR_DECIMAL_PERIOD = '.';
+    /** Reusable char constant for decimal separator character comma. */
+    public static final char CHAR_DECIMAL_COMMA = ',';
+    /** Reusable char constant for exponent character (lowercase e). */
+    public static final char CHAR_EXPONENT = 'e';
     /** Reusable Long constant for zero. */
     public static final Long LONG_ZERO = Long.valueOf(0L);
     /** Reusable Long constant for one. */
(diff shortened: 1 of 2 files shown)


aherbert:
This cannot parse standard floating point numbers generated by String valueOf as the exponent character in that case is E.

What if the exponent is negative?

What if the number is negative?

What if either number or exponent have an explicit + sign?

What if the number has more than 1 decimal point?

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -32,6 +33,20 @@
  */
 public class NumberUtils {
 
+    /** Reusable char constant for zero. */
+    public static final char CHAR_ZERO = '0';
```
elharo, line 37: These are all superfluous. Named constants are for number, not string and char literals.

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1847,4 +1862,211 @@ public static int compare(final short x, final short y) {
     public static int compare(final byte x, final byte y) {
         return x - y;
     }
-}
+
+    /**
+     * Counts the number of significant figures within a decimal value string.
+     * {@link org.apache.commons.lang3.math.NumberUtils#countSignificantFigures(String, char)}
+     *
+     * @param decimalNumber The decimal number string. Must be numeric digits with optional decimal
```
elharo, line 1870: Please follow Oracle guidelines for formatting doc comments

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1847,4 +1862,211 @@ public static int compare(final short x, final short y) {
...
+     * and optional integer exponential suffix
+     * @return The number of significant figures in the string
+     */
+    public static int countSignificantFigures(String decimalNumber) {
+        return countSignificantFigures(decimalNumber, CHAR_DECIMAL_PERIOD);
+    }
+
+    /**
+     * Counts the number of significant figures within a decimal value string. This method must
```
elharo, line 1879: Need a link to an authoritati=ve definition  o0f signifcant figures

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1847,4 +1862,211 @@ public static int compare(final short x, final short y) {
...
+     * accept a string because significant figures can only be inferred from the a user provided
```
elharo, line 1880: delete the

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1847,4 +1862,211 @@ public static int compare(final short x, final short y) {
...
+     * Figures</a>
+     *
+     * @param decimalNumber The decimal number string. Must be optional sign + numeric digits with optional decimal
+     * and optional e|E + sign + integer exponential suffix
+     * @param decimalSeparator the decimal separator character to use
+     * @return The number of significant figures in the string
+     */
+    public static int countSignificantFigures(String decimalNumber, char decimalSeparator) {
+        if (decimalNumber.length() == 0) {
+            throw new IllegalArgumentException("Decimal Number string was empty");
```
elharo, line 1962: needs a doc comment throws clause

src/main/java/org/apache/commons/lang3/math/NumberUtils.java
```diff
@@ -1847,4 +1862,211 @@ public static int compare(final short x, final short y) {
...
+        }
+        if (CharUtils.isAsciiNumeric(decimalSeparator)) {
+            throw new IllegalArgumentException("Decimal Separator cannot be numeric");
+        }
+        if (decimalSeparator == CHAR_EXPONENT_UPPER || decimalSeparator == CHAR_EXPONENT_LOWER) {
```
elharo, line 1967: inline the constants and I could read this without scrolling

elharo (changes requested):

garydgregory:
Doesn't this belong in Commons Math?

PR #1100, opened 2023-08-24
Fix ThresholdCircuitBreaker

yichinzhu:
According to the document, checkState() should return `true` if the circuit breaker is currently closed。

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/ThresholdCircuitBreaker.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/ThresholdCircuitBreaker.java
@@ -89,7 +89,7 @@ public class ThresholdCircuitBreaker extends AbstractCircuitBreaker<Long> {
      */
     @Override
     public boolean checkState() {
-        return isOpen();
+        return !isOpen();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
-1, the docs read "<strong>true</strong> means that the circuit breaker is currently open"

yichinzhu:
> -1, the docs read "**true** means that the circuit breaker is currently open"

The comments in CircuitBreaker.java shows that "The return value indicates whether the circuit breaker is now in state closed a value  of true typically means that the current operation can continue." .
And EventCountCircuitBreaker also returns true for CLOSED state as an implementation

links:

https://github.com/apache/commons-lang/blob/2fb38b1bb76a0a33d72864c8e39b536bce6b39e7/src/main/java/org/apache/commons/lang3/concurrent/CircuitBreaker.java#L61-L68

https://github.com/apache/commons-lang/blob/2fb38b1bb76a0a33d72864c8e39b536bce6b39e7/src/main/java/org/apache/commons/lang3/concurrent/EventCountCircuitBreaker.java#L336

garydgregory:
Hi @yichinzhu 
Thank  you for the clarification, you are correct! :-)

PR #1102, opened 2023-08-27
LANG-1711: Modify ReflectionToStringBuilder to use try set accessible. Use Java 9 and above

OksiBlack:
- Modify ReflectionToStringBuilder to use try set accessible instead of set accessible.
- Use Java 9 and above


garydgregory:
-1 don't change the compiler version.

OksiBlack:
Understood, than it is not possible, unfortunately, to use this api, sadly though.

garydgregory:
Closing: No action on request.

PR #1105, opened 2023-09-08
add NPE testcase in LockingVisitorsTest

sujin0529:
Hi, I found the test case triggering the NPE bug in LockingVisitors.
https://github.com/apache/commons-lang/blob/8c6d874637b2e3342cbc1770c9a61ff184e483b0/src/main/java/org/apache/commons/lang3/concurrent/locks/LockingVisitors.java#L323-L325
I think this bug needs to be fixed.
Please check this test case.

Thanks.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/concurrent/locks/LockingVisitorsTest.java
+++ b/src/test/java/org/apache/commons/lang3/concurrent/locks/LockingVisitorsTest.java
@@ -30,6 +30,7 @@ import java.util.function.LongConsumer;
 import org.apache.commons.lang3.AbstractLangTest;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.ThreadUtils;
+import org.apache.commons.lang3.concurrent.locks.LockingVisitors;
 import org.apache.commons.lang3.concurrent.locks.LockingVisitors.LockVisitor;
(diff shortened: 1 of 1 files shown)


garydgregory:
Hello @sujin0529 
Fix? How? Wouldn't "garbage in, garbage out" apply h we're? What would the behavior be for a null lock?

sujin0529:
Hi, @garydgregory

sujin0529:
Hi, @garydgregory 
In this project, the guard conditioning is well done. However, there is no guard for this function `ReadWriteLockVisitor.<init>`. So I thought it would be good to add guard condition in the same way as below.
https://github.com/apache/commons-lang/blob/14be5e577767aa959510f7ebda8dc9647568a789/src/main/java/org/apache/commons/lang3/ArrayUtils.java#L769-L779
I think this error is closer to `IllegalArgumentException` than `NullPointerException`.

Thanks.

garydgregory:
Hello @sujin0529 

>I think this error is closer to IllegalArgumentException than NullPointerException.

I disagree: `NullPointerException` is more _precise_ than `IllegalArgumentException`. Note that `Objects.requireNonNull()` throws NPE.  What I can imagine is a better exception _message_ for the NPE.

You've not replied to: "What would the behavior be for a null lock?"
Are you arguing that the null input should throw an exception or not?

garydgregory:
Closing: No reply.

PR #1110, opened 2023-09-10
[LANG-1705] SerializationUtils.clone() throws a ClassCastException when override writeReplace()

a1batr0ss:
JIRA: https://issues.apache.org/jira/browse/LANG-1705

Due to the possibility that the serialized object type may be a different type after overriding writeReplace(), directly converting the serialized object to the original type(the type of parameter) will result in a ClassCastException.

Fix: Two templates are used for the return value and parameters of SerializationUtils.clone().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -624,7 +624,7 @@
 
     <checkstyle.configdir>src/site/resources/checkstyle</checkstyle.configdir>
 
-    <japicmp.skip>true</japicmp.skip>
+    <japicmp.skip>false</japicmp.skip>
     <!-- 0.6.0 requires Java 11 -->
     <commons.spdx.version>0.5.5</commons.spdx.version>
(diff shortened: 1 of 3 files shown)


garydgregory:
@a1batr0ss 
Please run `mvn` -- which uses the default goal -- locally before pushing and you'll catch build failures like what just happened.

pom.xml
```diff
@@ -624,7 +624,7 @@
 
     <checkstyle.configdir>src/site/resources/checkstyle</checkstyle.configdir>
 
-    <japicmp.skip>false</japicmp.skip>
+    <japicmp.skip>true</japicmp.skip>
```
garydgregory, line 627: -1, you cannot disable this check within a major release line.

garydgregory (changes requested):
-1, you cannot disable the JApiCmp check within a major release line.

garydgregory:
Closing: No reply.

PR #1114, opened 2023-09-25
Add TypedEqualsBuilder class and test

Cousnouf:
This class allows to have a more complete equals builder that takes in consideration the null and class comparison before the appended fields.
It makes the equals method more compact and avoids IDE warning when we do not check nullity and class in the equals method.


ecki:
Change looks quite good and complete. Not sure about the size, it might require a Apache contributor agreement. With this PR and Dev mail you probably don’t need an extra JIRA (or a committer can do that).

Do you have an contributors agreement on file, by any chance? https://www.apache.org/licenses/contributor-agreements.html

garydgregory:
Hello @Cousnouf 
Please run `mvn` locally and fix issues before you push to avoid broken builds.

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.lang3.builder;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * An extension of {@link EqualsBuilder} aimed to perform additionally the base objects and class equals checks.
```
garydgregory, line 23: Documentation needs a rework. I can't understand what this means.

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+        Class<T> currentInstanceClass = (Class<T>) currentInstance.getClass();
+        if (other == null || currentInstanceClass != other.getClass()) {
+            isEquals = false;
+            return;
+        }
+        this.other = (T) other;
+    }
+
+    @Override
+    boolean shouldLeaveEarly() {
```
garydgregory, line 66: Leave what early? This methods needs a better name IMO?

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+        return sameReference || super.shouldLeaveEarly();
+    }
+
+    public TypedEqualsBuilder<T> append(Function<T, ?> extractor) {
```
garydgregory, line 70: What if I my function throws an exception? Should this use `FailableFunction` instead?

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+ *         .append(TestObject::getB)
+ *         .isEquals();
+ *     }
+ * </pre>
+ *
+ * @param <T> the type of the compared object.
+ *
+ * @since 3.14.0
+ */
+public class TypedEqualsBuilder<T> extends EqualsBuilder {
```
garydgregory, line 42: This is a confusing class name IMO. What is and where is the Type? It's not a Java `Type` or `Class` apparently...

garydgregory (changes requested):
Needs discussion. See my comments. The main comment would be the: Why is this a new class? The class name and docs are confusing. I should not have to read a code example (thankfully in the Javadoc at least) to try to understand what this does. I may not be a native English speaker but this means nothing to me: "An extension of {@link EqualsBuilder} aimed to perform additionally the base objects and class equals checks."

src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
```diff
@@ -198,7 +198,7 @@ private static void unregister(final Object lhs, final Object rhs) {
      * If the fields tested are equals.
```
elharo, line 198: equals --> equal

src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
```diff
@@ -450,6 +450,15 @@ public static boolean reflectionEquals(final Object lhs, final Object rhs, final
                     .isEquals();
     }
 
+    /**
+     * Indicates if an early append method leave should be done
```
elharo, line 454: unclear, rephrase

src/test/java/org/apache/commons/lang3/builder/TypedEqualsBuilderTest.java
```diff
@@ -0,0 +1,89 @@
...
+    void test_complete_equals() {
+        TestObject testObject1 = new TestObject(1, 2);
+        TestObject testObject2 = new TestObject(1, 3);
+        TestObject testObject3 = new TestObject(1, 2);
+
+        assertEquals(testObject1, testObject1);
+        assertNotEquals(testObject1, testObject2);
+        assertNotEquals(testObject2, testObject1);
+
+        assertEquals(Boolean.TRUE,
```
elharo, line 35: probably use assertTrue and assertFalse, unless perhaps there's some auto-boxing detail I'm missing

src/test/java/org/apache/commons/lang3/builder/TypedEqualsBuilderTest.java
```diff
@@ -0,0 +1,89 @@
...
+                new TypedEqualsBuilder<>(testObject1, null)
+                        .append(TestObject::getA)
+                        .isEquals());
+        assertEquals(Boolean.FALSE,
+                new TypedEqualsBuilder<>(testObject1, "")
+                        .append(TestObject::getA)
+                        .isEquals());
+    }
+
+    static class TestObject {
```
elharo, line 64: private

elharo (changes requested):
I do not understand the point of this PR. Please file an issue in the JIRA and add documentation in this PR.

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
ecki: It is the class type. its refering to the T class parameter which is supported by this Builder subtype.

src/test/java/org/apache/commons/lang3/builder/TypedEqualsBuilderTest.java
Cousnouf: It was done to make the test slightly more quick to read with of the syntax coloration.

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
Cousnouf: Yes it is a java Class. But maybe I should rename it differently or extend the EqualsBuilder and add some special appending.... I will try that...

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
Cousnouf: It's better not to use the FailableFunction as it changes the signature of the method, hence forces you to surround with try/catch the build in your equals method.
image

Cousnouf: If the function throws a checked exception it can use `FailableFunction`, yes. I'll catch the Exception class (Throwable are meant to not be caught)

src/test/java/org/apache/commons/lang3/builder/TypedEqualsBuilderTest.java
elharo: In general, do not assume devs are using IDEs. E.g. right now I'm reading this code in Github, and it's less clear than assertTrue

elharo (changes requested):

src/test/java/org/apache/commons/lang3/builder/TypedEqualsBuilderTest.java
Cousnouf: GitHub has also syntax coloration. But yes let's change these asserts.

Cousnouf:
I've addressed the comments. This version may tend to something better :)

src/main/java/org/apache/commons/lang3/builder/TypedEqualsBuilder.java
Cousnouf: This equals builder is typed. We provide a type and then the field functions to build the equals method.
So the Type is a Java Class.

garydgregory: > It is the class type. its refering to the T class parameter which is supported by this Builder subtype.

✅️ Thanks for pointing that out! 👍

Cousnouf:
Hello all,

I've added some changes to address your comments.

It's still in a separate class as I think it's better to separate the
concept of EqualsBuilder with concrete values passed each time with append
and the concept of TypedEqualsBuilder which receive the two objects in the
construction and then calls the failable function during each appending.

Inputs welcome :)

Regards,

Marc

Le jeu. 28 sept. 2023 à 15:39, Gary Gregory ***@***.***> a
écrit :

elharo (changes requested):
1. This certainly requires a CLA
2. There needs to be a JIRA and a clear proposal both for what problem is being solved, and how this proposal solves it.
3. I might see what problem is being solved here — I'm not sure — but if so this is not the way to go about solving it and this is not the right API.

That is, please focus on documenting and getting agreement on the plan before writing any more code. As is, it's hard to fairly consider this since we don't understand it fully.

PR #1117, opened 2023-10-03
Check for null value with the appropriate assertion method

valery1707:
Replace all occurrences of `Validate.isTrue\((\w+)\s?!=\s?null\s*,` with `Validate.notNull($1,`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/FieldUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/FieldUtils.java
@@ -473,7 +473,7 @@ public class FieldUtils {
         Objects.requireNonNull(target, "target");
         final Class<?> cls = target.getClass();
         final Field field = getField(cls, fieldName, forceAccess);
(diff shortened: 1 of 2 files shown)


aherbert:
This is not fixing any bugs, and introduces behaviour changes.

The behavioural change is to throw NPE instead of IAE when an argument to reflect on a `target` using `FieldUtils` has created an error. The methods are only documented to throw an NPE if the `target` is null. All other arguments when used incorrectly are documented to throw a IAE.

The current code, although internally appears to be incorrect by using `ValidationsUtils.isTrue(something != null, ...)`, is behaving as documented. If you only read the javadoc, and not the source code, the methods are working as specified.

garydgregory:
-1
I agree with @aherbert 
Throwing an NPE for null input to a method makes perfect sense. In this case, this PR throws an NPE because internal processing happens to see a null value. The existing code correctly interprets that this null intermediary value is caused by bad input to the method and therefore properly throws `IllegalArgumentException`. Closing.

PR #1119, opened 2023-10-09
Create LazyInitializerWithDisposer and test

benjamin-confino:
This Pull Request creates a variant of LazyInitializer that also allows you to dispose of the wrapped object.

The use case is if you have an object that might or might not be needed; but if it is created it needs to be disposed. LazyInitalizer does not allow you to check if the wrapped object has been created, so the only way to dispose it is to call get() then invoke a disposal method on the acquired object. Creating the wrapped object just so you can dispose it defeats the point, so this new class allows for this use case.

Here are the design decisions I made, and the rationales for them: 

- Using FailableConsumer and FailableSupplier instead of making LazyInitializerWithDisposer an abstract class like LazyInitializer.
- - I think being able to create a new LazyInitializerWithDisposer in one line using Lambdas or `::` method reference operators provided enough of an advantage to justify using a different design to LazyInitializer.
- Not having any way to reset the object after calling `dispose()` or `close()`
- - I chose this because it makes it easier to reason about the lifecylce of a LazyInitializerWithDisposer. If you ever observe that it has been disposed you never have to worry about the possibility it will be reset elsewhere. I thought this was more valuable than the ability to push a change in state to all objects with a reference to LazyInitializerWithDisposer.
- Using AlreadyDisposedException instead of ConcurrentException
- - Even though I'm not pleased with having the statement `throws ConcurrentException` refer to both AlreadyDisposedException and ConcurrentException I thought that it was useful enough to have separate catch blocks for when the Initalizer is disposed and when something goes wrong inside the supplier or consumer to justify this.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/LazyInitializerWithDisposer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/LazyInitializerWithDisposer.java
@@ -1,0 +1,294 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.lang3.concurrent;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.Optional;
+
+import org.apache.commons.lang3.function.FailableConsumer;
+import org.apache.commons.lang3.function.FailableSupplier;
+
+/**
+ * This class provides a generic implementation of the lazy initialization
+ * pattern and also contains a way to dispose of the wrapped object if it has
+ * been previously created by using a supplier and a consumer respectively.
+ * To ensure the supplier and disposer are only used once each this class 
+ * uses the double-check idiom for an instance field as discussed in  Joshua 
+ * Bloch's "Effective Java", 2nd edition, item 71. 
+ *
+ * <p>
+ * Sometimes an application has to deal with an object only under certain
+ * circumstances, e.g. when the user selects a specific menu item or if a
+ * special event is received. If the creation of the object is costly or the
+ * consumption of memory or other system resources is significant, it may make
+ * sense to defer the creation of this object until it is really needed. This is
+ * a use case for the lazy initialization pattern.
+ * </p>
+ *
+ * <p>
+ * If these objects must be disposed of, it would not make sense to create them
+ * just so they can be disposed of. Therefore this class provides the ability to
+ * dispose of objects if and only if they have been created.
+ * </p>
+ *
+ * <p>
+ * Access to the data object is provided through the {@code get()} method. and
+ * the data object is disposed with the {@code dispose()} So, code that obtains
+ * and eventually disposes of the {@code ComplexObject} instance would simply look
+ * like this:
+ * </p>
+ *
+ * <pre>
+ * // Create the supplier and disposer: 
+ * Supplier<ComplexObject> initializer = () -> new ComplexObject();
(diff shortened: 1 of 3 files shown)


elharo (changes requested):
New features should start as JIRA issues and might need discussion on the dev mailing list

benjamin-confino:
As requested, I have opened a JIRA issue at https://issues.apache.org/jira/browse/LANG-1716

garydgregory:
Can't we merge the feature into the existing class? A new class feels over the top... needs studying...

src/main/java/org/apache/commons/lang3/concurrent/LazyInitializerWithDisposer.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /** 
+     * Constructs a LazyInitializerWithDisposer wtih a given initializer and disposer
+     *
+     * @param initializer an implimentation of the FailableSupplier functional interface which will create the wrapped object
+     * @param disposer an implimentation of the FailableConsumer functional interface which will dispose of the wrapped object
+     */
+    public LazyInitializerWithDisposer(FailableSupplier<? extends T, ? extends Exception> initializer, FailableConsumer<? super T, ? extends Exception> disposer) {
+        this(initializer, disposer, 4);
```
garydgregory, line 122: 4 seems random to me. It's not clear to me why this would be needed. Closable and AutoCloseable don't support anything like that for example.

benjamin-confino: There is a logic to that 4. Three retries is a standard default. And I'm using a variable representing tries rather than retries to distinguish between 0 tries left - a fail state - and -1 or below representing infinite tries.

As for why have retries at all. LazyInitializers are most useful when it is expensive to initialize something. Needing some form of IO is both a common reason for an operation to be expensive, and a case where trying again after a failure might actually work. So I think it is useful to make retries possible.

elharo:
My initial thought is that this is too esoteric to be worth the added API surface and maintenance burden. Is there any existing code or project you can show that would benefit heavily from this?

src/main/java/org/apache/commons/lang3/concurrent/LazyInitializerWithDisposer.java
```diff
@@ -0,0 +1,294 @@
...
+                }
+
+                try {
+                    object = (T) DISPOSED;
+                    disposer.accept(disposable);
+                    return true;
+                } catch (RuntimeException e) {//So it doesn't get wrapped in the next block.
+                    throw e; 
+                } catch (Exception e) {
+                    throw new ConcurrentException(e);
```
garydgregory, line 262: Just FYI: `throw new ConcurrentException(ExceptionUtils.throwUnchecked(e));`

garydgregory:
> My initial thought is that this is too esoteric to be worth the added API surface and maintenance burden. Is there any existing code or project you can show that would benefit heavily from this?

I agree with @elharo. 

There is also the question of the other `ConcurrentInitializer` implementations like `AtomicInitializer`, `AtomicSafeInitializer`, and so on.

One scenario that is worth handling though is: "My app is shutting down, I want to release all resources, but I can't release what's in my lazy-initlializer without allocating it in the first place." As a simple and pragmatic solution to this issue, I think it is worth adding a tester method `LazyInitializer#isInitialized()`, so that an app can check to see if calling `get()` would allocate the (expensive) resource. Then, we can see which of the other `ConcurrentInitializer` implementations can provide an `#isInitialized()` method. Then we can consider pulling-up `#isInitialized()` to `ConcurrentInitializer` if indeed all `ConcurrentInitializer` can provide a sensible implementation.

We should also consider a much simpler implementation and much reduced feature set to see how that would handle @benjamin-confino's use cases. The above might be enough for the stated use case.

One feature I do not like is for the new initializer class to track state. Especially WRT disposal/closing. Only the receiver of the close call (I'm not calling it "dispose"), really knows if the call was processed normally, even if no exception was thrown.

I do like the idea of providing a supplier as instead of subclassing, so I think I'll experiment with that as well. I also like the idea of a closer function. IOW, all of this makes up a few discrete features worth considering, but not in one giant PR IMO, especially since we have many `ConcurrentInitializer` implementations and it would be nice to provide some level of consistency where sensible.

benjamin-confino:
Defaulting to a `new Object()` is the plan. Not only dose it help with write `isInitialized()`, it will fix a bug I spotted in `AtomicReference`. As per the `AtomicReference` javadoc

> the algorithm outlined above guarantees that {@link #get()} always returns the same object though.

However if you had an initializer method that returns null in some cases (lets say it reads a value from a network and returns null on a network error) you could call get() twice, get a null the first time, and a non-null the second time. 

Its late in my part of the world but I should be able to on `isInitialized()` tomorrow.

garydgregory:
Closing, replaced by #1123. TY @benjamin-confino ! 🥇

PR #1120, opened 2023-10-13
Add method ConcurrentInitializer#isInitialized()

benjamin-confino:
This is associated with https://issues.apache.org/jira/browse/LANG-1716

See also discussion on #1119

------

This pull request adds a method `isInitialized()` to all implementations of ConcurrentInitializer. 

Here are the design decisions I made, and the rationales for them:
* Using `NO_INIT` for `AtomicInitializer` and `AtomicSafeInitializer`
* * Apart from making it easier to write an `isInitialized()` method this ensures that if `Atomic*Initializer` returns null it will correctly cache that behaviour as the docs say it should. This is a bug fix, but it may break backwards compatibility if someone inadvertently depends on the old behaviour. I chose to the correct thing now, and am drawing attention to it here to discuss if we actually want this fix.
* Including `isInitialized()` in `ConstantInitializer`
* * This is not strictly needed but I included it just for consistency.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
@@ -63,8 +63,10 @@ import java.util.concurrent.atomic.AtomicReference;
  * @param <T> the type of the object managed by this initializer class
  */
 public abstract class AtomicInitializer<T> extends AbstractConcurrentInitializer<T, RuntimeException> {
+
+    private final T NO_INIT = (T) new Object(){};
     /** Holds the reference to the managed object. */
-    private final AtomicReference<T> reference = new AtomicReference<>();
+    private final AtomicReference<T> reference = new AtomicReference<>(NO_INIT);
 
     /**
      * Returns the object managed by this initializer. The object is created if
@@ -79,9 +81,9 @@ public abstract class AtomicInitializer<T> extends AbstractConcurrentInitializer
     public T get() throws ConcurrentException {
         T result = reference.get();
 
-        if (result == null) {
+        if (result == NO_INIT) {
             result = initialize();
-            if (!reference.compareAndSet(null, result)) {
+            if (!reference.compareAndSet(NO_INIT, result)) {
                 // another thread has initialized the reference
                 result = reference.get();
             }
--- a/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
@@ -64,7 +64,7 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public abstract class AtomicInitializer<T> extends AbstractConcurrentInitializer<T, RuntimeException> {
 
-    private static final Object NO_INIT = new Object(){};
+    private static final Object NO_INIT = new Object();
     /** Holds the reference to the managed object. */
     private final AtomicReference<T> reference = new AtomicReference<>((T) NO_INIT);
 
--- a/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
@@ -64,9 +64,9 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public abstract class AtomicInitializer<T> extends AbstractConcurrentInitializer<T, RuntimeException> {
 
-    private final T NO_INIT = (T) new Object(){};
+    private static final Object NO_INIT = new Object(){};
     /** Holds the reference to the managed object. */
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java
```diff
@@ -63,8 +63,10 @@
  * @param <T> the type of the object managed by this initializer class
  */
 public abstract class AtomicInitializer<T> extends AbstractConcurrentInitializer<T, RuntimeException> {
+
+    private final T NO_INIT = (T) new Object(){};
```
garydgregory, line 67: - Should be `static`, we don't need a new `Object` instance for each new initializer.
- Why is the Object an anonymous class? IOW: `new Object(){};` --> `new Object();`

src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
```diff
@@ -53,12 +53,13 @@
  */
 public abstract class AtomicSafeInitializer<T> extends AbstractConcurrentInitializer<T, RuntimeException> {
 
+    private final T NO_INIT = (T) new Object(){};
```
garydgregory, line 56: - Should be `static`, we don't need a new `Object` instance for each new initializer.
- Why is the Object an anonymous class? IOW: `new Object(){};` --> `new Object();`

garydgregory (changes requested):
Hello @benjamin-confino 

Thank you for your PR.

I've fixed all Checkstyle errors. Run `mvn` by itself to run the default Maven goal which includes all of our build checks. Doing this before you push will avoid failing builds.

Please see my other comments in 2 files.

Also, Codecov reports that coverage is down, it would be best to avoid that.

Locally but not Codecov, you can run `mvn clean install site` and look into `target\site` for the JaCoCo code coverage report. To be really nice to have new code as close to 100% covered by tests when reasonably possible. It's also possible that the delta in code coverage reports are just noise, I've not looked in this case.

TY!
Gary

src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
benjamin-confino: Good call on static.

Object is an anonymous class because that way it will show up as `AtomicSafeInitializer$1` instead of `Object` in a debugger, which makes it easier to distinguish between `NO_INIT` and a return value of type `Object` when debugging.

benjamin-confino:
I've added static to the previously mentioned fields, added new tests to fix the reduced code coverage, and squashed everything back down to two commits.

garydgregory:
@benjamin-confino 
Uh? Did you even try to compile the code?
```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/commons-lang/commons-lang/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java:[67,26] non-static type variable T cannot be referenced from a static context
Error:  /home/runner/work/commons-lang/commons-lang/src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java:[56,26] non-static type variable T cannot be referenced from a static context
Error:  /home/runner/work/commons-lang/commons-lang/src/main/java/org/apache/commons/lang3/concurrent/AtomicInitializer.java:[67,39] non-static type variable T cannot be referenced from a static context
Error:  /home/runner/work/commons-lang/commons-lang/src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java:[56,39] non-static type variable T cannot be referenced from a static context
```

benjamin-confino:
How embarrassing, this should fix it.

src/main/java/org/apache/commons/lang3/concurrent/AtomicSafeInitializer.java
garydgregory: WRT `new Object(){};` vs `new Object();`, IMO, that's not enough of a reason to force an anonymous subclass into the runtime. It's also different from what `LazyInitializer` does, which is confusing to the maintainer at large (especially when uncommented). I'll approve the PR but use `new Object();`.

garydgregory:
@benjamin-confino 
Thank you! Merged.

benjamin-confino:
@garydgregory 

This was a nice news to start a Monday. Let me know if you want me to make a PR for supporting suppliers next, or one for adding a close method. If not we can close the other PR and the JIRA issue.

garydgregory:
> @garydgregory
> 
> This was nice news to start a Monday. Let me know if you want me to make a PR for supporting suppliers next, or one for adding a close method. If not we can close the other PR and the JIRA issue.

Hi @benjamin-confino 
Please see https://github.com/apache/commons-lang/pull/1123
Let me know what you think.

PR #1123, opened 2023-10-16
Use LazyInitializer without subclassing.

garydgregory:
- Allow a Supplier for initialized object
- Allow a Consumer to close the managed object

### DIFF
--- a/src/main/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializer.java
+++ b/src/main/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializer.java
@@ -20,6 +20,7 @@ package org.apache.commons.lang3.concurrent;
 import java.util.Objects;
 
 import org.apache.commons.lang3.builder.AbstractSupplier;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.function.FailableConsumer;
 import org.apache.commons.lang3.function.FailableSupplier;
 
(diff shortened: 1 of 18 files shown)


garydgregory:
CC: @benjamin-confino

src/main/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializer.java
```diff
@@ -26,15 +32,128 @@
...
+            }
+        }
+    }
+
     /**
      * Creates and initializes the object managed by this {@code
      * ConcurrentInitializer}. This method is called by {@link #get()} when the object is accessed for the first time. An implementation can focus on the
      * creation of the object. No synchronization is needed, as this is already handled by {@code get()}.
+     * <p>
+     * Subclasses and clients that do not an initializer are expected to implement this method.
```
benjamin-confino, line 148: There is a missing word "provide" after "do not".

benjamin-confino:
I do have one thought. Currently when you create a supplier for LazyInitializer the type signature means that it must declare no exceptions or declare `throws ConcurrentException`. I think this is a serious limitation. It means you cannot do something like `LazyInitializer.builder().setInitializer(SomeLibrary::methodWithCheckedException)` without writing boilerplate code to catch that exception and wrap it. 

This can be solved by making the supplier and consumer take an exception of `? extends Exception` and then casting it as necessary inside `initialize()`. I think the cleanest solution would be to have an abstract method like `getExceptionType` that initialize could use to tell what it needs to cast an exception into. I'll write it locally and test to see if it works.

garydgregory:
I agree with your point on the less elegant throwing of ConcurrentException. The issue is that we have three methods that can throw: get, initialize and close. Each could throw its own different Exception. I am trying to avoid declaring 3 generic parameter types and keep with just one. This might not be possible but OTOH the 3 parameter variant might be "too much".

src/main/java/org/apache/commons/lang3/concurrent/AbstractConcurrentInitializer.java
garydgregory: Fixed, ty.

benjamin-confino:
I've just finished testing locally, I'll PR your branch.

benjamin-confino:
The PR is here: https://github.com/garydgregory/commons-lang/pull/1

Regarding `get`, `initialize` and `close`. I think it makes sense to use one shared type. `ConcurrentException` is a wrapper around the real exception so its reasonable that even if `get`, `initialize` and `close` return three different exceptions they all use the same wrapper type. 

(I didn't remove the recast to IllegalStateException because I wanted to ask why it was there first?)

For `BackgroundInitializer` I think the sensible thing to do is to use a `ConcurrentException` for `get()` and `close()` but not recast any exception during `initialize()` since that will be wrapped by Future anyway. 

The above is for the outputs. For the inputs, I think the Supplier and Consumer should accept possible exception deceleration so you can plug in library methods without glue code, and because I think boilerplate like placing the real exception in a wrapper should be done in one central location if possible.

PR #1136, opened 2023-11-20
Updates javadoc of ExceptionUtils.getRootCause()

palacsint:
getRootCause() does not return null anymore when the input throwable has a cause of itself. The behavior was changed in commit 60412131f.


garydgregory:
@palacsint 
Thank you for your PR.
I committed changes elsewhere in the class to improve the text. You are credited in `changes.xml`. The method you edited was still hard to read and understand (to me) as were other methods. Hopefully, the text is better now in several of the related methods.
If you want more changes in this area, rebase this PR on git master and go from there. Please close this PR if all is well.
TY!

palacsint:
@garydgregory : Thanks, it's much better with your changes!

PR #1137, opened 2023-11-23
Add an ability to compare equals by equals() method objects with ReflectionBuilder

yurbar:
Useful when need to compare JPA entities with the same identifier.

Example:  OptimisticLockingException thrown and you need to recover and print what's a difference between two entities ts with the same identifiers. You can't do it with the ReflectionDiffBuilder now because it will print nothing because those objects are equals by equals() method. So with this change you can have such an ability for ReflectionDiffBuilder

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -75,6 +75,7 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
 
     private final T left;
     private final T right;
+    private final boolean testTriviallyEqual;
     private final DiffBuilder<T> diffBuilder;
 
     /**
@@ -103,9 +104,38 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
      *             if {@code lhs} or {@code rhs} is {@code null}
      */
     public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style) {
+        this(lhs, rhs, style, true);
+    }
+
+    /**
+     * Constructs a builder for the specified objects with the specified style.
+     *
+     * <p>
+     * If {@code lhs == rhs} or {@code lhs.equals(rhs)} then the builder will
+     * not evaluate any calls to {@code append(...)} and will return an empty
+     * {@link DiffResult} when {@link #build()} is executed.
+     * </p>
+     * @param lhs
+     *            {@code this} object
+     * @param rhs
+     *            the object to diff against
+     * @param style
+     *            the style will use when outputting the objects, {@code null}
+     *            uses the default
+     * @param testTriviallyEqual
+     *            If true, this will test if lhs and rhs are the same or equal.
+     *            All of the append(fieldName, lhs, rhs) methods will abort
+     *            without creating a field {@link Diff} if the trivially equal
+     *            test is enabled and returns true.  The result of this test
+     *            is never changed throughout the life of this {@link ReflectionDiffBuilder}
+     * @throws IllegalArgumentException
+     *             if {@code lhs} or {@code rhs} is {@code null}
+     */
+    public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style, boolean testTriviallyEqual) {
         this.left = lhs;
         this.right = rhs;
-        this.diffBuilder = new DiffBuilder<>(lhs, rhs, style);
+        this.testTriviallyEqual = testTriviallyEqual;
+        this.diffBuilder = new DiffBuilder<>(lhs, rhs, style, testTriviallyEqual);
     }
 
     private boolean accept(final Field field) {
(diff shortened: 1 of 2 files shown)


akalashnyk-callfire (approved):

garydgregory:
@yurbar 
This does not make sense to me; use `@DiffExclude` or `ReflectionDiffBuilder.Builder.setExcludeFieldNames(String...)`.

berse2212:
Just came here to create the same PR. I need to compare JPA Entities with the same ID but different fields. This is because I want to find the changes between different versions of the entities. Sadly the ReflectionDiffBuilder always uses the shortcut of comparig by equals() which is always true in this case. 

@garydgregory you cannot exclude the equals check this why because it's done before even creating the diff.

garydgregory:
I'll need to check back on this PR later this week.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
```diff
@@ -103,9 +104,38 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
...
+     * @param testTriviallyEqual
+     *            If true, this will test if lhs and rhs are the same or equal.
+     *            All of the append(fieldName, lhs, rhs) methods will abort
+     *            without creating a field {@link Diff} if the trivially equal
+     *            test is enabled and returns true.  The result of this test
+     *            is never changed throughout the life of this {@link ReflectionDiffBuilder}
+     * @throws IllegalArgumentException
+     *             if {@code lhs} or {@code rhs} is {@code null}
+     */
+    public ReflectionDiffBuilder(final T lhs, final T rhs, final ToStringStyle style, boolean testTriviallyEqual) {
```
garydgregory, line 134: Do _not_ add another constructor. This class provides a builder for new settings.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -184,4 +209,18 @@ public void testGetExcludeFieldNamesWithNullValuesInExcludedFieldNames() {
...
+        final TypeTestEntityClass firstObject = new TypeTestEntityClass();
+        firstObject.id = 1L;
+        firstObject.field1 = "a";
+
+        final TypeTestEntityClass secondObject = new TypeTestEntityClass();
+        secondObject.id = 1L;
+        secondObject.field1 = "b";
+
+        final DiffResult list = firstObject.diff(secondObject);
+        assertEquals(1, list.getNumberOfDiffs());
```
garydgregory, line 223: This test should assert _what_ the difference is; IOW, make assertions on the other properties of the `DiffResult` and `Diff` instances.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -29,6 +29,31 @@ private static final class TypeTestChildClass extends TypeTestClass {
         String field = "a";
     }
 
+    @SuppressWarnings("unused")
+    private static class TypeTestEntityClass implements Diffable<TypeTestEntityClass> {
```
garydgregory, line 33: Rename `TypeTestEntityClass` to `EntityTestClass`.
Add a class Javadoc that states the intent for the test fixture to emulate a JPA entity class.

src/test/java/org/apache/commons/lang3/builder/ReflectionDiffBuilderTest.java
```diff
@@ -29,6 +29,31 @@ private static final class TypeTestChildClass extends TypeTestClass {
...
+        private String field1;
+
+        @Override
+        public DiffResult<TypeTestEntityClass> diff(TypeTestEntityClass obj) {
+            return new ReflectionDiffBuilder(this, obj, SHORT_STYLE, false).build();
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) return true;
```
garydgregory, line 45: Always use blocks (`{...}`}

garydgregory (changes requested):
Hello @yurbar 

Thank you the PR.

You did not reply to my previous question:

Why not use @DiffExclude or `ReflectionDiffBuilder.Builder.setExcludeFieldNames(String...)`.

Please see my embedded comments.

New elements should have a Javadoc `@since` tag of `3.17.0`.

yurbar:
> Hello @yurbar
> 
> Thank you the PR.
> 
> You did not reply to my previous question:
> 
> Why not use @DiffExclude or `ReflectionDiffBuilder.Builder.setExcludeFieldNames(String...)`.
> 
> Please see my embedded comments.
> 
> New elements should have a Javadoc `@since` tag of `3.17.0`.

@DiffExclude just excludes particular field from output, but I need an ability to compare object which are equal by equals method, but different by other fields.

src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
yurbar: The constructor is the purpose of this change. If I remove it, then the change becomes useless

PR #1148, opened 2023-12-11
Address minor redundancies after code inspection

ParanoidUser:
This time the focus is on unnecessary string concatenation and redundant local assignments.


src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
```diff
@@ -51,7 +51,7 @@ public class RandomStringUtilsTest extends AbstractLangTest {
      */
     private double chiSquare(final int[] expected, final int[] observed) {
         double sumSq = 0.0d;
-        double dev = 0.0d;
+        double dev;
```
elharo, line 54: This one should be declared in the for loop

src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
```diff
@@ -511,7 +511,7 @@ public void testRandomStringUtils() {
     public void testRandomStringUtilsHomog() {
         final String set = "abc";
         final char[] chars = set.toCharArray();
-        String gen = "";
+        String gen;
```
elharo, line 514: Again, if you're doing this move the declaration into the fior loop; don't leave variables undeclared as it's bug prone

src/test/java/org/apache/commons/lang3/RangeTest.java
```diff
@@ -117,20 +117,16 @@ public void testConstructorSignatureWithAbstractComparableClasses() {
         final DerivedComparableA derivedComparableA = new DerivedComparableA();
         final DerivedComparableB derivedComparableB = new DerivedComparableB();
 
-        Range<AbstractComparable> mixed = Range.between(derivedComparableA, derivedComparableB);
-        mixed = Range.between(derivedComparableA, derivedComparableB, null);
+        Range<AbstractComparable> mixed = Range.between(derivedComparableA, derivedComparableB, null);
```
elharo, line 120: This looks OK but I'm not sure what was intended here, but tests code can be weird about things like this. Any chance this was testing that the two initializations in sequence were potentially problematic? I do wish whoever wrote this had left a comment behind explaining their reasoning.

src/test/java/org/apache/commons/lang3/function/MethodFixtures.java
```diff
@@ -124,7 +124,7 @@ public static String staticGetString() {
     @AfterEach
     public void clear() {
```
elharo, line 125: I don't think this method is needed at all. Whoever wrote it probably had a common misunderstanding about how JUNit works. Every test method gets a completely new object. There's no need to reset fields

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
```diff
@@ -543,221 +539,108 @@ public void testIsArrayTypeClasses() {
     @Test
     public void testIsAssignable() throws SecurityException, NoSuchMethodException,
             NoSuchFieldException {
-        List list0 = null;
```
elharo, line 546: I don't understand at all what rhe original author was doing here. This is weird.

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
```diff
@@ -543,221 +539,108 @@ public void testIsArrayTypeClasses() {
...
-        List<? extends String>[] list12;
-        List<? super String>[] list13;
         final Class<?> clazz = getClass();
         final Method method = clazz.getMethod("dummyMethod", List.class, List.class, List.class,
                 List.class, List.class, List.class, List.class, List[].class, List[].class,
                 List[].class, List[].class, List[].class, List[].class, List[].class);
         final Type[] types = method.getGenericParameterTypes();
-//        list0 = list0;
         delegateBooleanAssertion(types, 0, 0, true);
-        list1 = list0;
```
elharo, line 567: OK, I might see what's going on, in which case these assignments are exactly what's being tested. I'm not sure, but if so they should not be removed

garydgregory:
I think we should close this PR unless it gets updated. CC @elharo

ParanoidUser:
Hi @garydgregory / @elharo,

My apologies for being slow on this thread for a while. I keep my eyes open for any feedback from you on this particular cleanup task, as well as others like it to come. Before I rush to change the code on the above points, I need to get a little more clarity. I'll add my doubts and intentions later today.

Thank you

src/test/java/org/apache/commons/lang3/RangeTest.java
ParanoidUser: Here is the initial commit (https://github.com/apache/commons-lang/commit/e6eba6e9acbd646e9a28546b14ae1d663dc6751a) for reference. It feels like the intent was to check generic derivatives passed to the method. Sequential initialization is not a problem, but having an extra assertion for 2-arg methods result might be a better way to test this.

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
ParanoidUser: Personally, I don't see how assigning a bunch of unused variables here and there could affect the generic types declared on the dummy method/fields in the test. But since removing these local variables doesn't cause any assertion failures, it feels like (A) they're redundant, and/or (B) we have a broken test that doesn't verify what was intended. Should we try to run it on different JVMs (Oracle/Eclise/etc.)?

src/test/java/org/apache/commons/lang3/RangeTest.java
ParanoidUser: For the assertion part, we could use [TypeUtils](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java) to check the range type. Let me know if you have a better idea.

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
elharo: I don't really follow the original code here either. To some extent Chesterton's fence is in play. Since I don't understand why somebody built this fence (or code) I'm much more cautious than usual about removing it. :-)

elharo: This code was originally written very poorly. It's possible it never tested what i was meant to test, though I am curious why certain lines were commented out as if the test would break if they were uncommented.

The delegateBooleanAssertion method should have been inlined, and the entire method should have been broken up into many test methods with one assert per test.

Possibly doing that now might help us understand it better. It might also help if we could figure out how changing the code here by removing one of the assignments causes the test to break. But maybe it doesn't?

It does seem true that all these assignments you've removed seem like they should have no effect.

elharo: Aha! I see what these assignments do. They're pseudo-comments, indicating what the next line is testing. They're bad comments, since they're not actually comments and since they're more confusing than not, but they should be safe to remove. 

But wow, is this method bad. The types array is equally confusing. There should be an individual named variable for each type being compared.

If you don't want to go that far, I suggest not deleting these assignments but instead making them comments and maybe expand the comment to something like 

// test the generic type of list 1 is assignable to the generic type of list 2

ParanoidUser: Wow! Thanks for taking the time to crack this puzzle. Using variable assignments as pseudo-comments is a profoundly original way to confuse people 🙂

Here are a few things I did to make this test a little easier to digest:
- Broke it down into 6 test groups. Let me know if you think we should go with smaller groups;
- Inlined `delegateBooleanAssertion` method;
- Removed the unused private `testIsAssignable` method;

src/main/java/org/apache/commons/lang3/concurrent/ConstantInitializer.java
```diff
@@ -133,7 +133,6 @@ public boolean isInitialized() {
      */
     @Override
     public String toString() {
-        return String.format(FMT_TO_STRING, Integer.valueOf(System.identityHashCode(this)),
-                String.valueOf(getObject()));
+        return String.format(FMT_TO_STRING, Integer.valueOf(System.identityHashCode(this)), getObject());
```
elharo, line 136: Is Integer.valueOf needed here? I think an int might work

src/test/java/org/apache/commons/lang3/text/ExtendedMessageFormatTest.java
```diff
@@ -415,9 +415,9 @@ public void testExtendedAndBuiltInFormats() {
         for (final Locale locale : testLocales) {
             final MessageFormat builtins = createMessageFormat(builtinsPattern, locale);
             final String expectedPattern = extendedPattern + builtins.toPattern();
-            DateFormat df = null;
-            NumberFormat nf = null;
-            ExtendedMessageFormat emf = null;
+            DateFormat df;
+            NumberFormat nf;
+            ExtendedMessageFormat emf;
             if (locale == null) {
```
elharo, line 421: locale really shouldn't be null here. If it is, that should be a separate test method.

ParanoidUser: The `testLocales` we're looping through contains the null locale scenario (last value added). So I split this test into the *default locale* scenario and the *available locales* group.

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
```diff
@@ -539,360 +521,243 @@ public void testIsArrayTypeClasses() {
...
+        assertTrue(TypeUtils.isAssignable(Integer[].class, intComparableArrayType));
+    }
+
+    @Test
+    public void testIsAssignableGenericListTypes() throws SecurityException, NoSuchMethodException {
+        final Method method = getClass().getMethod("dummyMethod", List.class, List.class, List.class,
+            List.class, List.class, List.class, List.class, List[].class, List[].class,
+            List[].class, List[].class, List[].class, List[].class, List[].class);
+
+        /*
```
elharo, line 569: It would be even more readable to pull each of these out into an appropriately named local variable and test the variables below.

e.g.

List list = types[0];
List<Object> list_object = types[1]
List<?> list_question_mark = types[2];

etc.

Up to you.

elharo (approved):

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
garydgregory: Except use camel case for variable names, no underscores.

src/main/java/org/apache/commons/lang3/concurrent/ConstantInitializer.java
```diff
@@ -133,6 +133,6 @@ public boolean isInitialized() {
      */
     @Override
     public String toString() {
-        return String.format(FMT_TO_STRING, Integer.valueOf(System.identityHashCode(this)), getObject());
+        return String.format(FMT_TO_STRING, System.identityHashCode(this), getObject());
```
garydgregory, line 136: The boxing happens no matter what so it's not unnecessary. Using boxing and unboxing in a level library like this one makes it obvious where to look for possible NPE and possibly unnecessary conversions. IOW, please don't change this code, it's more clear IMO.

ParanoidUser: I think it was @elharo's suggestion ([quote](https://github.com/apache/commons-lang/pull/1148/commits/1f17f592279d3d9b63ea2fd02be26b62f0656dab#r1449249499)) to remove the explicit integer boxing since it will be automatically boxed either way, so I am not sure which side to take 🙂 .

Since the `System.identityHashCode(this)` method returns a primitive `int` (which is not subject to NPE), I don't see how explicit boxing can help here. Also, by default, IntelliJ IDEA highlights this code with an "unnecessary boxing" warning, which can be annoying and motivate other developers to remove the code. If we decide to leave explicit boxing as is, I'd prefer a comment describing its purpose.

garydgregory: RE boxing we are only going to go in circles because you can also configure an IDE to generate warnings when implicit boxing sites are found (like in Eclipse). Adding comments everywhere this happens is unreasonable IMO.

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
ParanoidUser: I share your sympathy for having clearer type variables. It feels like the `GenericTypeHolder` class might be a better place to store generic types instead of using the `dummyMethod` field.

src/main/java/org/apache/commons/lang3/concurrent/ConstantInitializer.java
ParanoidUser: Agreed. Do you think this rule works for implicit string concatenation as well? Also, should it be project scoped or apply mostly to the code in the `main' folder?
```
String string = "The answer to the universe ";
int number = 42;

return string + number; // implicit
return string + String.valueOf(number); // explicit
```

src/test/java/org/apache/commons/lang3/reflect/TypeUtilsTest.java
ParanoidUser, line 1: @garydgregory, it seems that the `GClass` class and field are never used in the tests. Do you think this is a unique case that needs to be covered, or can this class be safely removed?

```suggestion
    public class GClass<T extends BClass<? extends T> & AInterface<AInterface<? super T>>> {
        // empty
    }
    
    ...
    
    public GClass gClass;
```

src/main/java/org/apache/commons/lang3/concurrent/ConstantInitializer.java
elharo: IDEs can be configured various ways, but what is it out of the box? I can't say I've ever seen a warning for implicit boxing in a case like the one in this PR. 

At the end of the day, implicit boxing is an implementation detail. I prefer the simpler and shorter implicit boxing to making the code longer and harder to read. Implicit boxing is in no way unclear in this method.

There are other improvements that could be make here like inlining FMT_TO_STRING so you don't have to scroll or expand when reading the code. E.g. right now on Github in this PR I can't see what FMT_TO_STRING actually is since Github hides those lines. I have to take extra steps to understand it. 

Even better would be to eliminate the format string completely and go back to simple, obvious string concatenation that puts the parts together in the order they're meant to appear instead of mixing all up them. :-)

ParanoidUser:
Is there anything else we need to discuss/address in this PR?

elharo:
I'm good with this but I'm not a committer so I can't approve the workflow run or merge the pR. @garydgregory

garydgregory:
Building...

PR #1151, opened 2023-12-19
[LANG-1702] test for LANG-1702

elharo:
This passed locally so this bug might not be present any more, or perhaps it depends on JDK versions, or I don't understand it, or something. Let's see what the CI says and find out.


garydgregory:
Hi @elharo 
Run `mvn` locally to catch all build errors.

PR #1159, opened 2024-01-12
Improve handling of malformed `Validate` messages

Marcono1234:
With these changes if the call to `String.format` fails it will generate the message as `message + "\nValues: " + Arrays.toString(values)` and add the `IllegalFormatException` as suppressed exception.
If the message and values are wellformed the behavior remains the same as before.

Rationale:
- Even if the message is malformed, showing it to the user still allows them to troubleshoot the `Validate` failure, which would not be possible when only propagating the `IllegalFormatException`
 This is similar to [Guava's `Preconditions`](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/base/Preconditions.html) which also does not fail on malformed messages
- Since the message is only constructed in the error case, it is likely that unit tests of the user don't cover it since they normally mostly focus on the 'happy path'

However, #953 has in the past already improved the situation for `Validate` a lot, so maybe these changes here are not that important or useful? What do you think?

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Validate.java
+++ b/src/main/java/org/apache/commons/lang3/Validate.java
@@ -23,7 +23,6 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
-import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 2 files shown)


garydgregory:
This feels over the top to me. If we ever wanted to do something like this, it should be done for (1) all call site of `String.format` throughout the code base and (2) refactored in a `StringUtils` method since this is a `String` method. But, I don't think this is needed. This is dealing with a programming error IMO.

Marcono1234:
> it should be done for (1) all call site of `String.format`

For most other calls in Commons Lang the format string seems to be hardcoded (and therefore `IllegalFormatException` is most likely not possible), or at least it is on the 'happy path' where an incorrect format string would be more likely to be caught by users' unit tests, or where propagating the `IllegalFormatException` is the right thing to do because the malformed format string prevents correct functionality of the program.

> This is dealing with a programming error IMO

You are right with this. My concern here is that in most use cases `Validate` checks should pass successfully and no exception should be thrown, but if validation fails then it can be pretty important to see the validation message. However, if then instead an `IllegalFormatException` is thrown, it can make troubleshooting the validation failure pretty difficult. And the `IllegalFormatException` (which extends `IllegalArgumentException`) can also be an unexpected exception type for some of the `Validate` methods which are supposed to throw `IllegalStateException`, `IndexOutOfBoundsException` or `NullPointerException`.

So even though this is definitely a programmer error, I think it is more important to preserve and show the original validation message in some form, than just propagating the `IllegalFormatException`.

Marcono1234:
I am not disagreeing with you that it is a user error. But when validation fails the first priority for a developer will be to investigate why the validation failed. And currently if an `IllegalFormatException` occurs you might not be able to properly investigate the validation failure at all, but instead have to fix the format string, redeploy the application and then hope that you somehow can trigger the validation failure again.
That is why I am proposing in this PR that the `IllegalFormatException` should be handled better, but _not_ be silently discarded.

However, if I cannot convince you with that, then feel free to close this pull request.

garydgregory:
Closing, as summarized by @elharo "If the format string is bad, it's a client error that should result in IllegalFormatException."

PR #1160, opened 2024-01-16
LANG-1720: Fix Javadoc description of exceptions thrown for existing classes

LLM4IG:
During my usage of commons-lang, I have noticed some exceptions (in `StringUtils`, `StrSubstitutor`, `CharSequenceTranslator` and `Conversion`) that are not documented in the Javadoc, which I consider to be unhandled exceptions.

Before creating my this PR, I noticed  #1139, which addresses similar issues. In order to avoid conflicts and show my respect, I have removed some of the Javadoc comments I added in `Conversion.java`, aligning with PR #1139.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/StrSubstitutor.java
+++ b/src/main/java/org/apache/commons/lang3/text/StrSubstitutor.java
@@ -538,6 +538,7 @@ public class StrSubstitutor {
      * @param offset  the start offset within the array, must be valid
      * @param length  the length within the array to be processed, must be valid
      * @return the result of the replace operation
+     * @throws StringIndexOutOfBoundsException if {@code offset < 0 || offset + length >= source.length}
      */
     public String replace(final char[] source, final int offset, final int length) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/lang3/text/StrSubstitutor.java
```diff
@@ -753,6 +757,7 @@ public boolean replaceIn(final StrBuilder source) {
      * @param offset  the start offset within the array, must be valid
      * @param length  the length within the builder to be processed, must be valid
      * @return true if altered
+     * @throws StringIndexOutOfBoundsException if {@code offset < 0 || offset + length >= source.length}
```
elharo, line 760: It does strike me as possible that some of these exceptions shouldn't be thrown at all. If offset or length are out of bounds, maybe nothing is replaced.

elharo (changes requested):

garydgregory:
Hm, where are we on this PR?

PR #1167, opened 2024-01-27
Allow EventListenerSupport to handle (and ignore) exception from listeners allowing invocation of all listeners

garydgregory:
Allow EventListenerSupport to handle (and ignore) exception from listeners allowing invocation of all listeners

This is an alternative to https://github.com/apache/commons-lang/pull/1156


theshoeshiner:
It looks like the only way to add the No-op functionality is to create an anonymous EventListenerSupport class, correct? That's a bit unintuitive but I suppose it requires fewer changes to the existing class. Given that firing quietly was the entire reason for this change maybe it's worth mentioning in the javadoc?

At the end of the day this does address the issue of firing quietly so this should work for me. If I end up needing a more robust implementation then it would probably be best to leave this class as-is and create something new that doesn't involve proxies.

garydgregory:
@theshoeshiner 
The test `testThrowingListenerContinues()` uses an anonymous class, but you can create a subclass in our code of course.

PR #1183, opened 2024-03-11
fix InaccessibleObjectException on java-17

stefansimon76:
"field.setAccessible(true);" throws an exception on java-17 because java.lang is not open for reflection. The problem cannot be worked around on the client side because "setAccessible(true)" is executed before evaluating the passed parameter "excludeFields". Both should be fixed with the change.
The existing tests also cover the change.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
@@ -16,12 +16,16 @@
  */
 package org.apache.commons.lang3.builder;
 
-import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.ObjectUtils;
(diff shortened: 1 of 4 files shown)


garydgregory:
@stefansimon76 
This PR currently has zero tests and no helpful explanation in the PR description.

garydgregory:
Closing: No reply in 2 weeks.

akuhtz:
@garydgregory I think I get the same issue that is described here when running a test that compares 2 ArrayList instances with JDK17. Is this reproducible?

```
@Test
void testEqualsArrayList() {
	List<String> tracks = new ArrayList<>();
	List<String> deserializedTracks = new ArrayList<>();
	
	Assertions.assertTrue(EqualsBuilder.reflectionEquals(tracks, deserializedTracks));
}
```
This is the stacktrace.
```
java.lang.reflect.InaccessibleObjectException: Unable to make field private static final long java.util.ArrayList.serialVersionUID accessible: module java.base does not "opens java.util" to unnamed module @4de5031f
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:178)
	at java.base/java.lang.reflect.AccessibleObject.setAccessible(AccessibleObject.java:130)
	at org.apache.commons.lang3.builder.EqualsBuilder.reflectionAppend(EqualsBuilder.java:1020)
	at org.apache.commons.lang3.builder.EqualsBuilder.reflectionAppend(EqualsBuilder.java:983)
	at org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals(EqualsBuilder.java:236)
	at org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals(EqualsBuilder.java:271)
	at org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals(EqualsBuilder.java:322)
```

garydgregory:
Hello @akuhtz 
If you want to use this feature, you have to give it permission, please see https://github.com/apache/commons-lang/blob/9babe58d895a0e4d6563268ca188794fbf066b38/pom.xml#L468

Axinet:
@garydgregory  based on the documentation of the `EqualsBuilder.reflectionEquals` static fields should *not* be compared, and `java.util.ArrayList.serialVersionUID` is a static field. Isn't it a bug in the EqualsBuilder.reflectionEquals as it doesn't behave in line with its description?

PR #1190, opened 2024-03-25
Add StringUtils#equalsAll and StringUtils#equalsAllIgnoreCase

manzhizhen:
Similar to the StringUtils#equalsAny and StringUtils#equalsAnyIgnoreCase methods, the equalsAll and equalsAllIgnoreCase methods have been added

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1876,6 +1876,67 @@ public class StringUtils {
     }
 
     /**
+     * Compares given CharSequences vararg of {@code searchStrings} to a {@code string},
+     * returning {@code true} if all of CharSequences vararg of {@code searchStrings} is equal to {@code string}.
+     *
+     * <pre>
+     * StringUtils.equalsAll(null, (CharSequence[]) null) = false
+     * StringUtils.equalsAll(null, null, null)    = true
+     * StringUtils.equalsAll(null, "abc", "def")  = false
+     * StringUtils.equalsAll("abc", null, "abc")  = false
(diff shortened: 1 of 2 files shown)


manzhizhen:
> I'm sorry, but I don't see what this would be used for? No point adding code that won't be used.

Are you saying there are no additional unit tests? Let me add

manzhizhen:
> No, I am asking: when would this be useful?
> 
> Is there some real-world situation where this method would be useful?

For example, I have a string constant to indicate that the function is enabled:
`
public static final String OPEN_STRING = "1";
`

Then, under certain conditions, I need to determine whether the different attributes of a certain object are all open. If they are all open, I will perform certain business processing:
`
        if (OPEN_STRING.equals(authority.canEdit()) && OPEN_STRING.equals(authority.canRead()) &&
                OPEN_STRING.equals(authority.isManager()) && OPEN_STRING.equals(authority.isWorking())) {
             // do something
        }
`
If we have StringUtils # equalsAll, we can write as follows:
`
        if (StringUtils.equalsAll(OPEN_STRING, authority.canEdit(), authority.canRead(), authority.isManager(), authority.isWorking())) {
             // do something
        }
`

I don't know if what I said is clear enough

garydgregory:
This whole class needs a rethink IMO, I am not a fan of growing it further.
Also: `Stream.of(array).distinct().count() == 1` or any other more efficient variations using `limit(2)` for example.

manzhizhen:
> The example is clear, but seems quite contrived to me.
> 
> Sorry, but I still don't see this being useful as a general purpose StringUtils method that would earn its keep in the library.

ok...

PR #1198, opened 2024-04-10
Fix inaccessible object exception

stefansimon76:
"field.setAccessible(true);" throws an exception on java-17 because java.lang is not open for reflection. The problem cannot be worked around on the client side because "setAccessible(true)" is executed before evaluating the passed parameter "excludeFields". Both should be fixed with the change.
The existing tests also cover the change.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
@@ -16,12 +16,16 @@
  */
 package org.apache.commons.lang3.builder;
 
-import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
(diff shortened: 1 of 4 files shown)


garydgregory:
Hello @stefansimon76 
- There is too much copy-pasta in this PR, the same code is duplicated at least 3 times.
- The whole point of the feature is to use reflection as the API name makes obvious, so you must enable the app to run with reflection, ignoring reflection exceptions will only hide the fact that the app is misconfigured. So this will likely break existing app behavior.
- The reason we are calling `AccessibleObject.setAccessible()` is to perform a single security check instead of one per field, this PR loses this feature.
- The valid remaining question IMO is: Is there a benefit to pruning the array of fields passed to `AccessibleObject.setAccessible()`? Semantically it seems "yes" but practically, the added complexity might not be worth it base on a quick look at the implementation of `AccessibleObject.setAccessible()`.

garydgregory:
Closing and -1: No reply; this is a reflection feature, if you want to use it, you must enable reflection.

PR #1201, opened 2024-04-16
StringUtils.stripAccents() should handle ligatures, UTF32 math blocks, etc.

speters33w:
Changed normalizer form to NFKD.

See [Unicode Standard Annex 15](https://unicode.org/reports/tr15/)

This fixes the issue for the most part.
See additional tests in StringUtilsTrimStripTest.testStripAccents.

This does not appear to fix LANG-1655

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8119,7 +8119,8 @@ public class StringUtils {
     /**
      * Removes diacritics (~= accents) from a string. The case will not be altered.
      * <p>For instance, '&agrave;' will be replaced by 'a'.</p>
-     * <p>Note that ligatures will be left as is.</p>
+     * <p>Decomposes ligatures and digraphs per the KD column in the
+     * <a href = "https://www.unicode.org/charts/normalization/">Unicode Normalization Chart.</a></p>
      *
      * <pre>
      * StringUtils.stripAccents(null)                = null
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8135,7 +8135,7 @@ public class StringUtils {
      */
     // See also Lucene's ASCIIFoldingFilter (Lucene 2.9) that replaces accented characters by their unaccented equivalent (and uncommitted bug fix: https://issues.apache.org/jira/browse/LUCENE-1343?focusedCommentId=12858907&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_12858907).
     public static String stripAccents(final String input) {
-        if (StringUtils.isEmpty(input)) {
+        if (isEmpty(input)) {
             return input;
         }
         final StringBuilder decomposed = new StringBuilder(Normalizer.normalize(input, Normalizer.Form.NFKD));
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -8135,12 +8135,11 @@ public static String strip(String str, final String stripChars) {
      */
     // See also Lucene's ASCIIFoldingFilter (Lucene 2.9) that replaces accented characters by their unaccented equivalent (and uncommitted bug fix: https://issues.apache.org/jira/browse/LUCENE-1343?focusedCommentId=12858907&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#action_12858907).
     public static String stripAccents(final String input) {
-        if (input == null) {
-            return null;
+        if (StringUtils.isEmpty(input)) {
```
garydgregory, line 8138: You don't need a FQCN here.

garydgregory (changes requested):
@speters33w 
Thank you for your PR.
Please see my one comment.

src/main/java/org/apache/commons/lang3/StringUtils.java
speters33w: You are correct, I pasted from text.CaseUtils so an empty string wouldn't have to be processed, and didn't think about the fact I was pasting it _into_ lang.StringUtils. I pushed the fix.

ecki:
Javadoc says laigatures unhandled

speters33w:
> Javadoc says ligatures unhandled. Also I wonder should it state that it can’t locale specific replacements like ä->ae and ß->ss

Fixed Javadoc. It will handle ligatures and digraphs per the KD column in the [Unicode Normalization charts](https://www.unicode.org/charts/normalization/).
ä will (and should be) be normalized to a. It is not ae in all languages.
Ligatures and digraphs that don't have compatibility decomposition, such as æ or ß will remain unaltered.
It would be possible to modify convertRemainingAccentCharacters() to include characters such as these. It's currently char based so it can't change one char to two, but it could be rewritten.
It would add another iteration to the method and would take me some time to do this. 
Do you think it is necessary?

ecki:
> Do you think it is necessary?

I think it can be a seperate method if somebody requires it. I would just point it out explicitely, because it was the first thing I was excited to get (before I had to code that myself and i did it only for the german conventions).

Having said that it would be useful, but are there any Unicode (database/normalisation) based methods to do that without a hand crafted replacement rule?

speters33w:
> Having said that it would be useful, but are there any Unicode (database/normalization) based methods to do that without a hand crafted replacement rule?

No, there really isn't. The problem is in the Unicode world, if the glyph is considered a letter, it's a letter even if typographically it is a ligature such as in the case of æ or a contraction such as in the case of ß.
For example, Đ (\u0110 D with stroke) is considered an accented character, where Ð (\u00D0, Eth) is considered a letter.
When normalized, Đ (DStroke) will be converted to D, Ð (Eth) will remain unaltered.
It's why Ł and ł aren't translated using normalization, and convertRemainingAccentCharacters() was required.

garydgregory:
@ecki 
How does this PR look good to you for merging?
In general, I'd like to NOT introduce new methods in this domain and instead shift new code to Apache Commons Text which feels like a better home now.

ecki:
I would agree it's a topic for commons-text Buntheit PR doesn't introduce new methods, right?

garydgregory:
> I would agree it's a topic for commons-text, but this PR part doesn't introduce new methods, right?

Correct.

cdouillard:
Hi @garydgregory,

Since the integration of this change (commons-lang3 3.15), the following String "½¼¾" is decomposed differently.

The following code gives a different output between 3.14 and 3.15.

3.14 : StringUtils.stripAccents("½¼¾") -> ½¼¾
3.15 : StringUtils.stripAccents("½¼¾") -> 1⁄21⁄43⁄4

Is this behavior as expected?

garydgregory:
I think so, since the method performs Unicode normalization; the name "Accent" in the method name feels misleading but was kept for compatibility. I updated the Javadoc with more examples in git master. Maybe your reports hints that we should deprecate this method name in favor a new method called "normalize".

PR #1205, opened 2024-04-22
Add StringUtils.notEqual and StringUtils.notEqualIgnoreCase methods

tomohavvk:
**Changes Description**:
Added new methods `notEqual` and `notEqualIgnoreCase` to the `StringUtils` class. These methods are intended for comparing two strings for inequality without using the negation operator.

**Motivation**:
In many cases, there is a need to compare strings for inequality without using the negation operator. Adding these methods makes the code more readable and reduces the likelihood of error.
Using the negation operator (e.g., `!StringUtils.equals(str1, str2)`) can be confusing, especially when comparing many strings. Therefore, introducing these methods simplifies code readability and makes it more understandable.
These new methods add consistency to the library, allowing developers to use a standard approach for comparing strings for inequality.
`ObjectUtils.notEqual` is inspired for the naming.

**Usage Example**:
```java
if (StringUtils.notEqual(str1, str2)) {
    // strings are not equal
}

if (StringUtils.notEqualIgnoreCase(str1, str2)) {
    // strings are not equal (case insensitive)
}
```
**Testing**:
Added corresponding unit tests for the `StringUtils.notEqual` and `StringUtils.notEqualIgnoreCase` methods to ensure they work correctly


garydgregory:
-1 : IMO, we should not add more "not" methods. Java has a "not" operator built in the language. In fact we should talk about deprecating any "not" methods. One disadvantage of using "not" methods is that these do not allow certain tools to work like null state analysis and some refactorings.

PR #1209, opened 2024-05-01
LANG-1732: Converts an Integer to a boolean handling null

imbf:
- Add a function to org.apache.commons.lang3.BooleanUtils class.
- This function converts a Integer object to a boolean value.
- If the Integer object is null, the function returns boolean value it received as an argument.


garydgregory:
Closing, no further comments or discussion.

PR #1215, opened 2024-05-12
LANG-1733: Add null handling features working with Java 8 specs

imbf:
JIRA Ticket: [LANG-1733](https://issues.apache.org/jira/projects/LANG/issues/LANG-1733?filter=allissues)

Suggestion: Introduce null handling features working with with the java.util.function package, or more generally, with Java 8 lambdas. 

- It simplifies null handling by eliminating unnecessary steps such as using specific types or if/else statements.
- It reduces the overhead associated with managing objects like the Optional type.
- It decreases the complexity involved in other null handling methods.


garydgregory (changes requested):
@imbf 
Please follow the pattern established by the 'Suppliers' class in this package. There is no need for a new class. TY.

jochenw:
@garydgregory Have to admit, that even I don't understand, what you are suggesting?

garydgregory:
Our `Suppliers` class implements a null-safe `get` method. If you want to do a null-safe version of `accept` for Consumers, then that new method should go in our `Consumers` class, not a new class.

imbf:
Thanks for your feedback.

I moved null handling features to `Consumers` and `Functions` class in the function package.

src/main/java/org/apache/commons/lang3/function/Consumers.java
```diff
@@ -46,4 +46,17 @@ private Consumers() {
...
 
+    /**
+     * Applies the given {@link Consumer} action to the object if the object is not null. If the object is null, it does
+     * nothing.
+     *
+     * @param object the object to be consumed.
+     * @param consumer the consumer to consume.
+     * @param <T> the type of the argument the consumer accepts.
+     */
+    public static <T> void acceptIfNotNull(final T object, final Consumer<T> consumer) {
```
garydgregory, line 57: - The name can just be `accept`.
- Add a Javadoc since tag.
- Add null-check for consumer, see `Suppliers.get()`

src/main/java/org/apache/commons/lang3/function/Consumers.java
```diff
@@ -21,7 +21,7 @@
 import java.util.function.Function;
 
 /**
- * Provides {@link Consumer} instances.
+ * Provides {@link Consumer} instances and utilities for working with {@link Consumer}.
```
garydgregory, line 24: Unnecessary edit.

src/main/java/org/apache/commons/lang3/function/Functions.java
```diff
@@ -41,4 +41,18 @@ public static <T, R> Function<T, R> function(final Function<T, R> function) {
...
+
+    /**
+     * Applies the {@link Function} on the object if the object is not {@code null}. If the object is {@code null}, it
+     * does nothing.
+     *
+     * @param object the object to apply the function.
+     * @param function the function to apply.
+     * @param <T> the type of the argument the function applies.
+     * @param <R> the type of the result the function returns.
+     * @return the value the function returns If the object is null; null otherwise.
```
garydgregory, line 53: - The name can just be `apply`.
- Add a Javadoc since tag.
- Add null-check for `function`, see `Suppliers.get()`

src/test/java/org/apache/commons/lang3/function/ConsumersTest.java
```diff
@@ -46,4 +50,15 @@ public void testNop() {
...
+    /**
+     * Tests {@link Consumers#acceptIfNotNull(Object, Consumer)}.
+     */
+    @Test
+    public void testAcceptIfNotNull() {
+        StringBuilder builder = new StringBuilder("jay");
+        Consumers.acceptIfNotNull(builder, sb -> sb.append("-bae"));
+        assertEquals("jay-bae", builder.toString());
+
+        assertDoesNotThrow(() -> Consumers.acceptIfNotNull((String) null, string -> fail()));
```
garydgregory, line 62: No need for `assertDoesNotThrow`

src/test/java/org/apache/commons/lang3/function/ConsumersTest.java
```diff
@@ -46,4 +50,15 @@ public void testNop() {
         Consumers.nop().accept("");
     }
 
+    /**
+     * Tests {@link Consumers#acceptIfNotNull(Object, Consumer)}.
+     */
+    @Test
+    public void testAcceptIfNotNull() {
+        StringBuilder builder = new StringBuilder("jay");
+        Consumers.acceptIfNotNull(builder, sb -> sb.append("-bae"));
```
garydgregory, line 59: Don't use your name as test data, just use data like foo and bar like you did in the other test.

garydgregory (changes requested):
@imbf 
Thank you for your update. Please see my comments.

imbf:
Thank you for your feedback.

I applied most of feedbacks. but the name suggestion is not good for me.
I think your suggestion is short and easy but the existing name describes the functionality of the method more accurately.

garydgregory:
Well, you can change the names now, and save me the work, or I'll change them after I review the PR and it is merged...

src/main/java/org/apache/commons/lang3/function/Functions.java
```diff
@@ -41,4 +41,20 @@ public static <T, R> Function<T, R> function(final Function<T, R> function) {
...
+     * @param object the object to apply the function.
+     * @param function the function to apply.
+     * @param <T> the type of the argument the function applies.
+     * @param <R> the type of the result the function returns.
+     * @return the value the function returns if the object and the function are not {@code null}; {@code null}
+     * otherwise.
+     * @since 3.15.0
+     */
+    public static <T, R> R applyIfNotNull(final T object, final Function<T, R> function) {
+        return object != null && function != null ? function.apply(object) : null;
```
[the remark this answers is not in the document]
imbf: That sounds great. null can also be a valid input.

Thank you.

imbf:
@garydgregory 

Initially, I added apis that only work when the object is null. In this case, there were many useful call sites in this project. (e.g. `ArrayFill.fill()`) After applying https://github.com/apache/commons-lang/pull/1215#discussion_r1611600920, these apis accepted null as valid input and lost null-safe functionality.

This is not my intent. If you want to apply a functional interface regardless of whether the object is null or not, making other methods is better than fixing existing methods.

What do you think about my opinion?

src/test/java/org/apache/commons/lang3/function/ConsumersTest.java
```diff
@@ -46,4 +51,23 @@ public void testNop() {
         Consumers.nop().accept("");
     }
 
+    /**
+     * Tests {@link Consumers#accept(Object, Consumer)}.
+     */
+    @Test
+    public void testAccept() {
+        final StringBuilder builder = new StringBuilder("foo");
```
garydgregory, line 59: Don't use mock testing here, it makes the test too obtuse; this is not a hard feature to test IMO.

garydgregory (changes requested):
Hello @imbf 
Thank you for your update.
I added one comment.

src/test/java/org/apache/commons/lang3/function/ConsumersTest.java
imbf: Thank you for your feedback.
I modified my code to not use mock testing.

garydgregory:
@imbf 
The latest changes break the build.
Run 'mvn' by itself to run all build checks before you push.

PR #1216, opened 2024-05-14
StringUtils.stripAccents(String) doesn't handle "\u0111" and "\u0110" (Vietnamese)

hunghhdev:
In Vietnamese, we have the characters "đ" and "Đ" (For reference, here is the Wikipedia page about this character: https://en.wikipedia.org/wiki/D_with_stroke.) When using your function, we need to manually replace these characters to remove their diacritical marks. It would be wonderful if you could add support for removing the diacritical marks from these two characters in your library. Thank you!


garydgregory:
@hunghhdev 
Thank you for your PR! Merged.

PR #1227, opened 2024-05-26
Support generation of single quotes

sebbASF:
This is a potential fix to allow for generation of single quotes.

Whilst this changes the behaviour of the class, does it matter?
Would users ever need to create a format string with an empty constant string?

### DIFF
--- a/src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
@@ -485,7 +485,7 @@ public class DurationFormatUtilsTest extends AbstractLangTest {
         assertEquals("a'", DurationFormatUtils.formatPeriod(time1970, time, "'a'''"));
         assertEquals("'b", DurationFormatUtils.formatPeriod(time1970, time, "'''b'"));
         assertEquals("a'b", DurationFormatUtils.formatPeriod(time1970, time, "'a''''b'"));
-        assertEquals("01' 04sec",DurationFormatUtils.formatDuration(64000L, "mm'' ss'sec'"));
+        assertEquals("01' 04sec", DurationFormatUtils.formatDuration(64000L, "mm'' ss'sec'"));
     }
 
     @Test


src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -264,7 +264,12 @@ static String format(final Token[] tokens, final long years, final long months,
             }
             if (isLiteral) {
                 if (!inOptional || !lastOutputZero) {
-                    buffer.append(value.toString());
+                    final String val = value.toString();
+                    if (val.length() > 0) {
+                        buffer.append(val);
+                    } else {
+                        buffer.append("'");
+                    }
```
garydgregory, line 272: Or, more succinctly: `buffer.append(val.isEmpty() ? "'" : val);`

garydgregory:
Good question, this is where we need to make a judgment call on what kind of input the class takes.

garydgregory:
Hm, in the Javadoc for SDF I see:
```
"EEE, MMM d, ''yy"                         Wed, Jul 4, '01  
```
So they DO support doubling  single quotes to generate one quote. What am I missing?

garydgregory:
Pardon my possible confusion but isn't this PR all about addressing this discrepancy?

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
```diff
@@ -264,7 +264,12 @@ static String format(final Token[] tokens, final long years, final long months,
             }
             if (isLiteral) {
                 if (!inOptional || !lastOutputZero) {
-                    buffer.append(value.toString());
+                    final String val = value.toString();
```
elharo, line 267: val --> s?

elharo (commented):
I would expect this PR would require docs changes or additions somewhere.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
garydgregory: What do you mean by `val --> s?` ? Assign `val` to `s`?

elharo: No, rename val to s since it's just a string. In this case there's no obvious difference between val and value and s is pretty standard in Java for arbitrary string variables that don't have a deeper meaning.

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -473,6 +473,21 @@ public void testFormatPeriod() {
...
     }
 
+    @Test
+    public void testSingleQuotes() {
+        final Calendar cal1970 = Calendar.getInstance();
+        cal1970.set(1970, Calendar.JANUARY, 1, 0, 0, 0);
+        cal1970.set(Calendar.MILLISECOND, 0);
+        final long time1970 = cal1970.getTime().getTime();
+        long time = time1970 + 60 * 1000;
+        // empty quoted strings: '' should now generate a single quote
```
elharo, line 483: OK, now I see why I was confused. Two single quotes is **not** or no longer is an empty quoted string. This is an escape sequence for a single quote. Rewrite comment to clarify.

src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
garydgregory: The name 's' is already in use in this scope so your comment was confusing to me.

elharo: OK, then perhaps valueString or valueAsString or something like that

elharo: On line 249 what's the actual type? Could we just cast that to string?

PR #1234, opened 2024-06-12
LANG-1737: add utility methods - isAnyNotBlank and isAnyNotEmpty

CagesThrottleUs:
These methods are just extra methods created on top of isAllBlank and isAllEmpty.

Found them to be in use with one of my original projects, thought would be helpful to community.

Also adds tests - can be much more extensive if required at a later date

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3564,6 +3564,56 @@ public class StringUtils {
     }
 
     /**
+     * Checks if any of the CharSequences is NOT empty (""), NOT null or NOT whitespace only.
+     *
+     * <p>Whitespace is defined by {@link Character#isWhitespace(char)}.</p>
+     *
+     * <pre>
+     * StringUtils.isAnyNotBlank(null)             = false
+     * StringUtils.isAnyNotBlank(null, "foo")      = true
+     * StringUtils.isAnyNotBlank(null, null)       = false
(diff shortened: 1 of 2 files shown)


garydgregory:
-1: Not worth it IMO; otherwise, we're going to end up with "not" APIs for all methods that return a boolean. The method names become so hard to read in the sense that you must go to the Javadoc to understand what this means. It feels simpler to me to understand what the base operation does and then stick a `!` in front of it.

What do others think?

CagesThrottleUs:
I agree with your points that understanding the base operations and then working on it is what seems to be the best case.

However looking at isNotBlank and other already existing APIs that basically negate an existing APIs in response.

I don't think we need to push it out immediately, but it feels more like on the use case of users, if we see a lot of users repeating the same code which is either the for loop and returning the opposite value from isAll cases OR actually just using negation then maybe we can push this.

Since you guys have been committing for quite some time, maybe using history lessons of why was isNotBlank etc. were added, maybe we can go ahead with this on a later date.

garydgregory:
> I don't think we need to push it out immediately

It won't be merged per the discussion here and https://lists.apache.org/thread/jgvmvb3mcjj4c3hmwy2fnjh8jrzs7kq8 

> However we have to consider if this is code bloat, or if the
addition of simple negation methods is so useful that it warrants
inclusion. I would currently consider this as redundant given the lack of
actual logic in the methods.

You might as well close this PR, or, we'll close it.

PR #1235, opened 2024-06-13
Reimplement RandomStringUtils on top of SecureRandom#getInstanceStrong()

garydgregory:
The previous implementation used  ThreadLocalRandom#current()

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CachedRandomBits.java
+++ b/src/main/java/org/apache/commons/lang3/CachedRandomBits.java
@@ -81,7 +81,7 @@ final class CachedRandomBits {
         int result = 0;
         int generatedBits = 0; // number of generated bits up to now
         while (generatedBits < bits) {
-            if (bitIndex / 8 >= cache.length) {
+            if (bitIndex >> 3 >= cache.length) {
                 // we exhausted the number of bits in the cache
                 // this should only happen if the bitIndex is exactly matching the cache length
                 assert bitIndex == cache.length * 8;
@@ -90,9 +90,9 @@ final class CachedRandomBits {
             }
             // generatedBitsInIteration is the number of bits that we will generate
             // in this iteration of the while loop
-            int generatedBitsInIteration = Math.min(8 - (bitIndex % 8), bits - generatedBits);
+            int generatedBitsInIteration = Math.min(8 - (bitIndex & 0x7), bits - generatedBits);
             result = result << generatedBitsInIteration;
-            result |= (cache[bitIndex / 8] >> (bitIndex % 8)) & ((1 << generatedBitsInIteration) - 1);
+            result |= (cache[bitIndex >> 3] >> (bitIndex & 0x7)) & ((1 << generatedBitsInIteration) - 1);
             generatedBits += generatedBitsInIteration;
             bitIndex += generatedBitsInIteration;
         }
--- a/src/main/java/org/apache/commons/lang3/CachedRandomBits.java
+++ b/src/main/java/org/apache/commons/lang3/CachedRandomBits.java
@@ -16,16 +16,15 @@
  */
 package org.apache.commons.lang3;
 
+import java.util.Objects;
 import java.util.Random;
 
 /**
  * Generates random integers of specific bit length.
  *
  * <p>
- * It is more efficient than calling Random.nextInt(1 << nbBits).
- * It uses a cache of cacheSize random bytes that it replenishes when it gets empty.
- * This is especially beneficial for SecureRandom Drbg implementations,
- * which incur a constant cost at each randomness generation.
+ * It is more efficient than calling Random.nextInt(1 << nbBits). It uses a cache of cacheSize random bytes that it replenishes when it gets empty. This is
+ * especially beneficial for SecureRandom Drbg implementations, which incur a constant cost at each randomness generation.
  * </p>
  *
  * <p>
@@ -36,7 +35,8 @@ import java.util.Random;
  * #NotThreadSafe#
  * </p>
  */
-class AmortizedRandomBits {
+final class CachedRandomBits {
+
     private final Random random;
 
     private final byte[] cache;
@@ -53,17 +53,17 @@ class AmortizedRandomBits {
     private int bitIndex;
 
     /**
-     * Creates an AmortizedRandomBits instance.
+     * Creates a new instance.
      *
      * @param cacheSize number of bytes cached (only affects performance)
      * @param random random source
      */
-    AmortizedRandomBits(final int cacheSize, final Random random) {
+    CachedRandomBits(final int cacheSize, final Random random) {
         if (cacheSize <= 0) {
             throw new IllegalArgumentException("cacheSize must be positive");
         }
         this.cache = new byte[cacheSize];
-        this.random = random;
+        this.random = Objects.requireNonNull(random, "random");
         this.random.nextBytes(this.cache);
         this.bitIndex = 0;
     }
@@ -78,10 +78,8 @@ class AmortizedRandomBits {
         if (bits > 32 || bits <= 0) {
             throw new IllegalArgumentException("number of bits must be between 1 and 32");
         }
-
         int result = 0;
         int generatedBits = 0; // number of generated bits up to now
-
         while (generatedBits < bits) {
             if (bitIndex / 8 >= cache.length) {
                 // we exhausted the number of bits in the cache
@@ -90,18 +88,14 @@ class AmortizedRandomBits {
                 random.nextBytes(cache);
                 bitIndex = 0;
             }
-
             // generatedBitsInIteration is the number of bits that we will generate
             // in this iteration of the while loop
             int generatedBitsInIteration = Math.min(8 - (bitIndex % 8), bits - generatedBits);
-
             result = result << generatedBitsInIteration;
             result |= (cache[bitIndex / 8] >> (bitIndex % 8)) & ((1 << generatedBitsInIteration) - 1);
-
             generatedBits += generatedBitsInIteration;
             bitIndex += generatedBitsInIteration;
         }
-
         return result;
     }
 }
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -16,44 +16,56 @@
  */
 package org.apache.commons.lang3;
 
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 
+import org.apache.commons.lang3.exception.UncheckedException;
+
 /**
  * Generates random {@link String}s.
+ * <p>
+ * Starting in version 3.15.0, this classes uses {@link SecureRandom#getInstanceStrong()}.
+ * </p>
+ * <p>
+ * Before version 3.15.0, this classes used {@link ThreadLocalRandom#current()}, which was NOT cryptographically secure.
+ * </p>
+ * <p>
+ * RandomStringUtils is intended for simple use cases. For more advanced use cases consider using Apache Commons Text's
+ * <a href="https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/RandomStringGenerator.html"> RandomStringGenerator</a>
+ * instead.
+ * </p>
+ * <p>
+ * The Apache Commons project provides <a href="https://commons.apache.org/proper/commons-rng/">Commons RNG</a> dedicated to pseudo-random number generation,
+ * that may be a better choice for applications with more stringent requirements (performance and/or correctness).
+ * </p>
+ * <p>
+ * Note that <em>private high surrogate</em> characters are ignored. These are Unicode characters that fall between the values 56192 (db80) and 56319 (dbff) as
+ * we don't know how to handle them. High and low surrogates are correctly dealt with - that is if a high surrogate is randomly chosen, 55296 (d800) to 56191
+ * (db7f) then it is followed by a low surrogate. If a low surrogate is chosen, 56320 (dc00) to 57343 (dfff) then it is placed after a randomly chosen high
+ * surrogate.
+ * </p>
+ * <p>
+ * #ThreadSafe#
+ * </p>
  *
- * <p><b>Caveat: Instances of {@link Random}, upon which the implementation of this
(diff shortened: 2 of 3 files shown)


psteitz:
SecureRandoms take a lot longer to initialize, which will slow down the loading of this class, which may impact performance for users who don't care about cryptographic security of the generated text.  Where does the crypto security requirement come from?  We already direct users elsewhere for more sophisticated use cases.  In [math] now I guess in rng, we used to provide separate impls so users who just wanted random text did not have to pay for the overhead of SecureRandom.

hyandell:
This came from a conversation Gary and I were having on Slack (which I need to take to dev@). 

Researching with a colleague in my dayjob, the number of use cases of RandomStringUtils that are ignoring the documentation of the class appears much higher than the number of use cases that are my original 2000s intention of do testing. Really feels like we need to take the road less desirable here.

psteitz:
I would expect the number of users who want / need cryptographic security to be tiny.  So why make that the default behavior?  I have not benchmarked recent JDKs, but it used to be a *lot* slower.  I remember tomcat ripping out some SecureRandom uses for performance reasons.  And it also kind of gives me the willies that the instance is initialized (the most expensive part) in a static initializer.

garydgregory:
Hi @psteitz 

Note that is it only initialized on demand.

CC: @hyandell

psteitz:
How is that, @garydgregory?  Doesn't getInstanceStrong initialize the instance?  I thought it blocked waiting for /dev/random.  Blocking in class initialization could be problematic.  Or am I misreading the code or missing something?

aherbert:
I believe @psteitz is correct that the initialisation of SecureRandom can be slow and potentially blocking. The way it is done in Commons RNG and the JDK source code for SplittableRandom (when it is configured for secure seeding via system properties) is to use the default constructor:
```java
SecureRandom rng = new SecureRandom();
```
See [The Right Way to Use SecureRandom](https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/).

A less intrusive modification to this PR would be a similar mechanism to the JDK seeding RNG routine where the source of seeding randomness is chosen based on system properties. But you still have to decide whether the default is secure or non-secure.

Note: A StringSampler implementation was discussed under RNG 54. There is a table showing some relative performance of Lang and Text here: [RNG-54 comment 2](https://issues.apache.org/jira/browse/RNG-54?focusedCommentId=16582795&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16582795). Currently String sampling is not implemented in RNG although the ticket remains open.

For fast random strings it is optimal to use a power of 2 alphabet. But this limits use cases to just outputting random junk text. I presume the case for SecureRandom is for password/salt generation. In this context a power of 2 alphabet is restrictive and a method using a uniform [0, n) choice for each character of alphabet size n is requied. There are samplers in RNG that will outperform use of `nextInt(n)` by precomputing the modulus `Integer.MAX_VALUE % n` which is used to avoid statistical bias from a random integer. The performance gain is negligible on a slow base RNG such as SecureRandom. From memory, we did testing in RNG using the JDK 9 SecureRandom (which added more cryptographic methods) and it was 2 orders of magnitude slower than a typical pseudo-RNG. This was done in a GSoC project and was not merged to the master branch. Repeating benchmarks using some of the framework in the RNG JMH project would be trivial.

garydgregory:
> How is that, @garydgregory? Doesn't getInstanceStrong initialize the instance? I thought it blocked waiting for /dev/random. Blocking in class initialization could be problematic. Or am I misreading the code or missing something?

Hello @psteitz 
The SecureRandom is only initialized if the thread local's initial value supplier lamba is called which only happens when the thread local's get() method is called which only happens when one of the random() methods in this class is called.

aherbert:
Using ThreadLocal still has a potential blocking call to when a random method is invoked.

garydgregory:
> Using ThreadLocal still has a potential blocking call when a random method is invoked.

Hi @aherbert,
Where do you see ThreadLocal documented as blocking on a get() call?

aherbert:
Sorry, clarification: It is the _first call_ to ThreadLocal.get() that will block on the instantiation of the SecureRandom. After that each thread has a SecureRandom and will be fine.

fabrice102:
I tried to research the risk of blocking when instantiating SecureRandom.

TL;DR: On modern OSes and common platforms, blocking is unlikely, unless the program is called very early in the boot process or (maybe but unlikely) just after boot finishes. Even then, blocking would only happen the first time a method from  RandomStringUtils  is called. Other calls to methods of RandomStringUtils in the same thread or other threads would not block. In all cases, if RandomStringUtils is not used, this PR has no performance impact on any system.

On Unix, the [default SUN cryptographic providers](https://github.com/openjdk/jdk/blob/8464ce6db5cbd5d50ac2a2bcba905b7255f510f5/src/java.base/unix/classes/sun/security/provider/NativePRNG.java) SUN/NativePRNG (obtained when using new SecureRandom() to instantiate SecureRandom) and SUN/NativePRNGBlocking (obtained when using SecureRandom.getInstanceStrong()) appear to be using /dev/urandom and /dev/random respectively.

On [modern Linux kernels](https://lwn.net/Articles/808575/) ([5.6](https://lwn.net/Articles/811230/) or newer, 5.6 has been released in 2020), the only time /dev/random would block is if the system never got enough entropy to initialize the randomness pool. In particular, only the first thread making calls to RandomStringUtils may block. Furthermore, on most commonly used systems, the [CPU Jitter source entropy](https://lwn.net/Articles/884875/) would initialize /dev/random less than 1s after boot. So on commonly-used modern Linux systems, there should be virtually no performance impact at initialization of SecureRandom.

Since at least 2016 (most likely even before that), FreeBSD and macOS [seem to](https://github.com/briansmith/ring/issues/316#issuecomment-257479426) have identical /dev/urandom and /dev/random . In addition, /dev/random blocks only on initialization like on modern linux kernels. But I could not find a first-party source confirming this.

At least since [OpenBSD 6.3](https://man.openbsd.org/OpenBSD-6.3/random.4) (released in 2018), /dev/urandom and /dev/random are the same and never block.

I did not check what happens on Windows in details. I could not find significant complaints about Windows SecureRandom blocking. It is possible it never blocks.

psteitz:
We used to have a mailing list where we discussed changes like this.  Sorry for the snarky comment, but this is a good discussion that will be hard to follow for anyone researching what led to this change.  With no threading, we can't really separate the issues either, which are 0) init performance cost and potential to block waiting for entropy (great analysis @fabrice102 , could be no longer an issue for most users) 1) gen time performance cost 2) whether it might be better to create SecureRandomStringUtils or otherwise support fast generation without cyrptographic security.

fabrice102:
Thanks @psteitz ! I am not sure where is the best place to post the following comment. If you think it’s more appropriate to post it to the mailing list, let me know and I will post it there.

To address the point "1) gen time performance cost", I’ve written some optimizations on top of the current PR: https://github.com/garydgregory/commons-lang/pull/2

Benchmarking using JMH with JDK 8 on an AWS c6a.xlarge instance with cryptographic providers SUN/NativePRNG, SUN/SHA1PRNG, [ACCP](https://github.com/corretto/amazon-corretto-crypto-provider) 1.6, and [ACCP](https://github.com/corretto/amazon-corretto-crypto-provider) 2.3, we have the following results. Compared to the current state (with the default java.util.Random/ThreadLocalRandom number generator), when any of the above SecureRandom is used with the optimized code, the time to generate random alphabetic, alphanumeric, and numeric strings of at least 16 characters is less than three times slower (and sometimes even 5x faster). For 4-character strings, we observe a slow-down of at most 8x.

If the optimized RandomStringUtils is called concurrently in multiple threads, the SUN/NativePrng SecureRandom implementation (the default on Linux) may be slower as calls are serialized (due to access to/dev/random. This seems an unlikely scenario in practice. In addition, switching to ACCP 2 for SecureRandom would fix the performance issue in that case.

psteitz:
On Fri, Jun 14, 2024 at 4:13 PM Fabrice Benhamouda ***@***.***>
wrote:

> Thanks @psteitz <https://github.com/psteitz> ! I am not sure where is the
> best place to post the following comment. If you think it’s more
> appropriate to post it to the mailing list, let me know and I will post it
> there.
>
> To address the point "1) gen time performance cost", I’ve written some
> optimizations on top of the current PR: garydgregory#2
> <https://github.com/garydgregory/commons-lang/pull/2>
>
> Benchmarking using JMH with JDK 8 on an AWS c6a.xlarge instance with
> cryptographic providers SUN/NativePRNG, SUN/SHA1PRNG, ACCP
> <https://github.com/corretto/amazon-corretto-crypto-provider> 1.6, and
> ACCP <https://github.com/corretto/amazon-corretto-crypto-provider> 2.3,
> we have the following results. Compared to the current state (with the
> default java.util.Random/ThreadLocalRandom number generator), when any of
> the above SecureRandom is used with the optimized code, the time to
> generate random alphabetic, alphanumeric, and numeric strings of at least
> 16 characters is less than three times slower (and sometimes even 5x
> faster). For 4-character strings, we observe a slow-down of at most 8x.
>
> If the optimized RandomStringUtils is called concurrently in multiple
> threads, the SUN/NativePrng SecureRandom implementation (the default on
> Linux) may be slower as calls are serialized (due to access to/dev/random.
>
This is the use case I would be most concerned with.  Many web applications
use this class in multi-threaded app server settings.  I would not use the
class in web apps after this change for that reason.  If we do make the
change, we should make it clear in the class javadoc that it is using
SecureRandom and prone to occasional stalls.

Since there was no discussion about why this change is being made, I will
ask again why it is being proposed.  The only rationale that I can see is
that somehow some users are using the class to generate things that should
be cryptographically secure, like passwords or hash salt, so to protect
these users, we make all other users take a potential performance hit.
Much better, IMO, would be to make the recommended improvements in
performance to the vanilla Random-based impl and add a new class or somehow
shoehorn in a secure impl for users that want / need that.

Phil

This seems an unlikely scenario in practice. In addition, switching to ACCP
2 for SecureRandom would fix the performance issue in that case.

garydgregory:
Hi Phil,

Please see the Commons security list. Are you not subscribed? You should be
since you are on the PMC. I did not want to copy paste the info Hen Y
brought up with me separately just in case.

Gary

fabrice102:
> > If the optimized RandomStringUtils is called concurrently in multiple threads, the SUN/NativePrng SecureRandom implementation (the default on Linux) may be slower as calls are serialized (due to access to/dev/random.
>
> This is the use case I would be most concerned with. Many web applications use this class in multi-threaded app server settings.

To clarify, the case where there may be a slow-down is the following: 

Multiple threads concurrently, at the same time, call `RandomStringUtils.random`.

Generation of a random 16-alphanumerical-character string takes less than 1us on the test machine (c6a.xlarge). (And with a non-optimized implementation using the default `SUN.NativePRNG` provider, this takes less than 3.5us.) For the vast majority of web applications, I believe this should amount to a negligible percentage of the total CPU time used to answer the query. This in turn would mean that there should be no noticeable slow-down due to concurrency in this context: it would be very rare that `RandomStringUtils.random` is called concurrently at the same time by two threads answering two different web queries.

garydgregory:
@nvsnvikram 
You can use the lower level API that takes a Random argument that pass it an implementation that's as weak/insecure as you wish 😉

nvsnvikram:
thanks. we will have to change our code to use that but should be fine.

garydgregory:
I'll experiment with a more friendly API on my end...

knoobie:
Just fyi: I've seen our builds taking 4-16 hours after this change instead of typical 30min in a Jenkins dedicated Docker env where we heavily use RandomStringUtils to generate arbitrary 100-200 character strings in tests.

garydgregory:
I will push out a new release candidate today hopefully where you can call .insecure() to get the old faster and insecure implementation. Right now you can test with 3.16.0-SNAPSHOT.

HelloDhero:
SecureRandom.getInstanceStrong() in RandomUtils while use NativePRNGBlocking default 
it cause blocked

HelloDhero:
SecureRandom.getInstanceStrong() in RandomUtils while use NativePRNGBlocking default 
it cause blocked

HelloDhero:
WechatIMG81

HelloDhero:
@garydgregory

garydgregory:
This is currently being discussed on an internal (private) mailing list.

Gary

fabrice102:
@knoobie and @HelloDHero, you reported important slow-down following the changes in this PR. What OS/kernel are your systems using? In particular, do they use pre-5.6 Linux kernel?

If this is the case, the issue is most likely that `/dev/random` (used by the default `SecureRandom.getInstanceStrong()`) can block at any time on pre-5.6 Linux kernel (see https://github.com/apache/commons-lang/pull/1235#issuecomment-2168423879). 

In that case, switching to `new SecureRandom()` instead of `SecureRandom.getInstanceStrong()` would most likely solve the issue. Doing so may slightly decrease security in Linux, if the code is executed just after boot and before enough entropy is gathered. But this is a very unlikely case and I personally think it is a good trade-off.

A temporary workaround may be to add `SUN/NativePRNG` to the list of strong randomness sources, by changing the [Security property securerandom.strongAlgorithms.](https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html#getInstanceStrong--).This is assuming none of your code relies on the fact that `SecureRandom.getInstanceStrong()` uses `/dev/random`.

HelloDhero:
Modifying the securerandom.strongAlgorithms configuration in a production environment is not a good practice, and we should avoid this situation at the code level

knoobie:
@fabrice102 CentOS7 / Kernel 3.10.0-1160.114.2.el7.x86_64 - so your comment about "pre-5.6" Linux Kernel might be correct. Even tho they are probably still widely spread with RHEL 7 and 8 also using pre-5.6 Kernels.

garydgregory:
FYI: You can experiment now with git master and [3.17.0-SNAPSHOT builds](https://repository.apache.org/content/repositories/snapshots/) to get:

Random[String]Utils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()

- RandomStringUtils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()
- RandomUtils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()
- Add RandomStringUtils.secureStrong()
- Add RandomUtils.secureStrong()

alex91ar:
Well if anyone s feeling generous: bc1qrjv23a92uyqrptlkc9mc77w83344k4v37pem7f

PR #1239, opened 2024-06-15
Replace the specified string with the one enclosed between two other strings(‘open’ and a ‘close’).

Yash-777:
Method Signature and Preconditions:

 * The method replaceSubstringInBetween takes four parameters: str, replace, open, and close. All parameters are marked as final and are expected to be non-null (@param comments indicate this precondition).
 * If any of the preconditions are violated (i.e., if str, open, close, or replace is null), the method simply returns the original str without any modifications.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -7082,15 +7082,19 @@ public class StringUtils {
 	 * <p>
 	 * Examples:
 	 * <pre>
-	 * replaceSubstringInBetween("", "abc", "", "")         = "abc"
+	 * replaceSubstringInBetween(null, "-", "a", null)      = null (Returns original string)
+	 * replaceSubstringInBetween("a", "-", "a", "a")        = a (Returns original string as its length is less than 2)
+	 * replaceSubstringInBetween("abc", "-", "a", "c")      = a-c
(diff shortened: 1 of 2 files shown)


Yash-777:
recreated the Pull request of https://github.com/apache/commons-lang/pull/395

garydgregory:
-1: This is such a niche use case that i don't think it belongs here. It is not even shown used within Commons Lang. The Javadoc shows no real word application as well IMO. This kind of text processing would not belong in Commons Lang anyway. We need to be careful and not grow this class into even more of a kitchen sink.

What do others think?

555vedant:
@garydgregory sir  i would like to work on it what is update of this PR ,
and  where i can find issue JIRA or issue template??

garydgregory:
> @garydgregory sir i would like to work on it what is update of this PR , and where i can find issue JIRA or issue template??

Hello @555vedant, please see https://github.com/apache/commons-lang/pull/1239#issuecomment-2169714855

sigee:
@Yash-777, @555vedant 
-1
In my opinion, replaceSubstringInBetween() is not needed. If you really need such replacements, RegExUtils.replacePattern() can do the same with proper parameters. I think.

garydgregory:
Closing, no more dialog. Also, regular expressions built into Java is another option.

PR #1249, opened 2024-07-20
Add Javadoc links from StopWatch to DurationUtils

obfischer:
Added a JavaDoc link from StopWatch to DurationUtils, as DurationUtils provide similar functiontionality, which could be used in simple use cases.


garydgregory:
Merged. TY @obfischer 👍

PR #1250, opened 2024-07-20
Feature/randomutils secure() and insecure()

garydgregory:
Refactor `RandomUtils` and `RandomStringUtils` with `secure()` and `insecure()`
instances.

For example:
```java
RandomStringUtils.secure().randomInt();
RandomStringUtils.insecure().randomInt();
```

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -16,201 +16,212 @@
  */
 package org.apache.commons.lang3;
 
-import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
+import java.security.Security;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
-
-import org.apache.commons.lang3.exception.UncheckedException;
+import java.util.function.Supplier;
 
 /**
  * Generates random {@link String}s.
  * <p>
- * Starting in version 3.15.0, this classes uses {@link SecureRandom#getInstanceStrong()}.
+ * Starting in version 3.16.0, this class uses {@link #secure()} for static methods and adds {@link #insecure()}.
  * </p>
  * <p>
- * Before version 3.15.0, this classes used {@link ThreadLocalRandom#current()}, which was NOT cryptographically secure.
+ * Starting in version 3.15.0, this class uses {@link SecureRandom#getInstanceStrong()} for static methods.
+ * </p>
+ * <p>
+ * Before version 3.15.0, this class used {@link ThreadLocalRandom#current()} for static methods, which is not
+ * cryptographically secure.
+ * </p>
+ * <p>
+ * Use {@link #secure()} to get the singleton instance based on {@link SecureRandom#getInstanceStrong()} which uses an
+ * algorithms/providers specified in the {@code securerandom.strongAlgorithms} {@link Security} property.
+ * </p>
+ * <p>
+ * Use {@link #insecure()} to get the singleton instance based on {@link ThreadLocalRandom#current()}; <b>which is not
+ * cryptographically secure</b>.
  * </p>
  * <p>
  * RandomStringUtils is intended for simple use cases. For more advanced use cases consider using Apache Commons Text's
- * <a href="https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/RandomStringGenerator.html"> RandomStringGenerator</a>
- * instead.
+ * <a href=
+ * "https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/RandomStringGenerator.html">
+ * RandomStringGenerator</a> instead.
  * </p>
  * <p>
- * The Apache Commons project provides <a href="https://commons.apache.org/proper/commons-rng/">Commons RNG</a> dedicated to pseudo-random number generation,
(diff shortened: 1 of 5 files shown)


nvsnvikram:
Thanks for adding these functions @garydgregory. For us, we would be ok with or without this change. We were not aware of this secure random change and our code started having high execution times until we identified the commons-lang3 upgrade. We will be ok to call the RandomStringUtils.random with our own random instance.

garydgregory:
@nvsnvikram 
Thank you for your POV  😄

danielthegray:
FYI @garydgregory , this API change can cause issues in production code where the process will block waiting for more entropy. This is a huge deal!!

I understand that there is a vulnerability where a password reset URL token was generated using RandomStringUtils and therefore the insecure random number generator. The solution can NOT be to just change the behavior for ALL random numbers to use the secure one! This is a very foundational library, used by many many applications for many many purposes, and just changing it its behavior so deeply (and in a way that **can cause hangs in production code**) is **UNACCEPTABLE**.

So now, to fix the breaks and hangs that people have been complaining about, your solution is to persist in your API-behavior-breaking state and just add a new `.insecure()` that preserves the old behavior? This is not a workable solution, and I'll show you how: imagine that your application depends on Apache Commons Lang, as well as your dependency (which is generating the random strings). You can change your code to use `.insecure()` and then upgrade, but the library is still calling the old API (without `.insecure()`) and your upgrade will preserve the break.

In short, I am asking the Apache Commons team to please revert the behavior to the previous non-hanging state. The solution to people choosing the wrong random number generator for things is not to force secure random number generators on everyone for every use-case!

garydgregory:
Hello @danielthegray 
The original issue was reported on our internal security mailing list and the changes were discussed there. In today's environment of increasing attacks and security breaches up and down software stacks, this felt like the right decision.
With the addition of 'secure()' and 'insecure()' we are offering developers the ability to make a clear and conciuous choice. The static APIs remain to maintain binary compatibility. I think deprecating the static methods is a good idea, which I'll get to next.
As a user of a library that depends on Commons Lang, I would ask that community to make that choice: Use 'secure()', 'insecure()', or make it configurable.

danielthegray:
I can see why it would feel urgent, but I feel that breaking production environments should be a big nono. It's very easy to miss a subtle case like this in testing and have it directly affect production (which are are seeing in repeated instances). It's important to not only preserve binary compatibility, but also behavioral compability (i.e. not hanging indefinitely).

However, it's not in my hands. I disagree but respect the decision that you have made as library maintainers. Have a great rest of your day!

marschall:
I wonder whether a system property could be introduced to revert back to the old behaviour and make `#insecure()` the default. This would help people for which the blocking introduced by the new default is an issue, depend on third party code they can not change and don't need `RandomStringUtils` to be cryptographically secure.

garydgregory:
This change was based on a security based report/request so it would be confusing to change it back. We could consider adding system property to change the default to insecure. Note that in git master all the static methods are now deprecated, signaling to developers that they should pick secure() or insecure().

aherbert:
Would this be mitigated be changing `SecureRandom.getInstanceStrong()` to `new SecureRandom()`? IIUC the later is still cryptographically strong but would not block via using `/dev/urandom`. This could be done using a system property. Thus you can switch the type of secure().

garydgregory:
Could someone run this experiment? We could add a secureStrong() option and change secure() to use this suggestion.

marschall:
> Would this be mitigated be changing `SecureRandom.getInstanceStrong()` to `new SecureRandom()`? IIUC the later is still cryptographically strong but would not block via using `/dev/urandom`. This could be done using a system property. Thus you can switch the type of secure().

No, this would be to get the behaviour of 3.14 back when using the static methods. The behavior specified by the API contract and therefore what a user would expect https://javadoc.io/doc/org.apache.commons/commons-lang3/3.14.0/org/apache/commons/lang3/RandomStringUtils.html

> Caveat: Instances of [Random](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html), upon which the implementation of this class relies, are not cryptographically secure.

marschall:
> Note that in git master all the static methods are now deprecated, signaling to developers that they should pick secure() or insecure().

As outlined earlier, this doesn't help if the usage of `RandomStringUtils` is in a third party dependency and therefore the code can not be changed.

garydgregory:
FYI: You can experiment now with git master and [3.17.0-SNAPSHOT builds](https://repository.apache.org/content/repositories/snapshots/) to get:

Random[String]Utils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()

- RandomStringUtils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()
- RandomUtils.secure() now uses SecureRandom() instead of SecureRandom.getInstanceStrong()
- Add RandomStringUtils.secureStrong()
- Add RandomUtils.secureStrong()

uyw4687:
Anyways it's being deprecated so we should use UniformRandomProvider, ThreadLocalRandom, or SecureRandom directly right?

garydgregory:
Hello @uyw4687 

The constructor and some methods are indeed deprecated, not the whole class.

The Commons RNG component provides a different, richer, and broader set of features; use what's best for your use case.

The real Javadoc is here:
https://commons.apache.org/proper/commons-lang/apidocs/index.html
and an archive is here:
https://javadoc.io/doc/org.apache.commons/commons-lang3/latest/index.html

PR #1254, opened 2024-07-26
LANG-1747: Add support for actions to simplify the measurement of times

obfischer:
TBW

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/StopWatch.java
+++ b/src/main/java/org/apache/commons/lang3/time/StopWatch.java
@@ -21,10 +21,16 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.function.FailableConsumer;
 import org.apache.commons.lang3.function.FailableRunnable;
+import org.apache.commons.lang3.function.FailableSupplier;
+import org.apache.commons.lang3.function.TriFunction;
 
 /**
  * {@link StopWatch} provides a convenient API for timings.
@@ -670,5 +676,69 @@ public class StopWatch {
         }
         splitState = SplitState.UNSPLIT;
     }
+
+    <T, E extends Throwable> T get(FailableSupplier<T, E> supplier) throws E {
+        if (isStopped()) {
+            start();
+        } else if (isSuspended()) {
+            resume();
+        }
+
+        try {
+            return supplier.get();
+        } finally {
+            suspend();
+        }
+    }
+
+
+    /**
+     * Take the time of the execution of a given {@linkplain Function}.
+     *
+     * <p>
+     * <b>Take the time of given {@linkplain Function}</b>
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     *
+     * String result = watch.apply(it -> it.toLowerCase(Locale.ROOT)).apply("A");
+     * }</pre>
+
+     * <b>Take the time of an applied {@linkplain Function} in a stream</b>
+     *
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     *
+     * String result = Stream.of("A", "B", "C")
+     *                       .map(watch.apply(it -> it.toLowerCase(Locale.ROOT)))
+     *                       .collect(Collectors.joining());
+     * }</pre>
+     *
+     * @param function the function those application should be measured
+     *
+     * @return the given function prepared to take time if applied
+     *
+     * @throws IllegalStateException if the StopWatch is not stopped or suspended
+     *
+     * @param <T> the type of the input to the function
+     * @param <R> the type of the result of the function
+     *
+     * @since 3.16
+     */
+    public <T, R> Function<T, R> apply(Function<T, R> function) {
+        return argument -> {
+            if (isStopped()) {
+                start();
+            } else if (isSuspended()) {
+                resume();
+            }
+
+            try {
+                return function.apply(argument);
+            } finally {
+                suspend();
+            }
+        };
+    }
+
 
 }
--- a/src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
+++ b/src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
@@ -32,7 +32,10 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Locale;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.AbstractLangTest;
 import org.apache.commons.lang3.ThreadUtils;
@@ -447,5 +450,30 @@ public class StopWatchTest extends AbstractLangTest {
         watch.split();
         final String splitStr = watch.toString();
         assertEquals(12 + MESSAGE.length() + 1, splitStr.length(), "Formatted split string not the correct length");
+    }
+
+    @Test
+    public void testGetForFailableSupplier() throws InterruptedException {
+        final StopWatch watch = StopWatch.create();
+
+        String result = watch.get(() -> {
+            StopWatchTest.this.sleep(MILLIS_200);
+            return "Foobar";
+        });
+
+        assertEquals("Foobar", result, "Watch returned result other then expected");
+        assertTrue(watch.isSuspended(), "Watch should be suspended");
+    }
+
+    @Test
+    public void testApplyForFunction() {
+        final StopWatch watch = StopWatch.create();
+
+        String result = Stream.of("A", "B", "C")
+                .map(watch.apply(s -> s.toLowerCase(Locale.ROOT)))
+                .collect(Collectors.joining());
+
+        assertEquals("abc", result, "Returned result other then expected");
+        assertTrue(watch.isSuspended(), "Watch should be suspended");
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ site-content
 .project
 .externalToolBuilders
 .checkstyle
+
+# jenv's version file
+.java-version


obfischer:
@garydgregory Could you please review the first draft, so that I could have a early feedback before I continue?

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +677,68 @@ public void unsplit() {
         splitState = SplitState.UNSPLIT;
     }
 
+    <T, E extends Throwable> T get(FailableSupplier<T, E> supplier) throws E {
```
garydgregory, line 680: Add Javadoc with a `@since` tag.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +677,68 @@ public void unsplit() {
...
+        if (isStopped()) {
+            start();
+        } else if (isSuspended()) {
+            resume();
+        }
+
```
garydgregory, line 686: Remove extra whitespace.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +677,68 @@ public void unsplit() {
...
+        try {
+            return supplier.get();
+        } finally {
+            suspend();
+        }
+    }
+
+
```
garydgregory, line 694: Remove extra whitespace.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +677,68 @@ public void unsplit() {
...
+     * @return the given function prepared to take time if applied
+     *
+     * @throws IllegalStateException if the StopWatch is not stopped or suspended
+     *
+     * @param <T> the type of the input to the function
+     * @param <R> the type of the result of the function
+     *
+     * @since 3.16
+     */
+    public <T, R> Function<T, R> apply(Function<T, R> function) {
```
garydgregory, line 727: Why is the new method above failable and this one not?

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +677,68 @@ public void unsplit() {
...
+        } finally {
+            suspend();
+        }
+    }
+
+
+    /**
+     * Take the time of the execution of a given {@linkplain Function}.
+     *
+     * <p>
```
garydgregory, line 698: Close HTML tags

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +451,29 @@ public void testToStringWithMessage() throws InterruptedException {
         final String splitStr = watch.toString();
         assertEquals(12 + MESSAGE.length() + 1, splitStr.length(), "Formatted split string not the correct length");
     }
+
+    @Test
```
garydgregory, line 455: Test failable code paths.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +451,29 @@ public void testToStringWithMessage() throws InterruptedException {
...
+        String result = watch.get(() -> {
+            StopWatchTest.this.sleep(MILLIS_200);
+            return "Foobar";
+        });
+
+        assertEquals("Foobar", result, "Watch returned result other then expected");
+        assertTrue(watch.isSuspended(), "Watch should be suspended");
+    }
+
+    @Test
```
garydgregory, line 468: Add tests for edge cases.

garydgregory (changes requested):
Hello @obfischer 

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Hi @garydgregory I don't get the question.

garydgregory: The argument to `get()` above is a `FailableSupplier`. Here, you use a `Function` but not a `FailableFunction`.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +680,368 @@ public void unsplit() {
...
+     * final StopWatch watch = StopWatch.create();
+     *
+     * String result = watch.get(() -> "A");
+     * }</pre>
+     * </p>
+     * <p>
+     *
+     * @param supplier the {@linkplain FailableSupplier} those execution should be measured
+     *
+     * @return the result of the {@linkplain FailableSupplier#get()} operation
```
garydgregory, line 698: No empty lines is needed between Javadoc tags.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +680,368 @@ public void unsplit() {
...
+    }
+
+    /**
+     * Take the time of the execution of a given {@linkplain Supplier}.
+     *
+     * <p>
+     * <b>Take the time of the execution of given {@linkplain Supplier}</b>
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     *
```
garydgregory, line 727: Remove whitespace.

garydgregory (changes requested):

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +680,368 @@ public void unsplit() {
...
+                consumer.accept(firstArg, secondArg, thirdArg);
+            } finally {
+                suspend();
+            }
+        };
+    }
+
+
+    /**
+     * Take the time of the execution of a given {@linkplain Predicate}.
```
garydgregory, line 1006: Use `@link` like the rest of the code base.

garydgregory (changes requested):
- Out of all of these new APIs, which ones do you actually need? 
- Let's not add every functional interface unless they are actually fulfill a use case
- Use failable versions

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +463,292 @@ public void testToStringWithMessage() throws InterruptedException {
...
+        @Test
+        public void testTestWhenStopWatchHasBeenSuspended() {
+            final StopWatch watch = StopWatch.create();
+
+            watch.start();
+            watch.suspend();
+
+            assertTrue(watch.isSuspended(), "Watch should be suspended");
+
+            Predicate<String> predicate = a -> true;
```
garydgregory, line 746: Also test using FailablePredicate.TRUE and other failable constants.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +463,292 @@ public void testToStringWithMessage() throws InterruptedException {
...
+            boolean result = watch.test(predicate).test("a");
+
+            assertTrue(result, "Result should be true");
+            assertTrue(watch.isSuspended(), "Watch should be suspended");
+        }
+
+        @Test
+        public void testTestWhenStopWatchHasBeenSuspended() {
+            final StopWatch watch = StopWatch.create();
+
```
garydgregory, line 740: Remove all these blank lines. If the test needs to say something else, use an inline comment `//`.

garydgregory (changes requested):

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +463,292 @@ public void testToStringWithMessage() throws InterruptedException {
...
+    public class FunctionSupportRelatedTests {
+        @Test
+        public void testApplyForFunction() {
+            final StopWatch watch = StopWatch.create();
+
+            String result = Stream.of("A", "B", "C")
+                                  .map(watch.apply((Function<String, String>) (s) -> s.toLowerCase(Locale.ROOT)))
+                                  .collect(Collectors.joining());
+
+            assertEquals("abc", result, "Returned result other then expected");
```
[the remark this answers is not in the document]
obfischer: Removed, will not add now ones.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Removed

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: s/linkplain/link/g

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +463,292 @@ public void testToStringWithMessage() throws InterruptedException {
...
+    }
+
+    @Nested
+    public class FunctionSupportRelatedTests {
+        @Test
+        public void testApplyForFunction() {
+            final StopWatch watch = StopWatch.create();
+
+            String result = Stream.of("A", "B", "C")
+                                  .map(watch.apply((Function<String, String>) (s) -> s.toLowerCase(Locale.ROOT)))
```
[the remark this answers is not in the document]
obfischer: Done

obfischer:
@garydgregory I nearly finished. I will ping you, if I am done.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: I think my tests cover all lines. I checked the coverage during writing tests in the IDE with the integrated coverage functionality

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: Removed all blank lines from the examples.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Not needed HTML tags have been removed

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: Done. I use them if possible

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Added, but not sure if the version is correct. Please let me know if I have to adjust it.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Done.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: Edge cases have been also tested

obfischer:
@garydgregory MR is ready for review

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,678 @@ public void unsplit() {
...
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     * String result = watch.get(() -> "A");
+     * }</pre>
+     *
+     * <p>
+     *  <em>Note:</em>
+     * </p>
+     * <ul>
+     *     <li>Be carefull if you pass a {@link FailableSupplier} to other methods, as the wrapped instance depends on
```
garydgregory, line 704: Fix spelling "carefull"

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,678 @@ public void unsplit() {
...
+     *
+     * @param supplier the {@link FailableSupplier} those execution should be measured
+     * @return the result of the {@link FailableSupplier#get()} operation
+     * @throws E if the supplier fails
+     * @param <T> the type of return value of the supplier
+     * @param <E> the kind of thrown exception or error
+     * @since 3.16
+     */
+    public <T, E extends Throwable> T get(FailableSupplier<T, E> supplier) throws E {
+        if (isStopped()) {
```
garydgregory, line 716: So much duplicate code... Refactor:
```java
            if (isStopped()) {
                start();
            } else if (isSuspended()) {
                resume();
            }
```
into a new private method.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,678 @@ public void unsplit() {
...
+     *         the state of used {@linkplain StopWatch}</li>
+     * </ul>
+     *
+     * @param supplier the {@link Supplier} those execution should be measured
+     * @return the result of the {@link Supplier#get()} operation
+     * @throws IllegalStateException if the {@link StopWatch} is not stopped or suspended
+     * @param <T> the type of the input to the function
+     * @since 3.16
+     */
+    public <T> T get(Supplier<T> supplier) {
```
garydgregory, line 753: supplier declaration: Use final where you can.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +470,712 @@ public void testToStringWithMessage() throws InterruptedException {
         final String splitStr = watch.toString();
         assertEquals(12 + MESSAGE.length() + 1, splitStr.length(), "Formatted split string not the correct length");
     }
+
+    @Nested
+    public class SupplierSupportRelatedTests {
+        @Test
+        public void testGet() {
+            final StopWatch watch = StopWatch.create();
+
```
garydgregory, line 479: Remove blank lines within methods. If you have something interesting to point out, use an inline comment.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +470,712 @@ public void testToStringWithMessage() throws InterruptedException {
...
+        public void testGetWhenStopWatchHasBeenSuspended() throws Throwable {
+            final StopWatch watch = StopWatch.create();
+
+            watch.start();
+            watch.suspend();
+
+            assertTrue(watch.isSuspended());
+
+            Supplier<String> supplier = () -> "Foobar";
+            String result = watch.get(supplier);
```
garydgregory, line 499: `result` local variable is not necessary.

garydgregory (changes requested):
Hello @obfischer 
Thank you for your updates. Please see my comments.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,678 @@ public void unsplit() {
...
+     *     <li>Be carefull if you pass a {@link TriConsumer} to other methods, as the wrapped instance depends on
+     *         the state of used {@linkplain StopWatch}</li>
+     * </ul>
+     *
+     * @param consumer the consumer those application should be measured
+     * @return the given consumer prepared to take time if applied
+     * @throws IllegalStateException if the {@link StopWatch} is not stopped or suspended
+     * @param <T> the type of the first argument to the consumer
+     * @param <U> the type of the second argument to the consumer
+     * @param <V> the type of the thrid argument to the consumer
```
garydgregory, line 1175: "thrid" spelling

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +470,712 @@ public void testToStringWithMessage() throws InterruptedException {
...
+
+            assertThrows(IllegalArgumentException.class, () -> watch.test(predicate).test("a", "b"));
+
+            assertTrue(watch.isSuspended());
+        }
+    }
+
+    @Nested
+    public class FailableBiConsumerSupportRelatedTests {
+        @SuppressWarnings("unchecked")
```
garydgregory, line 1050: Remove all `@SuppressWarnings("unchecked")`s and use matching a method matching the class, for example `FailableBiConsumer.nop()`.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Fixed spelling

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Introduced method `resumeOrStartStopWatch()`

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: Removed all local variables for single line invocations, but left then for multline statements

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: Removed annotations

obfischer:
> * Out of all of these new APIs, which ones do you actually need?
> 
>     * Let's not add every functional interface unless they are actually fulfill a use case
> 
>     * Use failable versions

All of them. Over the years I hat to measure some of them for statistics or debugging. So my decision was to provide an API which does not only covers my needs just in this monthm, but supports all standard interfaces in the JDK.

obfischer:
@garydgregory I think I solved all open isssues. Ready for review again

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,622 @@ public void unsplit() {
...
+     *
+     * @param predicate the predicate those application should be measured
+     * @return the given predicate prepared to take time if applied
+     * @throws IllegalStateException if the {@link StopWatch} is not stopped or suspended
+     * @param <T> the type of the argument to the predicate
+     * @param <E> The kind of thrown exception or error
+     * @since 3.17
+     */
+    public <T, E extends Throwable> FailablePredicate<T, E> test(final FailablePredicate<T, E> predicate) {
+        resumeOrStartStopWatch();
```
thc202, line 1290: Tests should catch this.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -294,7 +309,7 @@ public String formatTime() {
      * </p>
      *
      * @return the Duration.
-     * @since 3.16.0
+     * @since 3.17
```
thc202, line 312: These were most likely changed by accident.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,622 @@ public void unsplit() {
...
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     * String result = watch.get(() -> "A");
+     * }</pre>
+     *
+     * <p>
+     *  <em>Note:</em>
+     * </p>
+     * <ul>
+     *     <li>Be carefull if you pass a {@link FailableSupplier} to other methods, as the wrapped instance depends on
```
thc202, line 704: Not all typos were fixed, there are still a couple of JavaDoc with "carefull" instead of "careful".

garydgregory: Yep, good catch.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -420,7 +435,7 @@ public long getStartTime() {
      *
      * @return the Instant this StopWatch was stopped in milliseconds, between the current time and midnight, January 1, 1970 UTC.
      * @throws IllegalStateException if this StopWatch has not been started
-     * @since 3.16.0
+     * @since 3.17
```
garydgregory, line 438: The next feature version will be `3.17.0`.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,622 @@ public void unsplit() {
...
+     * </p>
+     * <ul>
+     *     <li>Be carefull if you pass a {@link FailableSupplier} to other methods, as the wrapped instance depends on
+     *         the state of used {@linkplain StopWatch}</li>
+     * </ul>
+     *
+     * @param supplier the {@link FailableSupplier} those execution should be measured
+     * @return the result of the {@link FailableSupplier#get()} operation
+     * @throws E if the supplier fails
+     * @param <T> the type of return value of the supplier
```
garydgregory, line 711: Generics parameters come _first_ in the `@` list.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,622 @@ public void unsplit() {
...
+     * <pre>{@code
+     * final StopWatch watch = StopWatch.create();
+     * String result = watch.apply((first, second) -> first + second)).apply("A", "B");
+     * }</pre>
+     *
+     * <p>
+     *  <em>Note:</em>
+     * </p>
+     * <ul>
+     *     <li>Be carefull if you pass a {@link BiFunction} to other methods, as the wrapped instance depends on
```
garydgregory, line 858: Fix ALL "carefull" spelling.

garydgregory (commented):
After looking some more at this PR, all the new methods "add" a duration to a stopwatch, which makes me wonder (as discussed on the mailing list), if all the new methods should be called "add". Food for thought...

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Sorry, I thought that I have done this. Sorry.

obfischer:
> After looking some more at this PR, all the new methods "add" a duration to a stopwatch, which makes me wonder (as discussed on the mailing list), if all the new methods should be called "add". Food for thought...

Yes, I think we should consider this. `add`, `takeTime` or `measure`. But we should keep in mind, that the actual measurement is done during the execution of some of these methods. For example `test(Predicate ..)` does not measure any time at the moment of the method call. It returns only a wrapped instance, which will measure the time if executed once or multiple times.

What do you think?

garydgregory:
Hello @obfischer 

- The tests don't assert that the stopwatch measured time. 
- Another assertion would be that the lambda asserts that the stopwatch is running while it is invoked.
- I don't understand your comment about `test(Predicate..)`, what it supposed to be `test(Predicate)` or did you mean a vararg version of the API `test(Predicate...)`? What "wrapping" are you talking about? All the new methods don't "wrap" anything in the composition sense, they do delegate to the given lambda while managing the stopwatch state before and after the delegation. I must be missing something ;-)

obfischer:
> Hello @obfischer
> 
>     * The tests don't assert that the stopwatch measured time.
> 
>     * Another assertion would be that the lambda asserts that the stopwatch is running while it is invoked. For example:
> 
> 
> ```java
>             assertEquals("Foobar", watch.get((FailableSupplier<String, Throwable>) () -> {
>                 assertTrue(watch.isStarted());
>                 return "Foobar";
>             }));
> ```
> 
>     * I don't understand your comment about `test(Predicate..)`, what it supposed to be `test(Predicate)` or did you mean a vararg version of the API `test(Predicate...)`? What "wrapping" are you talking about? All the new methods don't "wrap" anything in the composition sense, they do delegate to the given lambda while managing the stopwatch state before and after the delegation. I must be missing something ;-)

Hi @garydgregory, could I use Mockito to verify that the measurement is done properly?

garydgregory:
@obfischer 

Sure, I don't see why not.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
thc202: Why was this one marked as resolved? Test was not added nor the bug fixed.

obfischer:
> @obfischer
> 
> Sure, I don't see why not.

@garydgregory The project uses already EasyMock. I would, because I am lazy, prefer Mockito. Is it still fine to use Mockito or should I start to learn EasyMock= 😄

garydgregory:
I don't see why mocking is needed in the first place since the stopwatch state change can be asserted from the lambda, but let's see.

obfischer:
Hi @garydgregory I am very busy at the moment. I will continue to work on it in some days...

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -767,6 +769,18 @@ void testAcceptHandlesExceptionsProperly() {
             assertThrows(IllegalArgumentException.class, () -> watch.accept(consumer).accept("a", "b"));
             assertTrue(watch.isSuspended());
         }
+
+        @Test
```
obfischer, line 773: @garydgregory That is the way I would prefer to ensure that the time is really taken. WDYT?

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: You are right. Sorry, I didn't see this

garydgregory:
@obfischer 

Run `mvn` by itself before you push to catch all builds failures.

obfischer:
> @obfischer
> 
> Run `mvn` by itself before you push to catch all builds failures.

I did, but I used a JDK 22...

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +473,570 @@ public void testToStringWithMessage() throws InterruptedException {
         final String splitStr = watch.toString();
         assertEquals(12 + MESSAGE.length() + 1, splitStr.length(), "Formatted split string not the correct length");
     }
+
+    @Nested
+    public class SupplierSupportRelatedTests {
```
garydgregory, line 478: I do not think we need "Related" or "Support" in all these test class names.

A FooClassTest tests a FooClass. If you need something more precise in the naming, just say what the test does, "support" and "related" do not help understand what the test does IMO.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
```diff
@@ -671,4 +686,622 @@ public void unsplit() {
...
+     *     <li>Be careful if you pass a {@link FailableSupplier} to other methods, as the wrapped instance depends on
+     *         the state of used {@linkplain StopWatch}</li>
+     * </ul>
+     *
+     * @param <T> the type of return value of the supplier
+     * @param <E> the kind of thrown exception or error
+     * @param supplier the {@link FailableSupplier} those execution should be measured
+     * @return the result of the {@link FailableSupplier#get()} operation
+     * @throws E if the supplier fails
+     * @since 3.18.0
```
garydgregory, line 713: Hello @obfischer 

This is missing a Javadoc throws tag for IllegalArgumentException like the other methods.

Since this method calls `resumeOrStartStopWatch` like all the others, we can assume:
```Java
@throws IllegalStateException if the {@link StopWatch} is not stopped or suspended
```
BUT, either the Javadoc is wrong (in all the other methods) or the implementation is wrong. According to the Javadoc, this should pass:
```java
        @Test
        public void testGetStopped() {
            final StopWatch watch = StopWatch.create();
            watch.start();
            watch.stop();
            assertTrue(watch.isStopped());
            Supplier<String> supplier = () -> "Foobar";
            assertEquals("Foobar", watch.get(supplier));
            assertTrue(watch.isSuspended());
        }
```
This test throws `IllegalStateException` on the `get()` even though the stopwatch is stopped.

What should happen in this case?

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +473,570 @@ public void testToStringWithMessage() throws InterruptedException {
...
+            assertTrue(watch.isSuspended());
+        }
+    }
+
+    @Nested
+    public class FailableBiConsumerSupportRelatedTests {
+        @Test
+        public void testTest() {
+            final StopWatch watch = StopWatch.create();
+            FailableBiConsumer<String, String, IllegalArgumentException> consumer = FailableBiConsumer.NOP;
```
garydgregory, line 946: Avoid compiler warnings like
```
The expression of type FailableBiConsumer needs unchecked conversion to conform to FailableBiConsumer<String,String,IllegalArgumentException>
```
(from Eclipse)
By using: `FailableBiConsumer.nop()`

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
```diff
@@ -448,4 +473,570 @@ public void testToStringWithMessage() throws InterruptedException {
...
+            assertTrue(watch.isSuspended());
+        }
+
+        @Test
+        public void testTestWhenStopWatchHasBeenSuspended() {
+            final StopWatch watch = StopWatch.create();
+            watch.start();
+            watch.suspend();
+            assertTrue(watch.isSuspended());
+            FailableBiConsumer<String, String, IllegalArgumentException> consumer = FailableBiConsumer.NOP;
```
garydgregory, line 957: See above.

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
garydgregory: Hm, this assumes that `start()` and `resume()` actually work and does not actually "tests that time is really taken" (`testTimeIsReallyTaken()`), since no time is checked. To me the test name and implementation say different things.

garydgregory (changes requested):
Hi @obfischer 

Please see my comments, especially the issue with Javadoc vs. implementation.

Also please rebase on git master as I've reworked the tests so that the whole test class runs in less than a second (on my machine at least).

TY!

src/test/java/org/apache/commons/lang3/time/StopWatchTest.java
obfischer: I removed the suffix.

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Good point, this is a bug. I have to think about it.

obfischer: I would say, that the method `resumeOrStartStopWatch` should

* start the stop watch only if `runningState == State.UNSTARTED`
* resume the stop watch if `runningState.isSuspended() == true`
* every other state should lead to an `IllegalStateException`

Futhermore the JavaDoc should:

* state the behavior above 
* state that the stop watch will be suspend after taking time

WDYT?

obfischer:
Ping

src/main/java/org/apache/commons/lang3/time/StopWatch.java
obfischer: Ping @garydgregory

garydgregory:
@obfischer 
Run 'mvn' by itself to run all build checks.

obfischer:
> @obfischer Run `mvn` by itself to run all build checks.

No, my ping was for my last comment on the implementation of `resumeOrStartStopWatch`. I didn't change any line of code until now.

garydgregory:
> > @obfischer Run `mvn` by itself to run all build checks.
> 
> No, my ping was for my last comment on the implementation of `resumeOrStartStopWatch`. I didn't change any line of code until now.

Hello @obfischer 
Please make sure the build is green. I'll review again then.

garydgregory:
-1 as is, this is now deeply confusing, the functional methods should actually perform the measurement, not return a functional object that the user must then invoke to perform the measurement. The functional methods should return the same value as their arguments, or void for the functional interfaces like consumers. I'll look over in more details over the next day or two.

garydgregory:
@obfischer 

I took a simpler route than we previously discussed by only adding `StopWatch.run([Failable]Runnable)` and
`get([Failable]Supplier)` which I believe should be able to handle the most common use cases.

You are credited in `changes.xml`.

If you there are some missing use cases, let's discuss.

garydgregory:
@obfischer 
Ping.

garydgregory:
Closing, no feedback in over a month.

PR #1273, opened 2024-09-06
Fix handling of non-ASCII letters & numbers in RandomStringUtils

fabrice102:
An optimization introduced in commit f382d61a03778ccf838c6c051bd8692e4834dec2 incorrectly assumed that letters and numbers/digits are ASCII.

For example, `RandomStringUtils.random(1, 0x4e00, 0x4e01, true, false)` would take too long to terminate and would not output the correct answer, that is the string with a single character `0x4e00`. The issue is that `end` would be replaced by `'z' + 1 = 123` (the latest ASCII letter + 1) there: https://github.com/apache/commons-lang/blob/f382d61a03778ccf838c6c051bd8692e4834dec2/src/main/java/org/apache/commons/lang3/RandomStringUtils.java#L266-L270
 Thus, we would have `end < start` and `gap = end - start < 0`.

This PR fixes this issue by restricting the optimization to cases where only ASCII characters are requested, that is `end < 0x7f`. This PR also slightly clarifies the code, using constant variables instead of '0', '9', 'A', and 'z'.

The PR also includes two regression tests.

The Github diff does not show it very clearly, but all the PR does is

1. moving the optimization code inside a bug `if (chars == null && end <= 0x7f)`
2. using constant variables instead of '0', '9', 'A', and 'z'
3. moving the  test `(numbers && end <= zeroDigitAscii || letters && end <= firstLetterAscii)` inside the big `if (chars == null && end <= 0x7f)` and before the problematic optimization: this does not change the semantics and is just done for clarity of the resulting code.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -277,43 +277,48 @@ public class RandomStringUtils {
             end = Character.MAX_CODE_POINT;
         }
 
-        // Optimize generation of full alphanumerical characters
-        // Normally, we would need to pick a 7-bit integer, since gap = 'z' - '0' + 1 = 75 > 64
-        // In turn, this would make us reject the sampling with probability 1 - 62 / 2^7 > 1 / 2
-        // Instead we can pick directly from the right set of 62 characters, which requires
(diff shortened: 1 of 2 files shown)


garydgregory:
TY @fabrice102 
Merged!

PR #1278, opened 2024-09-24
Add EnumUtils.equalsAny

mfg92:
Added `EnumUtils.equalsAny` similar to `StringUtils.equalsAny`. This can help to have shorter, easier to read code.

Usage example:
```
if (EnumUtils.equalsAny(myObject.getMyField().getMyEnum(), Vehicle.CAR, Vehicle.TRUCK, Vehicle.MOTORBIKE)) {
    // do something ...
}
```

If desired, I can also implement an `EnumUtils.equalsNone`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/EnumUtils.java
+++ b/src/main/java/org/apache/commons/lang3/EnumUtils.java
@@ -420,6 +420,44 @@ public class EnumUtils {
     }
 
     /**
+     * Compares given {@code testEnum} to the enum vararg {@code searchEnums},
+     * returning {@code true} if the {@code testEnum} is equal to any of the {@code searchEnums}.
+     *
+     * <p>
+     * Examples
+     * </p>
+     *
+     * <pre>
+     * EnumUtils.equalsAny(null, (Enum[]) null)       = false
+     * EnumUtils.equalsAny(null,     null,      null)    = true
+     * EnumUtils.equalsAny(null,     MyEnum.A, MyEnum.B) = false
+     * EnumUtils.equalsAny(MyEnum.A, null,     MyEnum.B) = false
+     * EnumUtils.equalsAny(MyEnum.B, null,     MyEnum.B) = true
+     * EnumUtils.equalsAny(MyEnum.A, MyEnum.A, MyEnum.B) = true
+     * EnumUtils.equalsAny(MyEnum.B, MyEnum.A, MyEnum.B) = true
+     * EnumUtils.equalsAny(MyEnum.C, MyEnum.A, MyEnum.B) = false
+     * </pre>
+     *
+     * @param testEnum to compare, may be {@code null}.
+     * @param searchEnums a vararg of enums, may be {@code null}.
+     * @param <E> the type of the enumeration
+     * @return {@code true} if the enum is equal to any other element of {@code searchEnums};
+     * {@code false} if {@code testEnum} is null or contains no matches.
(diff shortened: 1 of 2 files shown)


aherbert:
Similar to (null checks not included):
```java
EnumSet.of(searchEnums[0], searchEnums).contains(testEnum)
```
This code would leverage the optimised EnumSet code for any Enum that has less than 64 members (as internally it uses bit flags in a long).

mfg92:
> How is this any different from `org.apache.commons.lang3.ArrayUtils.contains(Object[], Object)`?

Using this approach results in longer, less readable code, as there is no vararg parameter.
```
if (ArrayUtils.contains(new Vehicle[]{ Vehicle.CAR, Vehicle.TRUCK, Vehicle.MOTORBIKE}, myObject.getMyField().getMyEnum())) {
    // do something ...
}
```
Also, `ArrayUtils.contains` is not generic, so no type checking can be done. I can write code like this and the compiler will not warn me that this expression will always evaluate to false:
```
if (ArrayUtils.contains(new Vehicle[]{ "123", "abc"}, myObject.getMyField().getMyEnum())) {
    // do something ...
}
```


> Similar to (null checks not included):
> 
> ```java
> EnumSet.of(searchEnums[0], searchEnums).contains(testEnum)
> ```
> 
> This code would leverage the optimised EnumSet code for any Enum that has less than 64 members (as internally it uses bit flags in a long).

```
if (EnumSet.of(Vehicle.CAR, Vehicle.TRUCK, Vehicle.MOTORBIKE).contains(myObject.getMyField().getMyEnum())) {
    // do something ...
}
```
This is certainly an option, but it has two disadvantages:
- less readable than `EnumUtils.equalsAny`.
- it is not possible to: check at once if my enum is null or has one of several given values

mfg92:
What can I do to move this issue forward?

garydgregory:
Hello @mfg92 

> What can I do to move this issue forward?

You need to gather positive consensus from the community, which I do not see ATM.

src/main/java/org/apache/commons/lang3/EnumUtils.java
```diff
@@ -419,6 +419,44 @@ public static <E extends Enum<E>> EnumSet<E> processBitVectors(final Class<E> en
...
+     * @param searchEnums a vararg of enums, may be {@code null}.
+     * @param <E> the type of the enumeration
+     * @return {@code true} if the enum is equal to any other element of {@code searchEnums};
+     * {@code false} if {@code testEnum} is null or contains no matches.
+     * @since 3.18.0
+     */
+    @SafeVarargs
+    public static <E> boolean equalsAny(E testEnum, E... searchEnums) {
+        if (ArrayUtils.isNotEmpty(searchEnums)) {
+            for (final E searchEnum : searchEnums) {
```
cdavid16, line 451: Hello, 
I think it would be good to add a contains method in ArraysUtils the reuse it here, since you are basically doing a search in the array.

garydgregory:
Hello @mfg92 
How about using the existing API, like this (from `ArrayUtilsTest`):
```java
    @Test
    public void testContainsAnyEnum() {
        assertTrue(ArrayUtils.containsAny(ElementType.values(), ElementType.ANNOTATION_TYPE));
        assertFalse(ArrayUtils.containsAny(ElementType.values(), (ElementType) null));
    }
```
?

garydgregory:
Closing: No reply in over 3 weeks.

PR #1279, opened 2024-09-26
LANG-1752 seperated test cases of Strings from StringUtils

MSaifAsif:
Separated out test cases from StringUtils. Added more cases for null check, it seems that we are running into a NPE when we use the new `Strings.CS.equals(null,null)` method. See [line](https://github.com/apache/commons-lang/pull/1279/files#diff-190bb6d0ff6389981367f96b7e05541b5c7598c718e611d04f07d9cf66e7f65bR124) where I created a test cases to highlight the failure. When using `StringUtils.equals(null, null)` it doesn't throw a NPE. IMO We should use the same functionality when using either of the `Strings.CS` or `Strings.CI` equals method

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -231,7 +231,7 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
     /**
      * Gets the field names that should be excluded from the diff.
      *
-     * @return Returns the excludeFieldNames.
+     * @return the excludeFieldNames.
      * @since 3.13.0
      */
     public String[] getExcludeFieldNames() {


src/test/java/org/apache/commons/lang3/StringUtilsContainsTest.java
```diff
@@ -37,54 +34,22 @@ public class StringUtilsContainsTest extends AbstractLangTest {
     public void testContains_Char() {
         assertFalse(StringUtils.contains(null, ' '));
         assertFalse(StringUtils.contains("", ' '));
-        assertFalse(StringUtils.contains("", null));
```
garydgregory, line 40: @MSaifAsif 
Do NOT delete existing tests please.

MSaifAsif: I migrated it over [here](https://github.com/apache/commons-lang/pull/1279/files#diff-26c06c4f25342a7e49477cabfc7198f12a7d76023e2a7f3987d2341619669eb6R33). Would you prefer that we keep the existing StringUtils test cases as well on top of the separation that I did ?

MSaifAsif:
Reverted back the existing test cases @garydgregory I thought about it and yes its too risky of a move to not keep them. Reverted them back

PR #1295, opened 2024-10-13
addition of filledString() method

Claudenw:
Implements [LANG-1756](https://issues.apache.org/jira/browse/LANG-1756)

Added `filledString(int length, char fillChar)` and associated test.


Claudenw:
Failed tests are unassociated with this change

PR #1296, opened 2024-10-14
Fix broken URL to project location in Maven Central

ParanoidUser:
Ahoy @garydgregory, 👋

It appears that [https://maven-badges.herokuapp.com ](https://maven-badges.herokuapp.com/)is no longer functioning, resulting in a broken Maven Central badge in the README. I've noticed that other Apache projects (e.g., [maven-stage-plugin](https://github.com/apache/maven-stage-plugin/blob/master/README.md)) use [shields.io ](https://shields.io/badges/maven-central-version) as an alternative. If there are no plans to restore the maven-badges app, this is a trivial patch to fix it. If you find it appropriate, I can raise similar PRs for other repositories in the Apache group affected by this issue.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Apache Commons Lang
 ===================
 
 [![Java CI](https://github.com/apache/commons-lang/actions/workflows/maven.yml/badge.svg)](https://github.com/apache/commons-lang/actions/workflows/maven.yml)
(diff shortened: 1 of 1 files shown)


garydgregory:
Hello @ParanoidUser 
The readme files are generated from the commons-build-plugin, so that's the one that really needs fixing.

PR #505, opened 2020-03-24
[LANG-1528] StringUtils.replaceEachRepeatedly gives IllegalStateException

edelgadoh:
I fixed this simple bug, I'd like to collaborate on it, I'm ready to consider any suggestions

PR #1297, opened 2024-10-15
[LANG-1753] StringUtils.replaceEachRepeatedly regression in 3.11+

ParanoidUser:
https://issues.apache.org/jira/browse/LANG-1753

The method `replaceEachRepeatedly` calls `replaceEach` repeatedly until no more replacements are possible or `timeToLive < 0`. The `timeToLive` starts with the value of `searchList.length`. The current implementation can't distinguish between infinite loops and deeply nested string patterns. In both cases, it continues replacements until it hits the time-to-live limit and throws `IllegalStateException` if replacements are still possible.

The problem with [LANG-1528](https://issues.apache.org/jira/browse/LANG-1528) fix is that in some cases, it allows partially replaced strings after TTL is exceeded. This fix reverts `replaceEach` logic back to version 3.10 (see PR #505), but allows deeper dives due to an increased initial TTL value. The initial TTL value is now set to the greater of `searchList.length` or up to 5 levels deep. This default value was chosen based on rational considerations.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -21,12 +21,10 @@ import java.nio.charset.Charset;
 import java.text.Normalizer;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
-import java.util.Set;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
@@ -180,6 +178,11 @@ public class StringUtils {
      * The maximum size to which the padding constant(s) can expand.
      */
     private static final int PAD_LIMIT = 8192;
+
+    /**
+     * The default maximum depth at which recursive replacement will continue until no further search replacements are possible.
+     */
+    private static final int DEFAULT_TTL = 5;
 
     /**
      * Pattern used in {@link #stripAccents(String)}.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -6611,7 +6608,8 @@ private static String replaceEach(
      * @since 2.4
      */
     public static String replaceEachRepeatedly(final String text, final String[] searchList, final String[] replacementList) {
-        return replaceEach(text, searchList, replacementList, true, ArrayUtils.getLength(searchList));
+        int timeToLive = Math.max(ArrayUtils.getLength(searchList), DEFAULT_TTL);
```
palacsint, line 6611: The Javadoc only mentions endless loops ("if the search is repeating and there is an endless loop"). It would be clearer if it also included a mention of the TTL to avoid confusion.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -1990,7 +1990,53 @@ public void testReplace_StringStringArrayStringArrayBoolean() {
...
         assertEquals("tcte", StringUtils.replaceEachRepeatedly("abcde", new String[]{"ab", "d"}, new String[]{"d", "t"}));
-        assertEquals("blaan", StringUtils.replaceEachRepeatedly("blllaan", new String[]{"llaan"}, new String[]{"laan"}) );
+
+        // Test recursive replacement - LANG-1528 & LANG-1753
+        assertEquals("blaan", StringUtils.replaceEachRepeatedly("blllaan", new String[]{"llaan"}, new String[]{"laan"}));
+        assertEquals("blaan", StringUtils.replaceEachRepeatedly("bllllaan", new String[]{"llaan"}, new String[]{"laan"}));
+
+        // Test default TTL for smaller search lists. 32 characters reduced to 16, then 8, 4, 2, 1.
+        assertEquals("a", StringUtils.replaceEachRepeatedly("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+                new String[]{"aa"}, new String[]{"a"}));
```
palacsint, line 2000: I would suggest adding an example of `IllegalStateException` with a shorter input, like:

	assertThrows(
		IllegalStateException.class,
		() -> StringUtils.replaceEachRepeatedly("aaaaaab", new String[] {"ab"}, new String[] {"b"}),
		"Should be a circular reference");

It might surprise newcomers that even far fewer than 32 repetitions can trigger an `IllegalStateException`.

palacsint:
Thanks for the fix! Seems perfect for us.

PR #1298, opened 2024-10-15
Use simplified JUnit assertion methods

ParanoidUser:
The primary objective of this PR is to start using simplified JUnit assertion methods, primarily [Assertions.assertInstanceOf()](https://junit.org/junit5/docs/5.10.0/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertInstanceOf(java.lang.Class,java.lang.Object)), which became stable with the JUnit 5.10.0 release in July 2023.

Additionally, a few test classes that cover deprecated source classes are also marked as deprecated to simplify housekeeping in 4.0 release.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/CharEncodingTest.java
+++ b/src/test/java/org/apache/commons/lang3/CharEncodingTest.java
@@ -28,9 +28,9 @@ import org.junit.jupiter.api.Test;
 /**
  * Tests CharEncoding.
  *
- * @see CharEncoding
+ * @deprecated this test can be removed once the deprecated source class {@link org.apache.commons.lang3.CharEncoding} is removed.
  */
-@SuppressWarnings("deprecation")
+@Deprecated
 public class CharEncodingTest extends AbstractLangTest {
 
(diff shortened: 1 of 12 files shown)


ParanoidUser:
This is unrelated to the PR subject, but I don't want to draw too much attention to this minor tweak in a separate PR. As part of , we added the mockito-core dependency in [this commit](https://github.com/apache/commons-lang/commit/da246244aebf91571b0689e3dee97fce0ad52b23), which is not actually used anywhere. Thus, I am removing it as unnecessary.

This is unrelated to the PR subject, but I don't want to draw too much attention to this minor tweak. As part of [LANG-1747](https://issues.apache.org/jira/projects/LANG/issues/LANG-1747), the `mockito-core` dependency was added in [this commit](https://github.com/apache/commons-lang/commit/da246244aebf91571b0689e3dee97fce0ad52b23), which is not actually used anywhere. Thus, it is being removed as unnecessary.

ppkarwasz (approved):
Looks good to me, thanks!

PR #1311, opened 2024-11-04
Removed the if-else and replaced with clearer data flow.

mosbat:
Just minor change in coding style to make code clearer and easier to manage.

I didn't create a Jira issue since as per the documentation, it's a minor issue and not a bug or feature. I'm getting accustomed to this project.

P.S.
This is my first PR with this project, I did go through your documentations/guides; I'd appreciate a bit of patience in case I missed something or if I didn't follow a certain point instead of rejecting my PR outright and leaving me confused. Thank you

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -272,14 +272,12 @@ public class WordUtils {
             final char ch = str.charAt(i);
             if (isDelimiter(ch, delimiters)) {
                 lastWasGap = true;
-                continue;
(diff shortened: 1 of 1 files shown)


garydgregory:
@mosbat 
Why did you remove the comment?

mosbat:
> @mosbat Why did you remove the comment? This change feels pretty arbitrary anyway.

I thought at the beginning that it should be obvious without the comment; but thinking about it again, I put back the comment just now since the comment makes it clearer.

It's a minor change since I hate if-else statements :) since they are nighmarish but it can be also just personal preference :)

PR #1312, opened 2024-11-04
LANG-1743 - when ArrayUtils.removeAll meeting 'null' array (PROPOSED CHANGE)

mosbat:
Issue details can be found here:
This is a proposed changed, not arbitrary. 
https://issues.apache.org/jira/browse/LANG-1743

P.S.
For project owners, please note that I do not expect you to like or approve this change so feel free to reject it if you don't agree with the issue; but if you like it and would like a change, you're welcome to request a change.


garydgregory (commented):
I doubt this is a good idea since it will blow up apps that previously didn't.

mosbat:
> I doubt this is a good idea since it will blow up apps that previously didn't.

You got a good point. I'll try something less invasive. I'll update this PR

garydgregory:
Closing: No more feedback.

PR #1322, opened 2024-11-30
Improve container detection if current process is pid 1

maxxedev:
If current process is pid 1, read directly from `System.getenv()` instead of performing file i/o on `/proc/1/environ`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
+++ b/src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
@@ -18,6 +18,7 @@
 package org.apache.commons.lang3;
 
 import java.io.IOException;
+import java.lang.management.ManagementFactory;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


ppkarwasz:
@maxxedev,

Can you explain in what consists the improvement? AFAIK the JVM will never have PID 1. Even if it does, checking `/proc/1/environ` and the environment of the JVM is equivalent.

maxxedev:
@ppkarwasz 

> JVM will never have PID 1

JVM can certainly be PID 1 inside a container. Example:
```
$ cat App.java
public class App {

        public static void main(String... args) {
                var mxBean = java.lang.management.ManagementFactory.getRuntimeMXBean();
                System.out.println(mxBean.getName());
        }
}

$ docker run -it --entrypoint java -v `pwd`:/tmp/ openjdk:21 /tmp/App.java
1@afecfa5e25bb
```

&nbsp;
> checking `/proc/1/environ` and the environment of the JVM is equivalent.

Yes, equivalent but perhaps reading from `System.getenv()` is a slight improvement by avoiding file i/o on procfs

ppkarwasz:
> > JVM will never have PID 1
> 
> JVM can certainly be PID 1 inside a container. Example:

Point taken.
 
> > checking `/proc/1/environ` and the environment of the JVM is equivalent.
> 
> Yes, equivalent but perhaps reading from `System.getenv()` is a slight improvement by avoiding file i/o on procfs

Your code requires the `java.management` Java module to handle a specific case that was already handled correctly by the previous code (which only requires `java.base`). IMHO this is not an improvement.

PR #1323, opened 2024-11-30
Improve container detection by mimicing systemd

maxxedev:


src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
```diff
@@ -63,6 +63,9 @@ public static Boolean inContainer() {
      */
     // Could be public at a later time.
     static Boolean inDocker() {
+        if (Files.exists(Paths.get("/.dockerenv"))) {
```
maxxedev, line 66: see https://superuser.com/a/1021925

src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
```diff
@@ -76,6 +79,9 @@ static Boolean inDocker() {
      */
     // Could be public at a later time.
     static Boolean inPodman() {
+        if (Files.exists(Paths.get("/run/.containerenv"))) {
```
maxxedev, line 82: see https://docs.podman.io/en/stable/markdown/podman-run.1.html
>  a container environment file is created in each container to indicate to programs they are running in a container. This file is located at /run/.containerenv (or /var/run/.containerenv for FreeBSD containers).

ppkarwasz:
Hi @maxxedev,

Since container detection is pretty much heuristic, I think that we should mimic the code of projects that **really** need to know if they are running in a container and put a link to that project in a comment, so we can follow its evolution.

The best reference I found is [`detect_container` in the `virt.c` source of SystemD](https://github.com/systemd/systemd/blob/main/src/basic/virt.c). For SystemD it is essential to know if it is running in a container, so we could follow its code to the letter. IMHO the overall logic of `inContainer()` should be:

1. We check the `container` environment variable of process `1` as an **opt-in** way for container developers to declare the type of container:
    - If the variable is **empty**, we return `false`. This includes the case, where the container developer wants to hide the fact that the application runs in a container.
    - If the variable is **not** empty, we return `true`.
    - If the variable is **absent**, we continue.
2. We check files in the container. According to SystemD:
    - `/.dockerenv` is used by Docker.
    - `/run/.containerenv` is used by PodMan.
    - I would ignore other files (e.g. `/var/run/.containerenv`) or first make a PR for SystemD and if they accept it, use it in Commons Lang.

garydgregory:
I like @ppkarwasz's approach, let's do that.

src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
```diff
@@ -29,67 +31,59 @@
...
+    private static String getenv(final String envVarFile, final String prefix) {
+        try {
+            byte[] bytes = Files.readAllBytes(Paths.get(envVarFile));
+            String content = new String(bytes, UTF_8);
+            // Split by null byte character
+            String[] lines = content.split("\u0000");
+            return Arrays.stream(lines).filter(test -> test.startsWith(prefix))
+                    .map(test -> StringUtils.substringAfter(test, prefix))
+                    .findFirst()
+                    .orElse(null);
```
ppkarwasz, line 53: _Nit_: Maybe we could split the lines using `line.split("=", 2)` first, so we don't need to use `StringUtils` after.

src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
```diff
@@ -29,67 +31,59 @@
  */
 public class RuntimeEnvironment {
 
+    // package-private non-static fields for testing.
+    static String rootDir = "/";
```
ppkarwasz, line 35: _Nit_: for testing I would prefer to create a package-private `inContainer(String rootDir)` method, instead of a static field.

src/main/java/org/apache/commons/lang3/RuntimeEnvironment.java
```diff
@@ -29,67 +31,59 @@
...
-        return containsLine("/proc/1/environ", "container=podman");
+        We check files in the container. According to SystemD:
+        /.dockerenv is used by Docker.
+        /run/.containerenv is used by PodMan.
+
+         */
+        String value = getenv(rootDir + "proc/1/environ", "container=");
+        return StringUtils.isNotEmpty(value)
+                || fileExists(rootDir + ".dockerenv")
+                || fileExists(rootDir + "run/.containerenv");
```
ppkarwasz, line 82:
```suggestion
        return value != null ?
                !value.isEmpty() :
                fileExists(rootDir + ".dockerenv") || fileExists(rootDir + "run/.containerenv");
```
If `value` is not `null`, the files should not be checked.

src/test/java/org/apache/commons/lang3/RuntimeEnvironmentTest.java
```diff
@@ -17,21 +17,85 @@
...
+    @TempDir
+    private Path tempDir;
+
+    @AfterAll
+    public static void tearDown() {
+        RuntimeEnvironment.rootDir = "/";
+    }
+
+    @TestFactory
+    public DynamicTest[] testIsContainerDocker() {
```
ppkarwasz, line 55: Can you add a test case, where the `container` variable exists, but is empty?

Also: since all the test cases are pretty similar, could you use `@ParameterizedTest` instead of `DynamicTest`?

ppkarwasz (changes requested):
@maxxedev,

This looks very nice, thanks! :100:
I dropped some comments below.

ppkarwasz (approved):
LGTM, thanks!

PR #1324, opened 2024-11-30
Improve reading from null-separated /proc/N/environ file

maxxedev:


ppkarwasz:
This was replaced by #1323.

PR #1325, opened 2024-12-02
Continuation of add StringUtils.substringBeforeLast(String, int)

Serwios:
Hi, it's continuation of #967 , basically just added tests to @mmariotti PR, please review @arturobernalg @garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8529,6 +8529,41 @@ public class StringUtils {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Serwios:
Hi, request to review, thanks @garydgregory

garydgregory:
- I don't think this is needed or that we should cover every combination of before/after/char/int/any/not and so on.
- don't delete existing tests
- whenever I see int for a character, I think about code points, UTF-16, and supplementary characters which the tests here don't cover.

Serwios:
I didn't delete any tests, just added new one, old test testSubstringBetween_StringString is below my test

If you think it's not needed then I'll close this pr

PR #1326, opened 2024-12-03
[tests] correct high/low naming of UTF-16 surrogate character test fixture constants

IBue:
The naming of the UTF-16 surrogate character test fixtures is semantically interchanged regarding "high" and "low".
This pull request corrects and clarifies the naming and also composes the supplementary character fixtures of the surrogate character test fixture constants where applicable.


garydgregory:
Hello @IBue 

Good catch! TY. 

The issue with this PR IMO is that it does more than addressing the name issue. Changing the tests makes reviewing and checking if something went wrong harder. For example, it's unclear if you are also fixing something in the test.  If you update your git master branch, you'll see I took a simpler approach and only changed the 2 constant names and updated the Javadoc.

You are credited in `changes.xml`.

IBue:
Thank you for your quick review and taking this into account.
Well, but actually, I still find my patches easier to synoptically check for — as intended — factual test identity than your rework (especially in split view)…

PR #1327, opened 2024-12-03
[StringUtils::indexOfAnyBut] redesign due to inconsistent/faulty behaviour regarding UTF-16 surrogates

IBue:
depends/stacked on #1326 

Both signatures of `StringUtils::indexOfAnyBut` currently behave
inconsistently in matching UTF-16 supplementary characters and single
UTF-16 surrogate characters (i.e. paired and unpaired surrogates), since
they differ unnecessarily in their algorithmic implementations, use
their own incomplete and faulty interpretation of UTF-16 and don't take
full advantage of the standard library.

The example cases below show that they may yield contradictory results
or correct results for the wrong reasons.

This proposal gives a unified algorithmic implementation of both
signatures that

- a) is much easier to grasp due to a clear mathematical set approach and
      safe iteration and doesn't become entangled in index arithmetic;
      stresses the set semantics of the 2nd argument
- b) fully relies on the standard library for defined UTF-16
       handling/interpretation;
      paired surrogates are merged into one codepoint, unpaired surrogates
      are left as they are
- c) scales _much_ better with input sizes and result index position
- d) can benefit from current and future improvements in the standard
       library and JVM
       (streams implementation, parallelization, JIT optimization, JEP 218,  ???…)

The algorithm boils down to:
**find index i of first char in srcChars such that**
`(srcChars.codePointAt(i) ∈ {x ∈ codepoints(srcChars) ∣ x ∉ codepoints(searchChars) })`

Examples:
---------

`<H>`: high-surrogate character
`<L>`: low-surrogate character
`(<H><L>)`: valid supplementary character
signature 1: `StringUtils::indexOfAnyBut(final CharSequence srcChars, final CharSequence searchChars)`
signature 2: `StringUtils::indexOfAnyBut(final CharSequence srcChars, final char... searchChars)`

**Case 1:** matching of unpaired high-surrogate
|| srcChars | searchChars | expected/new  |   sig.1    |   sig.2   |
|---|---|---|---|---|---|
| 1.1  | `<H>aaaa` | `<H>abcd` | !found | !found | !found |  |
| | | | | sig.1: 'a' is _somewhere_ in `searchChars` |   sig.2: 'a' happens to follow `<H>` in `searchChars`; |
| 1.2  | `<H>baaa` |   `<H>abcd`  | !found | !found| 0 |
| | | | | sig.1: 'b' is _somewhere_ in searchChars |
| 1.3  | `<H>aaaa` |  `(<H><L>)abcd` | 0 |  !found   | 0 |
| | | | | sig.1: 'a' is _somewhere_ in searchChars |
| 1.4  | `aaaa<H>` | `(<H><L>)abcd`  | 4 | !found | !found | sig.1+2 don't interpret suppl. character |

**Case 2:** matching of unpaired low-surrogate
|| srcChars | searchChars | expected/new  |   sig.1    |   sig.2   |
|---|---|---|---|---|---|
| 2.1 | `<L>aaaa` | `(<H><L>)abcd` |  0 | !found | !found | 
| | | | | | sig.1+2 don't interpret suppl. character |
| 2.2 | `aaaa<L>` | ` (<H><L>)abcd` | 4 | !found | !found |
| | | | | | sig.1+2 don't interpret suppl. character |

**Case 3:** matching of supplementary character
|| srcChars | searchChars | expected/new  |   sig.1    |   sig.2   |
|---|---|---|---|---|---|
| 3.1 | `(<H><L>)aaaa` | `<L>ab<H>cd` | 0 | !found | 0 |
| | | | | sig.1: `<L>` is _somewhere_ in `searchChars` | |
| 3.2 | `(<H><L>)aaaa` | `abcd` | 0 | 1 | 0 |
||| | | sig.1 always points to low-surrogate of (fully) unmatched  suppl. character | |
| 3.3 | `(<H><L>)aaaa` | `abcd<H>` | 0 | 0 | 1 |
| 3.4  | `(<H><L>)aaaa` | `abcd<L>` | 0 | !found | 0 |
| | | | | sig.1: `<H>` skipped by algorithm | |


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2829,41 +2835,23 @@ public static int indexOfAny(final CharSequence cs, final String searchChars) {
 
      * </pre>
      *
-     * @param cs  the CharSequence to check, may be null
+     * @param srcChars  the CharSequence to check, may be null
```
garydgregory, line 2838: Don't rename parameters in methods, it makes the PR noisy and makes it harder to review.

src/test/java/org/apache/commons/lang3/Supplementary.java
```diff
@@ -25,25 +25,24 @@
 public class Supplementary {
 
     /**
-     * Supplementary character U+20000 See https://www.oracle.com/technical-resources/articles/javase/supplementary.html
+     * Incomplete supplementary character U+20000, low surrogate only. See
+     * https://www.oracle.com/technical-resources/articles/javase/supplementary.html
      */
-    static final String CharU20000 = "\uD840\uDC00";
+    static final String CharU20000SupplCharLow = "\uDC00";
```
garydgregory, line 31: These changes duplicate your changes in your other PR. Rebase on git master and don't make these changes here.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2878,27 +2866,30 @@ public static int indexOfAnyBut(final CharSequence cs, final char... searchChars
...
      * @param searchChars  the chars to search for, may be null
      * @return the index of any of the chars, -1 if no match or null input
      * @since 2.0
      * @since 3.0 Changed signature from indexOfAnyBut(String, String) to indexOfAnyBut(CharSequence, CharSequence)
      */
-    public static int indexOfAnyBut(final CharSequence seq, final CharSequence searchChars) {
-        if (isEmpty(seq) || isEmpty(searchChars)) {
+    public static int indexOfAnyBut(final CharSequence srcChars, final CharSequence searchChars) {
+        if (srcChars == null || searchChars == null || srcChars.length() == 0 || searchChars.length() == 0) {
```
garydgregory, line 2876: Not sure why you inlined these calls. Keep the PR to only the required changes to fix a bug.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2878,27 +2866,30 @@ public static int indexOfAnyBut(final CharSequence cs, final char... searchChars
...
-            final char ch = seq.charAt(i);
-            final boolean chFound = CharSequenceUtils.indexOf(searchChars, ch, 0) >= 0;
-            if (i + 1 < strLen && Character.isHighSurrogate(ch)) {
-                final char ch2 = seq.charAt(i + 1);
-                if (chFound && CharSequenceUtils.indexOf(searchChars, ch2, 0) < 0) {
-                    return i;
-                }
-            } else if (!chFound) {
-                return i;
+        final Set<Integer> srcSetCodePoints = srcChars.codePoints().boxed().collect(Collectors.toSet()); // >=10: Collectors::toUnmodifiableSet
```
garydgregory, line 2879: What does comments like >=10, 11, ... mean here? Can you be more explicit? Are these magic numbers?

garydgregory (changes requested):
Hello @IBue 

Please rebate on git master and see my comments. It's hard to read the changes since some changes make the PR noise than it has to be, like unnecessary parameter name changes. Also the PR contains duplicate work from your other PR. 

TY!

IBue:
I have rebased onto `master` and have followed your comments.

garydgregory:
This one needs a second pair of eyes... @ppkarwasz @chtompki @struberg ?

One big difference is that the current implementation does not allocate _any_ extra memory. I wonder what happens to performance of existing apps that call these methods on large strings.

garydgregory:
> I have rebased onto `master` and have followed your comments.

Hi @IBue 

It looks like you forgot to run `mvn` (by itself) before pushing, see the checkstyle error in `StringUtils`.

ppkarwasz (changes requested):

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2888,17 +2876,20 @@ public static int indexOfAnyBut(final CharSequence seq, final CharSequence searc
...
-                if (chFound && CharSequenceUtils.indexOf(searchChars, ch2, 0) < 0) {
-                    return i;
-                }
-            } else if (!chFound) {
-                return i;
+        final Set<Integer> seqSetCodePoints = seq.codePoints().boxed().collect(Collectors.toSet()); // JDK >=10: Collectors::toUnmodifiableSet
+        final Set<Integer> searchSetCodePoints = searchChars.codePoints().boxed()
+                .collect(Collectors.toSet()); // JDK >=10: Collectors::toUnmodifiableSet
+        final Set<Integer> complSetCodePoints = seqSetCodePoints.stream().filter(((Predicate<Integer>) searchSetCodePoints::contains).negate()) // JDK >=11: Predicate.not(searchSetCodePoints::contains)
+                .collect(Collectors.toSet()); // JDK >=10: Collectors::toUnmodifiableSet
```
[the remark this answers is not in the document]
garydgregory: We can't jump off of Java 8 yet. I think we'd leave too many people behind. At least based on current surveys.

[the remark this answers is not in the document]
ppkarwasz: We don't need to use Java 9. The above stream expression is almost identical to the [loop I posted below](https://github.com/apache/commons-lang/pull/1327#discussion_r1873091371).

In this case the use of a `Stream` seems pretty much syntactic sugar.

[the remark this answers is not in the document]
IBue: > 
> ```java
> Set<Integer> searchCodePoints = searchChars.codePoints().boxed().collect(Collectors.toSet());
> int idx = seq.codePoints()
>         .takeWhile(c -> !searchCodePoints.contains(c))
>         .reduce(0, (count, codePoint) -> count + Character.charCount(codePoint));
> return idx < seq.length() ? idx : INDEX_NOT_FOUND;
> ```

This is a really clever stream-native solution (though `!searchCodePoints.contains…` is the `StringUtils::indexOfAny()` variant)
that appears to have almost no memory footprint penalty and is even 2-3× faster than this pull-request.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2888,17 +2876,20 @@ public static int indexOfAnyBut(final CharSequence seq, final CharSequence searc
...
-        for (int i = 0; i < strLen; i++) {
-            final char ch = seq.charAt(i);
-            final boolean chFound = CharSequenceUtils.indexOf(searchChars, ch, 0) >= 0;
-            if (i + 1 < strLen && Character.isHighSurrogate(ch)) {
-                final char ch2 = seq.charAt(i + 1);
-                if (chFound && CharSequenceUtils.indexOf(searchChars, ch2, 0) < 0) {
-                    return i;
-                }
-            } else if (!chFound) {
-                return i;
```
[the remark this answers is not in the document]
IBue: <ins>note on performance:</ins>
boxing is the main responsible for increase in memory footprint here, that is +3GB for a 100M input sequence,
but the current implementation takes already a 10² order of magnitude more runtime than the proposed one for a 10M input sequence (minutes!) and a result index at half of the input size.

src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -2888,17 +2875,17 @@ public static int indexOfAnyBut(final CharSequence seq, final CharSequence searc
...
+                .collect(Collectors.toSet()); // JDK >=10: Collectors::toUnmodifiableSet
+        for (final ListIterator<Integer> seqListIt = seq.chars().boxed().collect(Collectors.toList()) // JDK >=16: Stream::toList, JDK >=10: Collectors::toUnmodifiableList
+                .listIterator(); seqListIt.hasNext(); seqListIt.next()) {
+            final int curSeqCharIdx = seqListIt.nextIndex();
+            final int curSeqCodePoint = Character.codePointAt(seq, curSeqCharIdx);
+            if (!searchSetCodePoints.contains(curSeqCodePoint)) {
+                return curSeqCharIdx;
+            }
+            if (Character.isSupplementaryCodePoint(curSeqCodePoint)) {
+                seqListIt.next(); // skip subsequent low-surrogate in next loop, since it merged into curSeqCodePoint
```
ppkarwasz, line 2888:
```suggestion
        int seqLength = seq.length();
        int curSeqCodePoint;
        // Skips the second character of a surrogate pair
        for (curSeqCharIdx = 0; curSeqCharIdx < seqLength; curSeqCharIdx += Character.charCount(curSeqCodePoint)) {
            curSeqCodePoint = Character.codePointAt(seq, curSeqCharIdx);
            if (!searchSetCodePoints.contains(curSeqCodePoint)) {
                return curSeqCharIdx;
            }
```

Using a `ListIterator` has IMHO the same disadvantage of creating sets: you need first to create a `List` (i.e. consume all the code points in `seq`) and only then you start to iterate. I don't see why we can't use a simple loop.

IBue:
<ins>note on performance:</ins>
boxing is the main responsible for increase in memory footprint here, that is +3GB for a 100M input sequence,
but the current implementation takes already a 10² order of magnitude more runtime than the proposed one for a 10M input sequence (minutes!) and a result index at half of the input size.

ppkarwasz:
@IBue,

> note on performance: boxing is the main responsible for increase in memory footprint here, that is +3GB for a 100M input sequence, but the current implementation takes already a 10² order of magnitude more runtime than the proposed one for a 10M input sequence (minutes!) and a result index at half of the input size.

Can you tell us more on the usage you have for this method? That will tell us a lot on how it should be optimized.

A common usage for this method could be extracting contiguous series of digits from a long string. With series of 10-20 digits it doesn't really make sense to go parallel, but if you want to validate a 100M input to make sure it only contains digits, you can sacrifice more memory.

**Note**: The `indexOfAnyBut` method can have multiple implementations and switched between them based on the input size.

garydgregory:
@IBue 
Please reply to @ppkarwasz's question:
> Can you tell us more on the usage you have for this method? That will tell us a lot on how it should be optimized.

IBue:
I have transformed the `ListIterator` loop into the proposed simplified index-based one.
Does anyone know if the index has to be protected against overflow?


> Can you tell us more on the usage you have for this method? That will tell us a lot on how it should be optimized.

Actually, I don't have productive usage for this (yet). I stumbled upon the problem in the context of UTF-16 and learning some Java.

src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -444,11 +446,24 @@ public void testIndexOfAnyBut_StringCharArray() {
     }
 
     @Test
-    public void testIndexOfAnyBut_StringCharArrayWithSupplementaryChars() {
+    public void testIndexOfAnyBut_StringCharArrayWithSurrogateChars() {
```
ppkarwasz, line 449: Could you replace this with a `@ParamTest`?

src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
```diff
@@ -469,11 +484,24 @@ public void testIndexOfAnyBut_StringString() {
     }
 
     @Test
-    public void testIndexOfAnyBut_StringStringWithSupplementaryChars() {
+    public void testIndexOfAnyBut_StringStringWithSurrogateChars() {
```
ppkarwasz, line 487: Same here (`@ParamTest`).

ppkarwasz (approved):
@IBue,

This looks good to me.

ppkarwasz:
@garydgregory,

> Another issue here is that for every single step in the `for` loop, the PR does a _sequential_ search in a collection. Is it worth doing a _binary_ search instead (at the cost for sorting the array)? Like this:

Currently it is a lookup in a `Set` (not a generic collection):

https://github.com/apache/commons-lang/blob/e0e3f0890c9164fcc9e76334aadd870dd4c8ef2a/src/main/java/org/apache/commons/lang3/StringUtils.java#L2877-L2887

garydgregory:
@IBue Would you please rebase on git master?

IBue:
I have parameterized the test functions and rebased onto master.

(You made premature, conflicting changes without proper comment in 1a89cca97e397feaf3c18924b86e74e2b5cda2ac )

ppkarwasz (approved):
LGTM

garydgregory:
@IBue @ppkarwasz 
Merged! TY for the work. Do you see where else Lang might have issues with UTF-16 surrogates?

IBue:
Thank you for reviewing and valuable feedback!
I am preparing the transfer to `StringUtils:indexOfAny`; and I think there are still quite a few functions that need reconsideration regarding UTF-16 correctness and scalability.

garydgregory:
@IBue 
Let us know what you find 👀

PR #1328, opened 2024-12-04
Fb jpms - fixes for illegal access to private information in java.* packages.

struberg:
please review and give feedback. Can push it later myself or you just apply it - txs!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -459,11 +459,6 @@
         <jdk>[9,)</jdk>
       </activation>
       <properties>
-        <!-- LANG-1265: allow tests to access private fields/methods of java.base classes via reflection -->
-        <!-- LANG-1667: allow tests to access private fields/methods of java.base/java.util such as ArrayList via reflection -->
-<!--
-        <argLine>-Xmx512m &#45;&#45;add-opens java.base/java.lang.reflect=ALL-UNNAMED &#45;&#45;add-opens java.base/java.lang=ALL-UNNAMED &#45;&#45;add-opens java.base/java.util=ALL-UNNAMED</argLine>
--->
         <argLine>-Xmx512m</argLine>
       </properties>
     </profile>
--- a/pom.xml
+++ b/pom.xml
@@ -202,6 +202,7 @@
               <exclude>src/site/resources/download_lang.cgi</exclude>
               <exclude>src/site/resources/release-notes/RELEASE-NOTES-*.txt</exclude>
               <exclude>src/test/resources/lang-708-input.txt</exclude>
+              <exclude>nb-configuration.xml</exclude>
             </excludes>
           </configuration>
         </plugin>
@@ -460,7 +461,10 @@
       <properties>
         <!-- LANG-1265: allow tests to access private fields/methods of java.base classes via reflection -->
         <!-- LANG-1667: allow tests to access private fields/methods of java.base/java.util such as ArrayList via reflection -->
-        <argLine>-Xmx512m --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED</argLine>
+<!--
+        <argLine>-Xmx512m &#45;&#45;add-opens java.base/java.lang.reflect=ALL-UNNAMED &#45;&#45;add-opens java.base/java.lang=ALL-UNNAMED &#45;&#45;add-opens java.base/java.util=ALL-UNNAMED</argLine>
+-->
+        <argLine>-Xmx512m</argLine>
       </properties>
     </profile>
     <profile>
--- a/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
@@ -19,6 +19,7 @@ package org.apache.commons.lang3.builder;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.time.temporal.Temporal;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Objects;
@@ -114,6 +115,11 @@ public class CompareToBuilder implements Builder<Integer> {
         final CompareToBuilder builder,
         final boolean useTransients,
         final String[] excludeFields) {
+
+        if ((lhs instanceof CharSequence || lhs instanceof Number || lhs instanceof Temporal) && lhs instanceof Comparable) {
+            builder.append(lhs, rhs);
+            return;
+        }
 
         final Field[] fields = clazz.getDeclaredFields();
         AccessibleObject.setAccessible(fields, true);
--- a/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
@@ -19,6 +19,7 @@ package org.apache.commons.lang3.builder;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.time.temporal.Temporal;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -731,7 +732,8 @@ public class EqualsBuilder implements Builder<Boolean> {
             // to be inlined
             appendArray(lhs, rhs);
         } else // The simple case, not an array, just test the element
-        if (testRecursive && !ClassUtils.isPrimitiveOrWrapper(lhsClass)) {
+        if (testRecursive && !ClassUtils.isPrimitiveOrWrapper(lhsClass)
+                && !(CharSequence.class.isAssignableFrom(lhsClass) || Number.class.isAssignableFrom(lhsClass) || Temporal.class.isAssignableFrom(lhsClass))) {
             reflectionAppend(lhs, rhs);
         } else {
             isEquals = lhs.equals(rhs);
@@ -956,7 +958,8 @@ public class EqualsBuilder implements Builder<Boolean> {
         }
 
         try {
-            if (testClass.isArray()) {
+            if (testClass.isArray() || Number.class.isAssignableFrom(testClass)
+                    || Temporal.class.isAssignableFrom(testClass) || CharSequence.class.isAssignableFrom(testClass)) {
                 append(lhs, rhs);
             } else //If either class is being excluded, call normal object equals method on lhsClass.
             if (bypassReflectionClasses != null
--- a/src/main/java/org/apache/commons/lang3/builder/HashCodeBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/HashCodeBuilder.java
@@ -20,6 +20,7 @@ package org.apache.commons.lang3.builder;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.time.temporal.Temporal;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -178,6 +179,10 @@ public class HashCodeBuilder implements Builder<Integer> {
     private static void reflectionAppend(final Object object, final Class<?> clazz, final HashCodeBuilder builder, final boolean useTransients,
             final String[] excludeFields) {
         if (isRegistered(object)) {
+            return;
+        }
+        if (object instanceof CharSequence || object instanceof Number || object instanceof Temporal) {
+            builder.append(object);
             return;
         }
         try {
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
@@ -20,9 +20,11 @@ package org.apache.commons.lang3.builder;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.time.temporal.TemporalAccessor;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Objects;
 
 import org.apache.commons.lang3.ArraySorter;
@@ -847,9 +849,14 @@ public class ReflectionToStringBuilder extends ToStringBuilder {
         }
 
         validate();
+        
+        if (handleNativeClasses()) {
+            return super.toString();
+        }
 
         Class<?> clazz = getObject().getClass();
         appendFieldsIn(clazz);
+        
         while (clazz.getSuperclass() != null && clazz != getUpToClass()) {
             clazz = clazz.getSuperclass();
             appendFieldsIn(clazz);
@@ -865,6 +872,23 @@ public class ReflectionToStringBuilder extends ToStringBuilder {
             ToStringStyle.unregister(getObject());
             throw new IllegalStateException("includeFieldNames and excludeFieldNames must not intersect");
         }
+    }
+
+    private boolean handleNativeClasses() {
+        Object value = getObject();
+        if (value instanceof Number || value instanceof Boolean || value instanceof Character
+            || value instanceof TemporalAccessor ) {
+            getStringBuffer().append("value=").append(getObject().toString());
+            return true;
+        }
+        
+        if (value.getClass().isArray() || value instanceof Collection || value instanceof Map) {
+            ToStringStyle.unregister(value);
+            getStyle().append(getStringBuffer(), null, value, true);
+            return true;
+        }
+        
+        return false;
     }
 
 }
--- a/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
+++ b/src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
@@ -992,17 +992,14 @@ public class ToStringBuilderTest extends AbstractLangTest {
         // LANG-1337 without this, the generated string can differ depending on the JVM version/vendor
         final List<Object> list = new ArrayList<>(ARRAYLIST_INITIAL_CAPACITY);
         final String baseString = toBaseString(list);
-        final String expectedWithTransients = baseString
-                + "[elementData={<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>},size=0,modCount=0]";
+        final String expected = baseString
+                + "[[]]";
         final String toStringWithTransients = ToStringBuilder.reflectionToString(list, null, true);
-        if (!expectedWithTransients.equals(toStringWithTransients)) {
-            assertEquals(expectedWithTransients, toStringWithTransients);
-        }
-        final String expectedWithoutTransients = baseString + "[size=0]";
+        assertEquals(expected, toStringWithTransients);
+
+        // no difference as Collections and Maps are not handled via reflection anymore
         final String toStringWithoutTransients = ToStringBuilder.reflectionToString(list, null, false);
-        if (!expectedWithoutTransients.equals(toStringWithoutTransients)) {
-            assertEquals(expectedWithoutTransients, toStringWithoutTransients);
-        }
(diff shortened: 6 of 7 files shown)


src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
```diff
@@ -115,6 +116,11 @@ private static void reflectionAppend(
         final boolean useTransients,
         final String[] excludeFields) {
 
+        if ((lhs instanceof CharSequence || lhs instanceof Number || lhs instanceof Temporal) && lhs instanceof Comparable) {
+            builder.append(lhs, rhs);
+            return;
+        }
+
```
garydgregory, line 123: Extra blank lines not needed IMO.

src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
```diff
@@ -115,6 +116,11 @@ private static void reflectionAppend(
         final boolean useTransients,
         final String[] excludeFields) {
 
+        if ((lhs instanceof CharSequence || lhs instanceof Number || lhs instanceof Temporal) && lhs instanceof Comparable) {
```
garydgregory, line 119: Needs a // comment IMO

src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
```diff
@@ -731,7 +732,8 @@ public EqualsBuilder append(final Object lhs, final Object rhs) {
             // to be inlined
             appendArray(lhs, rhs);
         } else // The simple case, not an array, just test the element
-        if (testRecursive && !ClassUtils.isPrimitiveOrWrapper(lhsClass)) {
+        if (testRecursive && !ClassUtils.isPrimitiveOrWrapper(lhsClass)
```
garydgregory, line 735: Needs a // comment IMO

src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
```diff
@@ -956,7 +958,8 @@ public EqualsBuilder reflectionAppend(final Object lhs, final Object rhs) {
         }
 
         try {
-            if (testClass.isArray()) {
+            if (testClass.isArray() || Number.class.isAssignableFrom(testClass)
```
garydgregory, line 961: Needs a // comment IMO

src/main/java/org/apache/commons/lang3/builder/HashCodeBuilder.java
```diff
@@ -180,6 +181,10 @@ private static void reflectionAppend(final Object object, final Class<?> clazz,
         if (isRegistered(object)) {
             return;
         }
+        if (object instanceof CharSequence || object instanceof Number || object instanceof Temporal) {
```
garydgregory, line 184: Needs a // comment IMO. Why not refactor the test exporession into a common (if possible package-private) method with the same code above?

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -867,4 +874,21 @@ private void validate() {
...
     }
 
+    private boolean handleNativeClasses() {
+        Object value = getObject();
+        if (value instanceof Number || value instanceof Boolean || value instanceof Character
+            || value instanceof TemporalAccessor ) {
+            getStringBuffer().append("value=").append(getObject().toString());
+            return true;
+        }
+
```
garydgregory: Extra blank line not needed IMO. Want I want call attention to something, I write a // comment.

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -867,4 +874,21 @@ private void validate() {
...
 
+    private boolean handleNativeClasses() {
+        Object value = getObject();
+        if (value instanceof Number || value instanceof Boolean || value instanceof Character
+            || value instanceof TemporalAccessor ) {
+            getStringBuffer().append("value=").append(getObject().toString());
+            return true;
+        }
+        
+        if (value.getClass().isArray() || value instanceof Collection || value instanceof Map) {
```
garydgregory, line 885: Can `value` be null here?

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -867,4 +874,21 @@ private void validate() {
...
+            ToStringStyle.unregister(value);
+            getStyle().append(getStringBuffer(), null, value, true);
+            return true;
+        }
+
```
garydgregory: Extra empty line not needed IMO.  Want I want call attention to something, I write a // comment.

garydgregory (changes requested):
Hi @struberg 
I have a few comments scattered throughout.

pom.xml
```diff
@@ -202,6 +202,7 @@
               <exclude>src/site/resources/download_lang.cgi</exclude>
               <exclude>src/site/resources/release-notes/RELEASE-NOTES-*.txt</exclude>
               <exclude>src/test/resources/lang-708-input.txt</exclude>
+              <exclude>nb-configuration.xml</exclude>
```
ppkarwasz, line 205: This is not necessary, since RAT has a implicit exclusion list populated from the `.gitignore` file.

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -867,4 +874,21 @@ private void validate() {
         }
     }
 
+    private boolean handleNativeClasses() {
+        Object value = getObject();
+        if (value instanceof Number || value instanceof Boolean || value instanceof Character
+                || value instanceof TemporalAccessor) {
```
garydgregory, line 880: This expression uses `TemporalAccessor` here but the new `isJavaInternalClass()` uses `Temporal`. `Temporal` extends `TemporalAccessor`, why is there a difference? It seems this expression could be rewritten to call  `isJavaInternalClass()`, this centralize the types to check.

garydgregory (changes requested):
Hi @struberg 

See my comments and @ppkarwasz's.

src/main/java/org/apache/commons/lang3/builder/CompareToBuilder.java
struberg: resolved

src/main/java/org/apache/commons/lang3/builder/EqualsBuilder.java
struberg: is now in the JavaDocs of Reflection#isJavaInternalClass(). That should resolve this and all similar situations.

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
```diff
@@ -867,4 +874,21 @@ private void validate() {
         }
     }
 
+    private boolean handleNativeClasses() {
+        Object value = getObject();
+        if (value instanceof Number || value instanceof Boolean || value instanceof Character
+            || value instanceof TemporalAccessor ) {
```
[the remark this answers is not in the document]
struberg: good catch, moved all to TomporalAccessor in Reflection.isJavaInternalClass()

src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
struberg: nope, that is covered in line 846 of toString()

pom.xml
struberg: that file is generated by NetBeans when using it. So no, does not have an AL header.
I've added the file to .gitignore but checkstyle did still complain.

ppkarwasz: Did Checkstyle or Apache RAT fail? Apache RAT should be OK without this line, Checkstyle does not read the configuration of Apache RAT AFAIK.

struberg: I'll try again without

struberg: yup, you're right it seems. Works now.

garydgregory (changes requested):
Hi @struberg 

We cannot test that an object implements a stock interface like `TemporalAccessor` or `CharSequence` because this won't work with custom implementations (`TemporalAmount` is also not accounted for, and others I am sure).

See the new test I added in git master: `EqualsBuilderReflectJreImplementationTest`

I think we can only test for an actual class or the the package that class lives in. 

TY

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
```diff
@@ -229,21 +230,33 @@ public void assertReflectionArray(final String expected, final Object actual) {
...
 
+    private static class HirAFixture extends HirBFixture {
+        int x = 1;
+    }
+
+    private static class HirBFixture extends HirCFixture {
+        int y = 2;
+    }
+
+    private static class HirCFixture {
```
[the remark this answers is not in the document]
struberg: Hierarchic

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
```diff
@@ -229,21 +230,33 @@ public void assertReflectionArray(final String expected, final Object actual) {
...
+
     /**
      * Tests ReflectionToStringBuilder setUpToClass().
      */
     @Test
     public void test_setUpToClass_valid() {
-        final Integer val = Integer.valueOf(5);
+        final HirAFixture val = new HirAFixture();
         final ReflectionToStringBuilder test = new ReflectionToStringBuilder(val);
-        test.setUpToClass(Number.class);
```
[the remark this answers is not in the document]
struberg: In this very case this test simply makes no sense. setUpToClass is INCLUDING the very class given. But if you introspect into any java.lang.Number, then it will just blow up your whole application from Java9 onwards...

src/test/java/org/apache/commons/lang3/builder/ToStringBuilderTest.java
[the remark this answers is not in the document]
struberg: Regarding `TemporalAccessor` and `CharSequence`: happy to replace CharSequence with a String.class check. Not quite sure about TemporalAccessor, no strong feeling, we could also move to `Temporal` if you want.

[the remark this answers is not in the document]
garydgregory: That's not what I'm saying ;-) What I'm trying to say is that the check _cannot_ use _any_ interfaces.

garydgregory (dismissed):
Hello @struberg 

Checking against an interface in `isJavaInternalClass()` is not going to work. 

I updated `EqualsBuilderReflectJreImplementationTest` in git master.

struberg:
> Hello @struberg
> 
> Checking against an interface in `isJavaInternalClass()` is not going to work.
> 
> I updated `EqualsBuilderReflectJreImplementationTest` in git master.

Gary, this has already long been resolved in https://github.com/apache/commons-lang/pull/1328/commits/f20639a039476826b9c146d96f64e6979ea7d772 and https://github.com/apache/commons-lang/pull/1328/commits/ec4150bc13cfd2705e5d672f47706930af578c39

I've even merged your changes (and removed that new --add-opens again...) from main to this very branch.

garydgregory:
> > Hello @struberg
> > Checking against an interface in `isJavaInternalClass()` is not going to work.
> > I updated `EqualsBuilderReflectJreImplementationTest` in git master.
> 
> Gary, this has already long been resolved in [f20639a](https://github.com/apache/commons-lang/commit/f20639a039476826b9c146d96f64e6979ea7d772) and [ec4150b](https://github.com/apache/commons-lang/commit/ec4150bc13cfd2705e5d672f47706930af578c39)
> 
> I've even merged your changes (and removed that new --add-opens again...) from main to this very branch.
> 
> This branch clearly improves the situation with standard jpms setup for a lot of common cases. Of course, this does not yet fix ALL the problems (albeit it's a good start). But if people come across another `InaccessibleObjectException` then they can still --add-opens as needed. So the current version is clearly an improvement. If you agree, then we should merge it into main instead of yet more parallel work in different branches...

Hello @struberg 

The issue is not that this PR is a wants to start improving something, the issue is that this PR _breaks_ existing behavior.

You cannot check for interfaces to short-circuit logic in `Reflection.isJavaInternalClass()`. This will break existing applications.

The branch is not up to date with master. This fails:
```
git clone https://github.com/struberg/commons-lang.git 
cd commons-lang
git checkout fb_jpms
git remote add upstream https://github.com/apache/commons-lang.git
git pull upstream master
mvn clean test -Dtest=EqualsBuilderReflectJreImplementationTest
```

struberg:
> You cannot check for interfaces to short-circuit logic in `Reflection.isJavaInternalClass()`. This will break existing applications.

Which of those is an interface?
https://github.com/struberg/commons-lang/blob/ec4150bc13cfd2705e5d672f47706930af578c39/src/main/java/org/apache/commons/lang3/builder/Reflection.java#L62

The thing you are talking about IS ALREADY RESOLVED since yesterday.

And one other thing: commons-lang3 ReflectionToString ReflectionEquals etc was BROKEN on anything Java9++ - period!
The build is ONLY green because you just disabled all the modularity - which is something not everyone can do in real projects!

My spare time to work on this stuff is very limited - as is yours I bet. So please use our both time productively. As I said: not everything will fully work from the beginning. You just added `java.time.Period`, fine, good catch. But why on earth don't we just take that PR and commit it and then improve it TOGETHER? I'm not willing to do more empty rounds getting sticks thrown at me at will. I really appreciate the review feedback as it indeed did highly improve the code. But now is the point to decide whether the direction is right or not - and then improve things from here TOGETHER.

struberg:
Btw, I'm looking for another method name for `isJavaInternalClass()`. It should rather express that the given clazz cannot be introspected via reflection. Maybe something like `isNonIntrospectibleClass()`?

Further down the road we need some way to tell whether a class is open or not. And all that in a way which also works on Java8 still...

struberg:
> I guess you could be bothered to look it up

Indeed, had in the back of my mind that Temporal is an abstract class.
Latest version uses reflection to figure that a class is not accessible. This works, but maybe we should try a mixture of the old approach + the new one?

Thought about having a static `Map<Class, Boolean> KNOWN_INACCESSIBLE_CLASSES` which contains the concrete classes like LocalDate, Boolean, Character, Integer, etc.
And then in a second step we do the reflection check. But as this might also have unexpected results (people might forget that they closed the jars) we should probably log it out?

src/main/java/org/apache/commons/lang3/builder/Reflection.java
```diff
@@ -18,13 +18,62 @@
...
+                Character.class,
+                String.class,
+                Byte.class,
+                Short.class,
+                Integer.class,
+                Long.class,
+                Float.class,
+                Double.class,
+                java.util.Date.class,
+                java.sql.Date.class,
```
garydgregory, line 58: This is not specific to this PR: I don't think we should introduce a _new JPMS module dependency_, here on `java.sql`. Lang does not use any classes from `java.sql` so this would drag in a new JPMS module. For 4.0, we've mentioned dropping our dependency on `java.desktop` and only depending on `java.base`.

src/main/java/org/apache/commons/lang3/builder/Reflection.java
```diff
@@ -41,4 +90,44 @@ static Object getUnchecked(final Field field, final Object obj) {
...
+        if (isAccessible != null) {
+            return isAccessible;
+        }
+        try {
+            final Field[] declaredFields = clazz.getDeclaredFields();
+            for (Field f : declaredFields) {
+                f.setAccessible(true);
+            }
+
+            nonIntrospectibleClasses.put(clazz, false);
```
garydgregory, line 125: This looks problematic since the map will grow uncontrollably, possibly until every class on your classpath gets referenced, there's no way to know really.

src/main/java/org/apache/commons/lang3/builder/Reflection.java
```diff
@@ -41,4 +90,44 @@ static Object getUnchecked(final Field field, final Object obj) {
...
+     * But it's the best we can do.
+     *
+     * @return {@code true} if the given class internas not accessible
+     */
+    static boolean isNonIntrospectibleClass(Object o) {
+        return o != null && isNonIntrospectibleClass(o.getClass());
+    }
+
+    /**
+     * Check whether the given class internas can be accessed. This might return false in Java9
```
garydgregory, line 105: "internas" spelling

garydgregory (changes requested):
Hi All,

So far, I'm still against this change.

These features are _designed to use reflection_, and this PR tries to turn some of that off. 

It seems to me like it will cause unintended consequences beyond all of the ones I pointed out through the new unit tests I wrote which showed the initial implementation would break existing apps. 

Now, we have a different implementation, which brings in a new set of problems: 
- It depends on a new JPMS module `java.sql`
- it grows a map without any bounds (a CVE waiting to happen IMO)
- you can't turn it off
- The behavior is completely hidden from the user

ppkarwasz:
> So far, I'm still against this change.
> 
> These features are _designed to use reflection_, and this PR tries to turn some of that off.

I agree. All the `reflectionX` methods have a warning:

> Fields can be private, thus `AccessibleObject.setAccessible` is used to bypass normal access control checks. This will fail under a security manager unless the appropriate permissions are set.

The warning might require an update, but it seems clear to me that it also applies to the JPMS access checks. So it is up to caller to ensure that the caller to ensure that the class is open to `org.apache.commons.lang3`. If the caller uses `testRecursive`, he must also ensure that the appropriate permissions are set for all the field types.

garydgregory:
This discussion is mirrored on the apache list here: 
https://lists.apache.org/thread/gwljc3ylz3cjvbg9m7sv9dtj139obprz

garydgregory (dismissed):
Note that this PR is worse than I thought because we already have a feature to bypass reflection! Use `EqualsBuilder.setBypassReflectionClasses(List<Class<?>>)`.

struberg:
further comments over at the commons dev list.

src/main/java/org/apache/commons/lang3/builder/Reflection.java
struberg: yup true

src/main/java/org/apache/commons/lang3/builder/Reflection.java
struberg: this is a trade-off vs performance. it will only cache those classes which got hit in a reflection equals, toString, etc. Thus in practice it is a very limited set of classes I'd say. Can you think of a possible place where we can hold such a cache at least during a single invocation? Maybe inside the builder and handing the Map as another parameter to this method?

struberg:
> I agree. All the `reflectionX` methods have a warning:

Well this is way more subtly weird right now.

Imagine the following code:
```
        Map<String, String> a = new HashMap<>(20);
        a.put("A", "A");

        Map<String, String> b = new HashMap<>();
        b.put("A", "A");
```
Here we have the case that `a.equals(b)` returns `true`. Because that's the contract of Maps!
But when you use `new EqualsBuilder().setTestRecursive(true).append(a, b).isEquals()` then you'll get `false`.
That's because by giving a different initial array size, we'll get a different *internal* (!) structure (bigger internal arrays, etc). 

But that means that right now the EqualsBuilder (even in Java8) does not even adheres to the contract of the Map interface. Because those internal details should not matter. Even if I would not have set an initial size - what if I add 30 elements to one of the 2 maps, but then remove everything but the "A" entry again? It will be equals according to the Map contract, but not for the EqualsBuilder. Looks rather broken to me.

One might argue that that's exactly how all the reflection stuff in commons-lang was designed from the beginning.
Otoh there are already a few places where Maps, Collections, etc get a special treatment. So why not also here?

struberg:
> Note that this PR is worse than I thought because we already have a feature to bypass reflection! Use `EqualsBuilder.setBypassReflectionClasses(List<Class<?>>)`.

This is just plain wrong. You cannot use `setBypassReflectionClasses` for e.g. Map without loosing the reflexive comparison of it's entries.

PR #1332, opened 2024-12-09
MINOR: Changed to simpler + concat as it's more readable and StringBuilder d…

Serwios:
…oesn't make a lot of sense here

Here StringBuilder doesn't giving significant benefit. The generated bytecode typically uses a StringBuilder for such concatenations automatically so for better readability I changed for simple + concat.

Didn't add tests cause not needed

### DIFF
--- a/src/main/java/org/apache/commons/lang3/arch/Processor.java
+++ b/src/main/java/org/apache/commons/lang3/arch/Processor.java
@@ -244,9 +244,6 @@ public class Processor {
 
     @Override
     public String toString() {
-        final StringBuilder builder = new StringBuilder();
-        builder.append(type.getLabel()).append(' ').append(arch.getLabel());
-        return builder.toString();
+        return type.getLabel() + ' ' + arch.getLabel();
     }
-
 }


garydgregory:
-1 @Serwios 

This is not a good idea. You've removed the guarantee `StringBuilder` provides of not using synchronized code, which is unnecessary here.

Serwios:
Hi, in JDK 8 and earlier, + compiles to StringBuilder. From JDK 9 onward, it uses StringConcatFactory, which optimizes simple cases like this to behave similarly to StringBuilder.

This guarantee should be present as return type.getLabel() + ' ' + arch.getLabel(); is compiled to return new StringBuilder().append(type.getLabel()).append(' ').append(arch.getLabel()).toString();

https://stackoverflow.com/questions/11942368/why-use-stringbuilder-explicitly-if-the-compiler-converts-string-concatenation-t

ppkarwasz:
@Serwios,

String concatenation is IMHO more readable than the equivalent `StringBuilder` construction, but changing this in a single method that is probably not used and certainly not tested, doesn't make sense to me.

Could you go the extra mile of:

- proposing on the `dev@commons.apache.org` mailing list to replace simple `StringBuilder` constructs with string concatenation. Whether to use `StringBuilder`, string concatenation or `String.format()` is mostly a matter of taste and IDEs can refactor each one of these formats into another.
- if the community accepts the change, apply it **everywhere**?

Serwios:
> @Serwios,
> 
> String concatenation is IMHO more readable than the equivalent `StringBuilder` construction, but changing this in a single method that is probably not used and certainly not tested, doesn't make sense to me.
> 
> Could you go the extra mile of:
> 
> * proposing on the `dev@commons.apache.org` mailing list to replace simple `StringBuilder` constructs with string concatenation. Whether to use `StringBuilder`, string concatenation or `String.format()` is mostly a matter of taste and IDEs can refactor each one of these formats into another.
> * if the community accepts the change, apply it **everywhere**?

I’ll propose this discussion on the dev@commons.apache.org mailing list to determine if there’s a preference within the team for standardizing string concatenation over StringBuilder or other alternatives like String.format(). If the community supports this, I’m happy to take the lead on refactoring similar cases project-wide to ensure consistency.

IMHO: Regarding standardization, I believe it’s about establishing a common pattern that promotes code readability and maintainability. Having a generally accepted approach ensures that contributors don’t waste time debating minor stylistic differences and instead focus on improving functionality. It also makes the codebase easier to review, debug, and extend, especially for new contributors.

garydgregory:
I like @ppkarwasz 's idea. I don't want to review one-offs.

PR #1333, opened 2024-12-09
Aggregated exceptions to one factory class

Serwios:
Created factory to throw common math exception as there was too much message duplication, simplified already created messages. (Maybe make sense to create custom math exception to separate it from type validations, and nan checks etc)

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -19,6 +19,8 @@ package org.apache.commons.lang3.math;
 import java.math.BigInteger;
 import java.util.Objects;
 
+import static org.apache.commons.lang3.math.MathExceptionFactory.*;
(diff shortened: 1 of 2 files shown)


garydgregory:
I don't see how this helps but that's just me.

I've experimented with this type of refactoring in the past and I've come to the conclusion that one
should leave the 'throw' statements where they are, this makes reading control flow easier, allows some static analysis tool, and IDEs to be less confused.

Serwios:
Math exceptions are pretty general and are duplicated everywhere. By aggregating them into one generic utill class we will have less code duplication and generified response for client.

garydgregory:
Refactoring in the "throw" statement part is still a bad idea IMO. Having an exception factory is a separate issue.

Serwios:
I understand you for already existed clients new exception types can be a problem and regarding factory for now I think it makes sense

PR #1337, opened 2024-12-13
Added new methods for Strings util

Serwios:
Added the following new useful utility methods to Strings:

removeAccents: Removes accents (diacritical marks) from a string.
maskString: Masks a part of a string with a given character.
containsOnly: Checks if a string contains only characters from a specified set.
reverseWords: Reverses the order of words in a string, preserving spaces.
truncateWithEllipsis: Truncates a string and appends ellipsis if truncated.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Strings.java
+++ b/src/main/java/org/apache/commons/lang3/Strings.java
@@ -22,8 +22,13 @@ import static org.apache.commons.lang3.StringUtils.INDEX_NOT_FOUND;
 import org.apache.commons.lang3.builder.AbstractSupplier;
 import org.apache.commons.lang3.function.ToBooleanBiFunction;
 
+import java.text.Normalizer;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.stream.Collectors;
+
 /**
- * String operations where you choose case-senstive {@link #CS} vs. case-insensitive {@link #CI} through a singleton instance.
+ * String operations where you choose case-sensitive {@link #CS} vs. case-insensitive {@link #CI} through a singleton instance.
  *
  * @see CharSequenceUtils
  * @see StringUtils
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @Serwios 

-1:
- You checked the box "Run a successful build using the default [Maven](https://maven.apache.org/) goal with mvn; that's mvn on the command line by itself." but you obviously didn't do it since the build fails for me locally:
```
[ERROR] Failures:
[ERROR]   StringsTest.testRemoveAccents:141 expected: <Cafe> but was: <Caf?>
```
Checkstyle also fails:
```
[ERROR] src\main\java\org\apache\commons\lang3\Strings.java:[25,1] (imports) ImportOrder: Wrong order for 'java.text.Normalizer' import.
[ERROR] src\main\java\org\apache\commons\lang3\Strings.java:[28,8] (imports) UnusedImports: Unused import - java.util.stream.Collectors.
[ERROR] src\test\java\org\apache\commons\lang3\StringsTest.java:[26,1] (imports) ImportOrder: Wrong order for 'org.junit.jupiter.api.Assertions.*' import.
[ERROR] src\test\java\org\apache\commons\lang3\StringsTest.java:[26,47] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - org.junit.jupiter.api.Assertions.*.
```
We've deprecated text processing functionality from Lang and moved it to Commons Text. For example, the whole `org.apache.commons.lang3.text` package is deprecated.

- `truncateWithEllipsis()`: See the general note on text processing above. I'm not sure the function is useful as is anyway. Unicode supports the horizontal ellipsis character but this function doesn't use it? Why three periods instead? Should it be configurable, and so on.
- `reverseWords`: See the general note on text processing above.
- `containsOnly`: How is this different from StringUtils.indexOfAnyBut()?
- `maskString`: See the general note on text processing above. In general, there are probably too many variations or use cases. Masking is usually much more complex, like "Replace phone numbers, credit cards, and so on, with X"
- `removeAccents`: See the general note on text processing above.

PR #1343, opened 2025-01-24
Update StringUtils.java

sirajalam049:
The current trim function does not trim down the non-breaking spaces, zero-width spaces, or other invisible characters. Added logic for that too.

- [ x Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8761,7 +8761,8 @@ public class StringUtils {
     }
 
     /**
-     * Removes control characters (char &lt;= 32) from both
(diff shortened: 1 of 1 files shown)


garydgregory:
-1: The point of the method is to provide a null-safe version of trim(), not to replace chars inside the string,  that's completely different. Even worse, this PR has ZERO tests.

ppkarwasz:
@sirajalam049,

Thank you for the proposal, but I am rather `-1`: this would be a **breaking** change for existent code.
However, you can propose a new method that eliminate invisible characters. In such a case, please provide some use cases for it.

garydgregory:
Closing: No feedback from OP.

PR #1355, opened 2025-03-14
deprecated ObjectUtils `defaultIfNull` add `getIfNull` and `getFirstNonNull`

pankratz76:


pankratz76:
is this any good?

I have adapted this change in spring as well, so I had the code quite ready and wanted to thank and back contribute this. 

https://github.com/spring-projects/spring-framework/pull/34587

brettkail-wk:
Out of curiosity, why is this rename/deprecation useful?  Subjectively, `defaultIfNull` seems clearer than `getIfNull`, particularly when statically imported.

Pankraz76:
The TODO directive given, it's considered deprecated to avoid breaking change.

I like both approaches, but the default context is somewhat implied. While having a verbose API is nice for beginners, for those used to it, this feels like overhead.

Since it's either/or, why do we need to explicitly mention the default? Asking this question could argue why not renaming to `returnEtherValueOrDefaultValueIfNull` this is a very verbose version giving good testament on naming.

considering:

**big scope -> small name**

**small scope ->  big name**

src/main/java/org/apache/commons/lang3/ObjectUtils.java
```diff
@@ -580,7 +580,7 @@ public static short CONST_SHORT(final int v) {
      * @param object  the {@link Object} to test, may be {@code null}
      * @param defaultValue  the default value to return, may be {@code null}
      * @return {@code object} if it is not {@code null}, defaultValue otherwise
-     * TODO Rename to getIfNull in 4.0
```
Pankraz76, line 583: @brettkail-wk 

kindly consider TODO.

brettkail-wk: I'm not familiar with the history of the TODO.

That said, I have a project with hundreds/thousands of `defaultIfNull`.  It's unclear that `getIfNull` is better than `defaultIfNull`, or at least it's unclear that it's enough better to deprecate `defaultIfNull`.

Pankraz76: > enough better

not saying anything. Just assuming rename implies a breaking change.

Not consider it to be happen anyways.

Thanks for the interaction.

PR #1362, opened 2025-03-24
Consolidating duplicated add logic

shraymoza:
By consolidating the duplicated logic into a generic method, the code becomes more DRY (Don’t Repeat Yourself) and easier to maintain. It also minimizes the risk of errors when similar changes need to be applied to all versions.


sgflt:
Java != C++.

This change introduces memory allocations caused by boxing of primitives and I think it is the cause, why code is not intentionally DRY.

PR #1373, opened 2025-04-13
added method TruncateToLength and a test case for the same

ajitpattar708:
Added `truncateToLength`, a null-safe method to truncate strings, returning "" for null.


garydgregory:
-1 for me: this is almost the same as https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#substring(int,int)

What do others think?

ppkarwasz:
> What do others think?

I was thinking about `substring` too.

garydgregory:
Closing, no more feedback, the consensus is against adding this.

PR #1381, opened 2025-05-10
[LANG-1773] Apache Commons Lang no longer builds on Android

eamonnmcmanus:
Remove an unnecessary reference to `java.lang.reflect.AnnotatedType`, which does not exist on Android. We immediately call `getType()` on it, so we might just as well use the method that gives us `Type[]` directly.


garydgregory:
TY @eamonnmcmanus 

The PR is now in git master. Please verify a local clone of git master or a snapshot build from https://repository.apache.org/content/repositories/snapshots/

PR #1382, opened 2025-05-15
Add extractMatches() to RegExUtils

komatzu1337:
Provides simple way to get all regex matches as list without manual Matcher handling.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RegExUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RegExUtils.java
@@ -757,62 +757,57 @@ public class RegExUtils {
     }
 
 
-    public class MatchFinder {
-
-        /**
-         * Finds all matches in the given text according to the specified pattern.
-         *
-         * @param text    the text to search for matches
-         * @param pattern the pattern to search for
-         * @return a list of found matches
-         * @throws IllegalArgumentException if text or pattern is null
-         */
-        public List<String> findMatches(CharSequence text, Pattern pattern) {
-            if (text == null)
-                throw new IllegalArgumentException("Text must not be null");
-
-            if (pattern == null)
-                throw new IllegalArgumentException("Pattern must not be null");
-
-            List<String> matches = new ArrayList<>();
-            Matcher matcher = pattern.matcher(text);
-
-            while (matcher.find()) {
-                matches.add(matcher.group());
-            }
-
-            return Collections.unmodifiableList(matches); // return an unmodifiable list
+    /**
+     * Finds all matches in the given text according to the specified pattern.
+     *
+     * @param text    the text to search for matches
+     * @param pattern the pattern to search for
+     * @return a list of found matches; returns an empty List if text or pattern is null
+     */
+    public List<String> findMatches(CharSequence text, Pattern pattern) {
+        if (text == null || pattern == null){
+            return Collections.emptyList();
         }
 
-        /**
-         * Finds all matches in the given text according to the specified pattern.
-         *
-         * @param text    the text to search for matches
-         * @param pattern the pattern to search for
-         * @return a list of found matches
-         * @throws IllegalArgumentException if text or pattern is null
-         */
-        public List<String> findMatches(String text, Pattern pattern) {
-            return findMatches((CharSequence) text, pattern);
+        List<String> matches = new ArrayList<>();
+        Matcher matcher = pattern.matcher(text);
+
+        while (matcher.find()) {
+            matches.add(matcher.group());
         }
 
-        /**
-         * Finds all matches in the given text according to the specified pattern
-         * and returns them as an array of strings.
-         *
-         * @param text  the text to search for matches (throws IllegalArgumentException if null)
-         * @param regex the regular expression pattern (throws IllegalArgumentException if null or invalid)
-         * @return an array of found matches (never null)
-         * @throws IllegalArgumentException if text or regex is null
-         * @throws PatternSyntaxException   if the regex syntax is invalid
-         */
-        public String[] findMatchesAsArray(CharSequence text, String regex) {
-            if (regex == null) {
-                throw new IllegalArgumentException("Regex must not be null");
-            }
+        return Collections.unmodifiableList(matches);
+    }
+
+    /**
+     * Finds all matches in the given text according to the specified pattern.
+     *
+     * @param text    the text to search for matches
+     * @param pattern the pattern to search for
+     * @return a list of found matches
+     */
+    public List<String> findMatches(String text, Pattern pattern) {
+        return findMatches((CharSequence) text, pattern);
+    }
+
+    /**
+     * Finds all matches in the given text according to the specified pattern
+     * and returns them as an array of strings.
+     *
+     * @param text  the text to search for matches
+     * @param regex the regular expression pattern
+     * @return an array of found matches; returns an empty array if text or regex is null
+     */
+    public String[] findMatchesAsArray(CharSequence text, String regex) {
+        if (text == null ||regex == null){
+            return new String[0];
+        }
+        try {
             Pattern pattern = Pattern.compile(regex);
             List<String> matches = findMatches(text, pattern);
             return matches.toArray(new String[0]);
+        }catch (PatternSyntaxException e){
+            return new String[0];
         }
     }
 }
--- a/src/main/java/org/apache/commons/lang3/RegExUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RegExUtils.java
@@ -17,10 +17,12 @@
 package org.apache.commons.lang3;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 /**
  * Helpers to process Strings using regular expressions.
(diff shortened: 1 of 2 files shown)


garydgregory (commented):
I'm not sure how this one. This is a pretty narrow use case. Some people will want a List, others an array; some will want String input, others will say it should be a CharSequence. What should happen if the string is null? What should happen if the pattern is null? Should an empty list (or array) be returned? A null? Should the List be immutable? Can this even be reused with Lang?

komatzu1337:
TY for feedback! I implemented the following improvements:

Now throws IllegalArgumentException with clear messages for null 
Supports both String and CharSequence 
Provides both List<String> and String[] return 
Added regex-string variant 
All returned collections are unmodifiable
Never returns null (empty collections for null) 
Improved documentation

I will add tests later if you find this class worthwhile

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,64 @@ private static String toStringOrNull(final CharSequence text) {
     public RegExUtils() {
         // empty
     }
+
+
+    public class MatchFinder {
```
ppkarwasz, line 760: There is no need for a nested class.

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,64 @@ private static String toStringOrNull(final CharSequence text) {
...
+         * @param pattern the pattern to search for
+         * @return a list of found matches
+         * @throws IllegalArgumentException if text or pattern is null
+         */
+        public List<String> findMatches(CharSequence text, Pattern pattern) {
+            if (text == null)
+                throw new IllegalArgumentException("Text must not be null");
+
+            if (pattern == null)
+                throw new IllegalArgumentException("Pattern must not be null");
```
ppkarwasz, line 775: Since most of the methods in Commons Lang are `null`-safe, it would be more appropriate to return an empty list.

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,64 @@ private static String toStringOrNull(final CharSequence text) {
...
+         * Finds all matches in the given text according to the specified pattern.
+         *
+         * @param text    the text to search for matches
+         * @param pattern the pattern to search for
+         * @return a list of found matches
+         * @throws IllegalArgumentException if text or pattern is null
+         */
+        public List<String> findMatches(String text, Pattern pattern) {
+            return findMatches((CharSequence) text, pattern);
+        }
```
ppkarwasz, line 797: Since `String` is a `CharSequence`, this method seems useless.

ppkarwasz (changes requested):
Hi @komatzu1337,

Thanks for the PR, could you add some unit tests please?

Note also that you [enabled "Vigilant mode"](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits) on your GitHub account, but you didn't [add the GPG key you are using to your GitHub account](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account). For this reason all your commits appear as "Unverified". Can you fix that?

garydgregory:
I'm still not convinced this belongs here.

ppkarwasz:
> I'm still not convinced this belongs here.

@komatzu1337, I appreciate the effort you've put into this contribution. I also have some reservations about the utility of `findMatches`, so please view my comments as suggestions for improvement rather than a final decision on acceptance.

Just as a reminder, per ASF guidelines, all maintainers need to reach consensus before any changes are merged.

src/main/java/org/apache/commons/lang3/RegExUtils.java
komatzu1337: I didn't know that, your suggest seems right

komatzu1337:
```suggestion
         if (text == null || pattern == null){
            return Collections.emptyList();
        }
```

ppkarwasz: Fixed in https://github.com/search?q=findAllMatches+language%3AJava&type=code

ppkarwasz:
> > I'm still not convinced this belongs here.
> 
> why?

When adding a new method to Commons Lang, we need to consider the impact on the entire ecosystem. Even small additions increase the library's size for thousands of downstream projects. Given the wide range of potential utility methods, we prioritize those likely to benefit a significant portion of users—typically, at least 1%.

While your proposed method may be useful in some contexts, it might not offer enough broad utility to justify inclusion in the core library.

komatzu1337:
@ppkarwasz

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,67 @@ private static String toStringOrNull(final CharSequence text) {
...
     }
+
+    /**
+     * Finds all matches in the given text according to the specified pattern.
+     *
+     * @param text    the text to search for matches
+     * @param pattern the pattern to search for
+     * @return an unmodifiable list of found matches; returns an empty List if text or pattern is null
+     */
+    public List<String> findMatches(CharSequence text, Pattern pattern) {
```
ppkarwasz, line 766:
```suggestion
    public static List<String> findAll(CharSequence text, Pattern pattern) {
```

This is an utility class, so all the methods must be `static`.

_Nitpick_: Naming is certainly on of the hardest tasks. Since all the methods in this class deal with matching and have names like `replaceAll`, `removeAll`, I would call this method `findAll`.

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,67 @@ private static String toStringOrNull(final CharSequence text) {
...
+
+    /**
+     * Finds all matches in the given text according to the specified pattern
+     * and returns them as an array of strings.
+     *
+     * @param text  the text to search for matches
+     * @param regex the regular expression pattern
+     * @return an array of found matches; returns an empty array if text or regex is null
+     */
+    public String[] findMatchesAsArray(CharSequence text, String regex) {
```
ppkarwasz, line 789: Looking at the remaining methods of this class, each method with signature `(CharSequence, Pattern)` has a corresponding method with signature `(String, String)`. Therefore I would define a method:

```java
public static List<String> findAll(String text, String regex)
```

I don't think it is useful to have both a method returning `List<String>` and `String[]`. The return type should be coherent.

src/main/java/org/apache/commons/lang3/RegExUtils.java
```diff
@@ -751,4 +755,67 @@ private static String toStringOrNull(final CharSequence text) {
...
+
+    /**
+     * Finds all matches in the given text according to the specified pattern
+     * and returns them as a modifiable ArrayList.
+     *
+     * @param text    the text to search for matches
+     * @param pattern the pattern to search for
+     * @return a modifiable ArrayList of found matches; returns an empty List if text or pattern is null
+     */
+    public List<String> findMatchesModifiable(CharSequence text, Pattern pattern){
```
ppkarwasz, line 810: -1

We should not start multiplying methods.

src/test/java/org/apache/commons/lang3/RegExUtilsTest.java
```diff
@@ -379,4 +381,149 @@ public void testReplacePatternDeprecated() {
         assertEquals("Lorem_ipsum_dolor_sit", RegExUtils.replacePattern("Lorem ipsum  dolor   sit", "( +)([a-z]+)", "_$2"));
     }
 
+    private RegExUtils utils;
+
+    @BeforeEach
+    void setUp() {
+        utils = new RegExUtils();
+    }
```
ppkarwasz, line 389: Utility classes are never instantiated.

src/test/java/org/apache/commons/lang3/RegExUtilsTest.java
```diff
@@ -379,4 +381,149 @@ public void testReplacePatternDeprecated() {
...
+
+    @Test
+    public void testFindMultipleMatches() {
```
ppkarwasz, line 392: Can you use parameterized logging instead of multiple methods with the same logic?

ppkarwasz (changes requested):
@komatzu1337,

As noted earlier, we're still not convinced that even a single findAll method is broadly useful. Including three separate implementations feels like more than we can justify at this point. Could you please limit the pull request to the `findAll(CharSequence, Pattern)` method, and optionally a second variant that delegates to it?

garydgregory (commented):
There are so many possibilities of combining `Pattern`, `Matcher`, and `MatchResult` that I don't see why this _specific_ use case needs to be here. For example, where would this be reused in Commons Lang or from other Commons libraries?

ppkarwasz:
> There are so many possibilities of combining `Pattern`, `Matcher`, and `MatchResult` that I don't see why this _specific_ use case needs to be here. For example, where would this be reused in Commons Lang or from other Commons libraries?

Good point! @komatzu1337, could you share a few concrete examples where this method would be especially useful—either in real-world projects or potential usage within Commons libraries?

The current unit tests illustrate some interesting scenarios, but they can be addressed using existing utilities:

- Splitting a sentence into words can be done with `StringUtils.split`.
- To extract numbers from a sentence, you could split into words and then filter those that are numeric.

Understanding your intended use cases would help us evaluate whether this method brings enough added value and general applicability to warrant inclusion here.

komatzu1337:
@ppkarwasz @garydgregory Thank you for your patience and valuable feedback.After considering your feedback, i agree that this method doesn’t align well with Commons Lang’s scope. I apologize for the time spent reviewing this proposal - in hindsight, I should have more thoroughly vetted the use case before submitting it. 

This experience has helped me better understand Commons Lang's design philosophy. I will be sure to:
1. Review the contribution guidelines more carefully next time
2. Get feedback in advance through discussion questions for future changes.

PR #1385, opened 2025-05-17
Add Date to LocalDateTime conversion methods to bridge legacy and modern date APIs

yuyu1025:
#### **Background and Motivation**  
With the widespread adoption of Java 8+, the project has gradually moved from the traditional `java.util.Date` to modern date-time APIs (such as `LocalDateTime`), but a large amount of existing code still relies on `Date`. To reduce migration costs and improve development efficiency, it is necessary to provide officially supported conversion tools in Apache Commons Lang to avoid developers repeating the implementation of underlying logic.

#### **Core Change Content**  
1. **New Conversion Methods**  
   - `toLocalDateTime(Date date)`：Uses the system's default time zone, simplifying common conversion scenarios.  
   - `toLocalDateTime(Date date, TimeZone tz)`：Explicitly specifies the time zone, meeting internationalization requirements (e.g., converting UTC time to local time in a specific time zone).  
   - Implementation principle: Converts using `Instant` and `ZoneId`, ensuring time zone accuracy (including daylight saving time, handling of extreme time zones).  

2. **Improve test coverage**  
   - Covers scenarios such as default time zones, specified time zones, null value validation, daylight saving time (e.g., New York in the United States), extreme time zones (e.g., Kiribati with UTC+14) to ensure code robustness.  

#### **Value Added**  
- **Smooth Migration**: Allows developers to gradually introduce `LocalDateTime` without modifying the existing `Date` interface, reducing refactoring risks.  
- **Consistency and Reliability**: Provides Apache's official conversion logic, avoiding potential bugs in third-party implementations, and enhancing project stability.  
- **Feature Extension**: Enhance `DateUtils` modern support to adapt to Java 8+ ecosystem, responding to the community's need for bridging new and old APIs.  

#### **Compatibility Note**  
- No breaking changes, new methods are independent extensions, do not affect existing `DateUtils` functionality.  
- Depends on Java 8+ (consistent with Apache Commons Lang 3.18+).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -452,8 +452,9 @@
       </build>
     </profile>
 
+    <!-- Profile inherited from `commmons-parent` -->
     <profile>
-      <id>java9+</id>
+      <id>java-9-up</id>
       <activation>
         <jdk>[9,)</jdk>
       </activation>
@@ -462,6 +463,32 @@
         <!-- LANG-1667: allow tests to access private fields/methods of java.base/java.util such as ArrayList via reflection -->
         <argLine>-Xmx512m --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.time=ALL-UNNAMED --add-opens java.base/java.time.chrono=ALL-UNNAMED</argLine>
       </properties>
+      <build>
+        <plugins>
+          <!--
+            ~ Modifies the inherited Moditect configuration to add `java.sql` as `static` (optional) dependency.
+            -->
+          <plugin>
+            <groupId>org.moditect</groupId>
+            <artifactId>moditect-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>add-module-infos</id>
+                <configuration>
+                  <module>
+                    <moduleInfo>
+                      <requires>
+                        static java.sql;
+                        *;
+                      </requires>
+                    </moduleInfo>
+                  </module>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
     </profile>
     <profile>
       <id>java15</id>
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -1688,12 +1688,12 @@ public class DateUtils {
      *             it will return 0.
      * @return The nanosecond part of the {@link java.sql.Timestamp} object.
      */
-    private static int extractNanosFromSqlTimestamp(Date date){
-        if (timestampClass==null){
+    private static int extractNanosFromSqlTimestamp(Date date) {
+        if (timestampClass == null) {
             return 0;
         }
         try {
-            return (int)timestampGetNanosMethod.invoke(date);
+            return (int) timestampGetNanosMethod.invoke(date);
         } catch (IllegalAccessException | InvocationTargetException e) {
             return 0;
         }
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -18,6 +18,8 @@ package org.apache.commons.lang3.time;
 
 import java.text.ParseException;
 import java.text.ParsePosition;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Iterator;
@@ -1623,6 +1625,29 @@ public class DateUtils {
         final Calendar c = Calendar.getInstance(tz);
         c.setTime(Objects.requireNonNull(date, "date"));
         return c;
+    }
+
+    /**
+     * Converts a {@link Date} into a {@link LocalDateTime}, using the default time zone.
+     * @param date the date to convert to a LocalDateTime
+     * @return the created LocalDateTime
+     * @throws NullPointerException if {@code date} is null
+     * @since 3.18
+     */
+    public static LocalDateTime toLocalDateTime(final Date date) {
+        return toLocalDateTime(date, TimeZone.getDefault());
+    }
+
+    /**
+     * Converts a {@link Date} into a {@link LocalDateTime}
+     * @param date the date to convert to a LocalDateTime
+     * @param tz the time zone of the {@code date}
+     * @return the created LocalDateTime
+     * @throws NullPointerException if {@code date} is null
+     * @since 3.18
+     */
+    public static LocalDateTime toLocalDateTime(final Date date, final TimeZone tz) {
+        return Objects.requireNonNull(date, "date").toInstant().atZone(ZoneId.of(tz.getID())).toLocalDateTime();
     }
 
     /**
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -1650,7 +1650,7 @@ public class DateUtils {
     public static LocalDateTime toLocalDateTime(final Date date, final TimeZone tz) {
         final LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(Objects.requireNonNull(date, "date").getTime()),
                 Objects.requireNonNull(tz, "tz").toZoneId());
-        if (date instanceof java.sql.Timestamp ) {
+        if (date instanceof java.sql.Timestamp) {
             return localDateTime.withNano(((Timestamp) date).getNanos());
         }
         return localDateTime;
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -1648,8 +1648,9 @@ public class DateUtils {
      * @since 3.18
      */
     public static LocalDateTime toLocalDateTime(final Date date, final TimeZone tz) {
-        final LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(Objects.requireNonNull(date, "date").getTime()),
-                Objects.requireNonNull(tz, "tz").toZoneId());
+        Objects.requireNonNull(date, "date");
+        Objects.requireNonNull(tz, "tz");
+        final LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(date.getTime()), tz.toZoneId());
         if (date instanceof java.sql.Timestamp) {
             return localDateTime.withNano(((Timestamp) date).getNanos());
         }
--- a/src/main/java/org/apache/commons/lang3/time/DateUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DateUtils.java
@@ -18,8 +18,8 @@ package org.apache.commons.lang3.time;
 
 import java.text.ParseException;
 import java.text.ParsePosition;
+import java.time.Instant;
 import java.time.LocalDateTime;
-import java.time.ZoneId;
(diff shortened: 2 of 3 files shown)


garydgregory (changes requested):
-1: The implementation doesn't make sense. I suppose the question is: Do we really need a utility method for:
```
LocalDateTime.ofInstant(date.toInstant(), timeZone.toZoneId());
```

yuyu1025:
> -1: The implementation doesn't make sense. I suppose the question is: Do we really need a utility method for:
> 
> ```
> LocalDateTime.ofInstant(date.toInstant(), timeZone.toZoneId());
> ```

For experienced developers, this indeed doesn't make much sense. However, for newbies who know nothing about LocalDateTime, it can often be of great help. In project engineering, it can also reduce a lot of repetitive logic.

ppkarwasz:
> I suppose the question is: Do we really need a utility method for:
> 
> ```
> LocalDateTime.ofInstant(date.toInstant(), timeZone.toZoneId());
> ```

If you also consider [`java.sql.Date`](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Date.html) and [`java.sql.Timestamp`](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Timestamp.html)—both subclasses of `java.util.Date`—the conversion becomes less straightforward:

- `java.sql.Date.toInstant()` throws an exception.
- Instead, there are now dedicated methods: `java.sql.Date.toLocalDate()` and `java.sql.Timestamp.toLocalDateTime()`.

When you have to deal with legacy code that uses `Date`, having conversion helpers is quite useful.

yuyu1025:
> > I suppose the question is: Do we really need a utility method for:
> > ```
> > LocalDateTime.ofInstant(date.toInstant(), timeZone.toZoneId());
> > ```
> 
> If you also consider [`java.sql.Date`](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Date.html) and [`java.sql.Timestamp`](https:docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Timestamp.html)—both subclasses of `java.util.Date`—the conversion becomes less straightforward:
> 
> * `java.sql.Date.toInstant()` throws an exception.
> * Instead, there are now dedicated methods: `java.sql.Date.toLocalDate()` and `java.sql.Timestamp.toLocalDateTime()`.
> 
> When you have to deal with legacy code that uses `Date`, having conversion helpers is quite useful.

You raised a very good question, I have handled all subclasses and completed the unit tests, which will greatly improve efficiency for the development of projects involving both old and new systems.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1625,6 +1628,34 @@ public static Calendar toCalendar(final Date date, final TimeZone tz) {
...
+     * @param date the date to convert to a LocalDateTime
+     * @param tz the time zone of the {@code date}
+     * @return the created LocalDateTime
+     * @throws NullPointerException if {@code date} is null
+     * @since 3.18
+     */
+    public static LocalDateTime toLocalDateTime(final Date date, final TimeZone tz) {
+        final LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(Objects.requireNonNull(date, "date").getTime()),
+                Objects.requireNonNull(tz, "tz").toZoneId());
+        if (date instanceof java.sql.Timestamp) {
```
garydgregory, line 1653: This dependency is not OK because it brings in a module we did not depend on in the past, `java.sql`, and we've had concerns in the past that we already depend on the `java.desktop` module, but we can't change that without breaking apps in surprising ways. We should do our best for now to avoid bringing in modules outside of `java.base`.

ppkarwasz (changes requested):
@yuyu1025,

Thank you so much for the updates you’ve made to this pull request—I really appreciate the effort you’ve put in so far. Before we dive into the points Gary raised about the method’s utility, let’s take a moment to clean things up a bit more. Once that’s done, we’ll be in a better place to evaluate the next steps together.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1625,6 +1628,34 @@ public static Calendar toCalendar(final Date date, final TimeZone tz) {
...
+            return localDateTime.withNano(((Timestamp) date).getNanos());
+        }
```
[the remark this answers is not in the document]
garydgregory: Please, no usage of reflection here. This functionality is not "required," IMO. 

(This kind of dynamic dependency is surprising and makes dependency management more difficult. An app depending on Lang behaves magically differently depending on input data, completely unbeknownst to the developer, who then has to investigate and tell downstream users or other developers that their app or library needs additional dependencies.)

ppkarwasz:
@yuyu1025,

I noticed you've enabled ["Vigilant mode"](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits), but it looks like your GPG key isn’t added to your GitHub account—so your commits are showing as “Unverified.” That’s probably not what you intended.

To resolve this, [add your GPG key to your GitHub account](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account) to verify that you are the author of your commits.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
```diff
@@ -1625,6 +1628,34 @@ public static Calendar toCalendar(final Date date, final TimeZone tz) {
...
+     * Converts a {@link Date} into a {@link LocalDateTime}
+     * @param date the date to convert to a LocalDateTime
+     * @param tz the time zone of the {@code date}
+     * @return the created LocalDateTime
+     * @throws NullPointerException if {@code date} is null
+     * @since 3.18
+     */
+    public static LocalDateTime toLocalDateTime(final Date date, final TimeZone tz) {
+        final LocalDateTime localDateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(Objects.requireNonNull(date, "date").getTime()),
+                Objects.requireNonNull(tz, "tz").toZoneId());
```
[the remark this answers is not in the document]
yuyu1025: Okay, thank you for the suggestions. I'll make some adjustments.

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
```diff
@@ -1285,6 +1289,173 @@ public void testToCalendarWithTimeZoneNull() {
         assertThrows(NullPointerException.class, () -> DateUtils.toCalendar(date1, null));
     }
 
+    @Test
+    void testToLocalDateTimeWithSqlDate() {
```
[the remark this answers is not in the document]
yuyu1025: I have made modifications to this. Thank you for your suggestions.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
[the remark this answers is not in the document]
yuyu1025: I've made some adjustments. Please check them. However, what I want to say is that if compatibility with java.sql.Timestamp is not necessary, perhaps it would be a good option for us to ignore the handling of nanoseconds.

yuyu1025:
> @yuyu1025,
> 
> I noticed you've enabled ["Vigilant mode"](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits), but it looks like your GPG key isn’t added to your GitHub account—so your commits are showing as “Unverified.” That’s probably not what you intended.
> 
> To resolve this, [add your GPG key to your GitHub account](https:docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account) to verify that you are the author of your commits.

I understand, thank you for the reminder.

src/main/java/org/apache/commons/lang3/time/DateUtils.java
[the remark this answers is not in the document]
ppkarwasz: @yuyu1025, thanks!

> An app depending on Lang behaves magically differently depending on input data, completely unbeknownst to the developer, who then has to investigate and tell downstream users or other developers that their app or library needs additional dependencies.

@garydgregory The behavior of `toLocalDateTime` will not change based on which Java 9 modules the application uses—it will depend solely on the type of the argument provided:

- If the input is a `java.sql.Timestamp`, then the application must include the `java.sql` module as a dependency.
- If the application does not include `java.sql`, then it simply cannot construct or pass a `java.sql.Timestamp`, so this scenario doesn't arise.

In other words, the function's behavior is entirely predictable and consistent: the presence or absence of `java.sql` only matters if you explicitly use `java.sql.Timestamp`.

pom.xml
```diff
@@ -462,6 +463,32 @@
...
+                      <requires>
+                        static java.sql;
+                        *;
+                      </requires>
+                    </moduleInfo>
+                  </module>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
```
ppkarwasz, line 489: I attempted to update the inherited Moditect Maven Plugin configuration to include the following directive in the generated module descriptor:

```java
requires static java.sql;
```

However, due to moditect/moditect#262, this change currently has no effect.

@garydgregory, should we keep the directive in place in anticipation of the issue being fixed, or remove it for now and reintroduce it once the plugin supports it correctly?

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
```diff
@@ -1285,6 +1291,107 @@ public void testToCalendarWithTimeZoneNull() {
...
+        assertNotNull(result);
+        assertEquals(expected, result);
+    }
+
+    private static Stream<Arguments> dateConversionProvider() {
+        return Stream.of(
+                Arguments.of(
+                        java.sql.Date.valueOf("2000-01-01"),
+                        LocalDateTime.of(2000, 1, 1, 0, 0, 0)
+                ),
```
ppkarwasz, line 1307: Could you add some test cases with `java.util.Date`?

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
```diff
@@ -1285,6 +1291,107 @@ public void testToCalendarWithTimeZoneNull() {
...
+    private static Stream<Arguments> dateWithTimeZoneProvider() {
+        return Stream.of(
+                Arguments.of(
+                        java.sql.Timestamp.valueOf("2000-01-01 12:30:45"),
+                        TimeZone.getTimeZone("America/New_York"),
+                        LocalDateTime.ofInstant(
+                                java.sql.Timestamp.valueOf("2000-01-01 12:30:45").toInstant(),
+                                TimeZone.getTimeZone("America/New_York").toZoneId()
+                        )
+                ),
```
ppkarwasz, line 1355: Could you add some test cases with `java.sql.Date`?

ppkarwasz (approved):
This looks good to me!

I would just make sure that each test has a test case with `java.util.Date`, `java.sql.Date` and `java.sql.Timestamp`.

src/test/java/org/apache/commons/lang3/time/DateUtilsTest.java
yuyu1025: > Could you add some test cases with `java.sql.Date`?

OK, I'll add some now.

pom.xml
garydgregory: @ppkarwasz 
Well, no, since we shouldn't depend on java[x].sql in the first place.

All,
Maybe this kind of this-to-that conversion code belongs in Commons BeanUtils.

In Lang, DateUtils has only one kind of conversion ATM with 'toCalendar'(...). I'm not sure we want to open the door to more convertions, then there would be endless combinations due to the scale of the Java Time package.

garydgregory:
Hello @yuyu1025 

Closing: Done more simply in git master, you are credited in `changes.xml`.

PR #1392, opened 2025-05-27
Add StringUtils.truncateToByteLength

kiddos:
We sometimes need to store Unicode text in a fixed space (e.g., in a database column of type `CHARACTER(32)`). It's acceptable for the text to be truncated, but because we're dealing with Unicode, we can't simply treat the text as raw bytes and truncate it at 16 bytes — that might split a character in the middle. The function `StringUtils.truncateToByteLength(String str, int maxBytes, Charset charset)` helps handle this by safely truncating the string based on byte length while preserving valid character boundaries.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -3090,6 +3090,18 @@ public class StringUtilsTest extends AbstractLangTest {
     }
 
     @Test
+    public void testTruncateToByteLength() {
+        assertNull(StringUtils.truncateToByteLength(null, 0, Charset.defaultCharset()));
+        assertEquals("abcdefghij", StringUtils.truncateToByteLength("abcdefghijklmno", 10, Charset.defaultCharset()));
+        assertEquals("abcdefghijklmno", StringUtils.truncateToByteLength("abcdefghijklmno", 15, Charset.defaultCharset()));
+        assertEquals("abcdefghijklmno", StringUtils.truncateToByteLength("abcdefghijklmno", 20, Charset.defaultCharset()));
+        assertEquals("\u4F60\u597D\u55CE", StringUtils.truncateToByteLength("\u4F60\u597D\u55CE", 10, Charset.defaultCharset()));
+        assertEquals("\u4F60", StringUtils.truncateToByteLength("\u4F60\u597D\u55CE", 5, Charset.defaultCharset()));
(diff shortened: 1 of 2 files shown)


ecki:
Agree, very useful when dealing with UTF8 databases. Wonder if it should also deal with other encodings.

garydgregory (changes requested):
Hello all,

I think you'll want tests that cover grapheme clusters to avoid problems like https://issues.apache.org/jira/browse/LANG-1770

garydgregory:
I'm not requesting support for grapheme cluster in the runtime, but we should set expectations in unit tests, whether they are supported or not. This is a larger discussion, which I raised in https://issues.apache.org/jira/browse/LANG-1770

kiddos:
for the test case, I only check if the final output is not null and the bytes size is actully smaller then specified byte size.
Is that ok?

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -3089,6 +3089,31 @@ void testTruncate_StringIntInt() {
         assertEquals("", StringUtils.truncate("abcdefghijklmno", Integer.MAX_VALUE, Integer.MAX_VALUE));
     }
 
+    @Test
+    void testTruncateToByteLength() {
```
elharo, line 3093: These should be separate test methods for different tests so they can pass or fail independently.

ppkarwasz:
In apache/commons-io#781 I used a similar approach to introduce an equivalent functionality in Commons IO.

While approaches might differ, I think we can reuse the test cases.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
elharo: Yes, you can reuse the test class, but this is about 10 different tests that should each be a separate method.

ppkarwasz: The test suite in Commons IO is parameterized and takes into consideration the introduction of support for Unicode 15 in JDK 20 (grapheme clusters and so on): https://github.com/apache/commons-io/blob/b4ee32c53c0036429d64c0d6fe82a62a0fc6dae2/src/test/java/org/apache/commons/io/FileSystemTest.java#L171

ecki: Maybe habe one method testing the various null cases (more than at the moment) one for the usual single byte ascii cases one for bmp Tests and then one for Graphen cases and one for Dynamic. I would also avoid Emoji literals in source

PR #1394, opened 2025-05-28
Implement a LatchUtils helper class to wrap CountDownLatch instantiation along with countDown() and await() method invocations, decoupling business logic from concurrency control for safer and cleaner thread coordination

wwwzzzggg333:


garydgregory:
I don't see this as useful or fitting in our existing classes. We already have concurrency and lock utilities that this code doesn't reuse, which would create confusion IMO.

garydgregory:
Closing: No more feedback, from anyone.

PR #1398, opened 2025-06-08
Add string utils resize consecutive chars

trungpnt-byte:

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -581,6 +581,7 @@ The <action> type attribute can be add,update,fix,remove.
     <action                   type="add" dev="ggregory" due-to="Gary Gregory">Add JavaVersion.JAVA_17.</action>
     <action issue="LANG-1636" type="add" dev="ggregory" due-to="">Add missing boolean[] join method #686.</action>
     <action                   type="add" dev="ggregory" due-to="Gary Gregory">Add StringUtils.substringBefore(String, int).</action>
+    <action                   type="add" dev="trungpnt" due-to="Trung Pham">Add StringUtils.resizeConsecutiveChars(String, char, int, boolean, boolean).</action>
     <action                   type="add" dev="ggregory" due-to="Gary Gregory">Add Range.INTEGER.</action>
     <action                   type="add" dev="ggregory" due-to="Gary Gregory">Add DurationUtils.</action>
     <action                   type="add" dev="jochen">Introduce the use of @Nonnull, and @Nullable, and the Objects class as a helper tool.</action>
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -9288,6 +9288,80 @@ public class StringUtils {
     }
 
     /**
+     * Resize consecutive characters down to given repeated times.
+     * <p>
+     *     A new string is created with the consecutive characters
+     * </p>
+     *
+     * <pre>
+     *     StringUtils.resizeConsecutiveChars("aaabbbcc", 'a', 2, false, false) = "aabbcc"
+     *     StringUtils.resizeConsecutiveChars("11111 111111 111111    111111", '1', 1, false, false) = "1 1 1    1"
+     * </pre>
+     *     For example, the string "aaabbbcc" with a appointedChar of 'a' and
+     * repeatedCount of 2 will return "aabbcc".
+     * <p>Note that the appointedChar is not removed from the string.
+     * <p>Note that the repeatedCount is not checked for negative values.
+     * @param str the string to resize
+     * @param appointedChar the character to resize
+     * @param repeatedCount the number of times to repeat the character, default is 1
+     * @param isRemovedTrailing if true, the trailing characters will be removed
+     * @param isRemovedLeading if true, the leading characters will be removed
+     * @return the resized string, or the original string if no changes were made
+     * @since 3.5
+     */
+
+    public static String resizeConsecutiveChars(String str, char appointedChar, Integer repeatedCount, boolean isRemovedTrailing, boolean isRemovedLeading) {
+        if (str == null || str.isEmpty()) {
+            return str;
+        }
+        int xTimes = repeatedCount != null ? repeatedCount : 1;
+        char[] collector = new char[str.length()];
+        int collectorIdx = 0;
+        char[] statefulSeparator = new char[1];
+        for (int i = 0; i < str.length(); i++) {
+            char current = str.charAt(i);
+            if (current == appointedChar) {
+                if (statefulSeparator[0] == '\u0000') {
+                    statefulSeparator[0] = appointedChar;
+                }
+            } else {
+                if (statefulSeparator[0] == appointedChar) {
+                    if (collectorIdx > 0 || !isRemovedLeading) {
+                        int repetition = xTimes;
+                        while (repetition-- > 0) {
+                            collector[collectorIdx++] = appointedChar;
+                        }
+                        statefulSeparator[0] = '\u0000';
+                    }
+                }
+                collector[collectorIdx++] = current;
+            }
+        }
+        if (!isRemovedTrailing && statefulSeparator[0] == appointedChar) {
+            int repetition = xTimes;
+            while (repetition-- > 0) {
+                collector[collectorIdx++] = appointedChar;
+            }
+        }
+        return new String(collector, 0, collectorIdx);
+    }
+
+    /**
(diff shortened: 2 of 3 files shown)


ppkarwasz:
Hi @trungpnt-byte, thanks for the PR!

This looks like a fairly specific use case, and I’m not sure it fits the general-purpose scope of Commons Lang. Could you share more about how you use it and why it might be broadly useful?

Appreciate your contribution!

garydgregory:
-1 as is, this is not event a draft since there is zero Javadoc to document the intent, and there is zero text in the PR description to us understand the use case. I also agree with the assessment from @ppkarwasz

trungpnt-byte:
Dear @ppkarwasz , @garydgregory, thank you for your suggestions. 
In fact I wrote the Javadoc on the method including the common usage but running ```mvn``` resulting in some unfamiliar errors so I opened a draft PR ( which I think is not an official PR yet ) to add more edge cases tests and fix it later.

ppkarwasz:
Hi @trungpnt-byte,

We typically don't review draft PRs, but in this case our concern is less about the implementation details and more about the fundamental utility of the proposed `repeatChars` method. As it stands, this method feels more like a novelty—perhaps useful in coding exercises, but lacking a **compelling use case** in practical, real-world scenarios.

Before investing further effort into polishing the PR, we **really** recommend starting a discussion—either here or on the [`dev@commons.apache.org`](mailto:dev@commons.apache.org) mailing list—about the *usefulness* of this method. If you can demonstrate concrete, real-world use cases where `repeatChars` would be genuinely helpful, and the community finds them convincing, then it would make sense to proceed with refining the PR.

Otherwise, there's a **real risk** that you’re solving a problem nobody actually has, and we want to spare you from spending time on work that will **not be accepted**.

You might also want to take a look at the [list of open LANG JIRA issues](https://issues.apache.org/jira/browse/LANG-1769?filter=-1&jql=project%20%3D%20LANG%20AND%20resolution%20%3D%20Unresolved%20order%20by%20updated%20DESC). These represent long-standing problems users have reported and could be a good starting point if you're looking for ways to contribute. @garydgregory may be able to suggest specific issues that need attention.

Thanks again, and we look forward to your use cases for the `repeatChars` method.

trungpnt-byte:
Dear @ppkarwasz, 
thank you for the detailed reply and I really appreciate them !
For the ```repeatChars()``` method, it's a generalized version of ```ensureOneCharacterBetweenWords()```, which then generalizes ```ensureOneWhitespaceBetweenCharacters()``` - that's why in my draft PR, the ```ensureOneWhitespaceBetweenCharacters()``` calls the ```repeatChars()``` method.
These are the malformed values I've encountered while working on my company products - which allow free-text inputs:

For general strings
1. ```"This is       a delivery    comment"```
2. ```"Please   call  Mr.  John  on arrival"```
3. ```"{prev_     loan} + 10 *      2"```
if we need to sanitize the above string nicely - calling 
```repeatChars("{prev_     loan} + 10 *      2", repeatOn = <space>, reps = 0)```
will output  ```{prev_loan}+10*2```
image
then consumer methods may benefit more - for e.g : the Shunting Yard algorithm won't have to check & skip those redundant spaces to reach the next expected operator or operand character, the reserved token prev_loan in the system will be correct before being compared.

For specific values such as longitude / latitude

4.  ``` "0...323" ``` , etc // we want this to be ```0.323```  => _**repeat the dot char only 1, only if dot chars are consecutive.**_

Supposed we need to check if those values exist or just sanitize them before persisting, from my own experience, I could hardly find an optimal way to sanitize these free-text strings to be nicely separated, think the ```trim()``` method would only discard spaces from both ends, not those between each word, so a combination of ```repeatChars()``` or ```ensureOneWSpace()``` and ```trim()``` or even internally trimming both-ended chars inside ```ensureOneWSpace()``` / ```repeatChars()``` would be nice. 

image

image

For this, if there is an existing optimal solution, then I'd love to be all ears.

To explain a bit regarding the ```repeatChars()```,
In those malformed values, the whitespaces appear consecutive ( at least 2 ) to be resized down to 1, so my mind kind of went to Mars to generalize everything into the ```repeatChars()``` - it means it will rebuild a new string that includes the appointed characters to be repeated [x] times if they are consecutive ( configurable ).
Some further use case examples :
- ```StringUtils.repeatChars("11111 111111 111111    111111",Collections.singletonList('1')),2,true)); // "11 11 11    11"```
-  ```StringUtils.repeatChars("11111 222222 333333   444",Arrays.asList('1','2','3','4')),2,true)); // "11 22 33   44"```
-  ```StringUtils.repeatChars("111111234",Arrays.asList('1','2','3','4')),2,true)); // "11234"```

As you can see, it basically auto-corrects / reduces the redundant consecutive characters to configurable repetitions.

Also, as I've extensively used some ```apache-commons'``` ```StringUtils``` method for production code, I think that the ```repeatChar()``` method may be useful elsewhere.

I think perhaps the ```repeatChars()``` method seems too generalized for the above, so if you think this will be helpful, I'll just include the ```ensureOneCharacterBetweenWords()``` or  ```ensureOneWhitespace()```

Thank you very much ! 
I would love to learn more from you guys !

ppkarwasz:
Hi @trungpnt-byte,

Thank you for providing these examples.

> These are the malformed values I've encountered while working on my company products - which allow free-text inputs:

I agree that handling malformed or inconsistent user input is a common challenge.

> For general strings
> 
> 1. `"This is       a delivery    comment"`
> 2. `"Please   call  Mr.  John  on arrival"`

In these cases, collapsing multiple consecutive whitespace characters into a single space is a sensible and common need, as extra spaces are **difficult to notice** and typically unintentional. However, [`StringUtils.normalizeSpace`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#normalizeSpace%28java.lang.String%29) already handles this case effectively.

`normalizeSpace` is more general than the proposed `repeatChars()`, as it collapses sequences of **any** ASCII whitespace character—such as spaces, tabs, and newlines—into a single space. In contrast, `repeatChars()` would not collapse mixed whitespace sequences (e.g., `" \t\n"`).

> 3. `"{prev_     loan} + 10 *      2"`
>     if we need to sanitize the above string nicely - calling `repeatChars("{prev_     loan} + 10 *      2", repeatOn = <space>, reps = 0)` will output  `{prev_loan}+10*2`

This is a good illustration of a scenario where whitespace needs to be removed entirely. A similar issue I’ve personally encountered was with IBAN inputs from banking UIs that included unexpected spaces, which caused validation to fail. In this case, however, [`StringUtils.deleteWhitespace`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#deleteWhitespace%28java.lang.String%29) is the most suitable utility.

> For specific values such as longitude / latitude
> 
> 4. `"0...323"` , etc // we want this to be `0.323`  => _**repeat the dot char only 1, only if dot chars are consecutive.**_

This seems less like a generic string manipulation case and more like **domain-specific sanitization**. Unlike whitespace, repeated `.` characters are visibl, and I’m not sure it's common enough to justify general-purpose support in `StringUtils`.

> To explain a bit regarding the `repeatChars()`, In those malformed values, the whitespaces appear consecutive ( at least 2 ) to be resized down to 1, so my mind kind of **went to Mars to generalize everything** into the `repeatChars()` - it means it will rebuild a new string that includes the appointed characters to be repeated [x] times if they are consecutive ( configurable ).

I appreciate the idea of a configurable utility like `repeatChars()`, and it’s a creative generalization. That said, I believe this approach may be **overgeneralizing**. In practice, I can only think of whitespace (and possibly a few other niche cases) where collapsing repeated characters is useful. I haven't seen a clear use case where collapsing any arbitrary character sequence to a configurable count is necessary, or where such logic wouldn't be better handled specifically within the consuming application.

For that reason, I’m hesitant to introduce this as part of the public API in `StringUtils`.

trungpnt-byte:
Dear @ppkarwasz , thank you very much for the details, definitely good learning resources for me as well xD 
I'll go through some issues in the Jira boards, I'll first request for an account to join the community, don't know if you guys are on slack xD

ppkarwasz:
Hi @trungpnt-byte,

While many of us are active on various Slack workspaces, the Apache Software Foundation only operates a Slack reserved for Apache committers. The best way to connect with the Apache Commons development team is via the [`dev@commons.apache.org` mailing list](https://commons.apache.org/mail-lists.html). Just make sure to subscribe before posting—this helps you receive replies directly and avoids moderation delays.

PR #1401, opened 2025-06-22
Add string utils repeat chars x times

trungpnt-byte:


ppkarwasz:
Hi @trungpnt-byte,

This PR looks like a modified version of your earlier proposal in #1398. Could you take a moment to respond to the questions raised there? Specifically:

> Could you share more about how you use it and why it might be broadly useful?

Understanding the broader use case will help us assess whether this addition makes sense for the library as a whole.

garydgregory:
Hello @trungpnt-byte,

I don't think we want to grow the public API in this class with these kinds of hyper-specifc and complex methods beyond what we already have.

This looks like a duplicate of https://github.com/apache/commons-lang/pull/1398, which causes additional churn for maintainers.

All the comments from https://github.com/apache/commons-lang/pull/1398 stand IMO.

garydgregory:
Closing: No feedback in 30 days.

PR #1405, opened 2025-07-06
[LANG-1776] Use GitHub URL in `project.scm.url`

ppkarwasz:
This PR updates the `project.scm.url` field in the Maven POM to point to the GitHub repository URL (`https://github.com/apache/commons-lang`).

While this field is not used directly by Maven during builds, it is increasingly leveraged by tools like *Dependabot* to:

* Associate the Maven artifact with its GitHub repository
* Automatically include release notes, changelogs, and commit history in pull requests for new versions

By using the GitHub URL, we improve the quality and completeness of metadata shown in automated upgrade PRs and make it easier for users to trace changes between releases.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
   <scm>
     <connection>scm:git:http://gitbox.apache.org/repos/asf/commons-lang.git</connection>
     <developerConnection>scm:git:https://gitbox.apache.org/repos/asf/commons-lang.git</developerConnection>
-    <url>https://gitbox.apache.org/repos/asf?p=commons-lang.git</url>
+    <url>https://github.com/apache/commons-lang</url>
     <tag>rel/commons-lang-3.18.0</tag>
   </scm>
   <!-- Lang should depend on very little -->
--- a/pom.xml
+++ b/pom.xml
@@ -41,14 +41,18 @@
   Please ensure your build environment is up-to-date and kindly report any build issues.
   </description>
   <url>https://commons.apache.org/proper/commons-lang/</url>
+  <ciManagement>
+    <system>github</system>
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @ppkarwasz 

Do we really want to point to a non-Apache resource for something that is NOT the canonical Apache repository? I would think twice about this...

Dependabot doesn't seem to need this change FWIW, it just works fine today.

ppkarwasz:
> Do we really want to point to a non-Apache resource for something that is NOT the canonical Apache repository? I would think twice about this...

That's a valid concern. I've reverted `project.scm.url` back to `https://gitbox.apache.org/repos/asf/commons-lang.git`, which remains under ASF control and currently redirects to GitHub.

To help tooling like Dependabot discover the GitHub repository (without compromising the canonical SCM reference), I've added a `ciManagement` element in commit 3b3ed38ab43975fc1df3f1ae1fbea70c5d3bb086, which points to GitHub Actions. This satisfies Dependabot’s fallback mechanism that looks for GitHub URLs elsewhere in the POM and enables it to retrieve changelogs and commit history for new releases.

> Dependabot doesn't seem to need this change FWIW, it just works fine today.

That's mostly true — Dependabot can detect and propose version bumps even without this change. However, without a GitHub URL present, it cannot automatically enrich PRs with changelogs and commit diffs. For an example of this difference in behavior, see PR ppkarwasz/logging-log4j2#566, where `commons-csv` includes rich metadata, while `commons-lang3` does not.

This small metadata tweak improves the user experience for downstream consumers without altering the authoritative SCM reference.

garydgregory:
TY @ppkarwasz Merged.

PR #1411, opened 2025-07-13
[LANG-927] Add method convert a iterator to array in ArrayUtils

Abyss-lord:
This PR fixes issue LANG-927 which ask to "Create Iterable APIs"

✅ Improvement:
Create Iterable APIs, add two methods in `ArrayUtils`.

- `T[] iteratorToArray(Iterator<T> iterator)`
-  `T[] iteratorToArray(Iterator<T> iterator, Class<T> clazz)`

✅ Tests:

Added two test methods in `ArrayUtilsTest` and All other tests pass

- `testIteratorToArrayWithClazz`
- `testIteratorToArray`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -21,11 +21,14 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.security.SecureRandom;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Random;
@@ -8860,6 +8863,63 @@ public class ArrayUtils {
      */
     public static <T> T[] toArray(@SuppressWarnings("unchecked") final T... items) {
         return items;
+    }
+
+    /**
+     * Converts an {@link Iterator} into an array.
+     * <p>
+     * Returns {@code null} if the input iterator is {@code null}.
+     * If the iterator has no elements, an empty {@code Object[]} is returned.
+     * </p>
+     * <p>
+     * Note: The returned array has runtime type {@code Object[]}, and requires that all elements
+     * in the iterator are of the same type. If a type-safe array is needed, use
+     * {@link #iteratorToArray(Iterator, Class)} instead.
+     * </p>
+     *
+     * @param iterator the iterator to convert, may be {@code null}
+     * @param <T> the element type
+     * @return the array containing elements from the iterator,
+     *         or {@code null} if the input is {@code null}
+     * @since 3.18
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T[] iteratorToArray(Iterator<T> iterator) {
+        if (iterator == null) {
+            return null;
+        }
+
+        if (!iterator.hasNext()) {
+            return (T[]) ArrayUtils.EMPTY_OBJECT_ARRAY;
+        }
+
+        return (T[]) Streams.of(iterator).toArray();
+    }
+
+    /**
+     * Converts an {@link Iterator} into a typed array.
+     * <p>
+     * Returns {@code null} if the input iterator or class is {@code null}.
+     * If the iterator has no elements, an empty array of the specified type is returned.
+     * </p>
+     *
+     * @param iterator the iterator to convert, may be {@code null}
+     * @param clazz    the class of the array component type, must not be {@code null}
+     * @param <T>      the element type
+     * @return the array containing elements from the iterator, or {@code null} if the input is {@code null}
+     * @since 3.18
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T[] iteratorToArray(Iterator<T> iterator, Class<T> clazz) {
+        if (iterator == null || clazz == null) {
+            return null;
+        }
+
+        List<T> list = new ArrayList<>();
+        iterator.forEachRemaining(list::add);
+
+        T[] array = (T[]) Array.newInstance(clazz, list.size());
+        return list.toArray(array);
     }
 
     /**
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -8915,10 +8915,10 @@ public class ArrayUtils {
             return null;
         }
 
-        List<T> list = new ArrayList<>();
+        final List<T> list = new ArrayList<>();
         iterator.forEachRemaining(list::add);
 
-        T[] array = (T[]) Array.newInstance(clazz, list.size());
+        final T[] array = (T[]) Array.newInstance(clazz, list.size());
         return list.toArray(array);
     }
 
--- a/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/ArrayUtilsTest.java
@@ -6454,50 +6454,50 @@ class ArrayUtilsTest extends AbstractLangTest {
     }
 
     @Test
-    void testIteratorToArray(){
-        Iterator<String> iterator = Arrays.asList("one", "two", "three").iterator();
-        Object[] normalResult = ArrayUtils.iteratorToArray(iterator);
+    void testIteratorToArray() {
+        final Iterator<String> iterator = Arrays.asList("one", "two", "three").iterator();
+        final Object[] normalResult = ArrayUtils.iteratorToArray(iterator);
 
         assertArrayEquals(new Object[]{"one", "two", "three"}, normalResult);
 
-        Iterator<Object> emptyIterator = Collections.emptyIterator();
-        Object[] emptyArray = ArrayUtils.iteratorToArray(emptyIterator);
+        final Iterator<Object> emptyIterator = Collections.emptyIterator();
+        final Object[] emptyArray = ArrayUtils.iteratorToArray(emptyIterator);
 
         assertEquals(0, emptyArray.length);
         assertNull(ArrayUtils.iteratorToArray(null));
 
-        Iterator<Integer> singletonIterator = Collections.singletonList(42).iterator();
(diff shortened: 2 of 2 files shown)


Abyss-lord:
Hi @garydgregory, could you please review this PR when you have time? I’d really appreciate your feedback.

garydgregory:
This looks like a duplicate for https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/IteratorUtils.html

Abyss-lord:
> This looks like a duplicate for https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/IteratorUtils.html

@garydgregory Thanks for pointing out, I think we should close this PR and jira-issue. WDYT.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8862,6 +8865,63 @@ public static <T> T[] toArray(@SuppressWarnings("unchecked") final T... items) {
...
+     * </p>
+     *
+     * @param iterator the iterator to convert, may be {@code null}
+     * @param <T> the element type
+     * @return the array containing elements from the iterator,
+     *         or {@code null} if the input is {@code null}
+     * @since 3.18
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T[] iteratorToArray(Iterator<T> iterator) {
```
ppkarwasz, line 8887: This method causes heap pollution: the return type `T[]` can't be created safely due to type erasure.
The returned type will always be `Object[]`, which will cause a `ClassCastException` at runtime, when used:

```java
List<String> list = List.of("one", "two", "three");
// Fails at runtime
String[] array = iteratorToArray(list.iterator());
```

src/main/java/org/apache/commons/lang3/ArrayUtils.java
```diff
@@ -8862,6 +8865,63 @@ public static <T> T[] toArray(@SuppressWarnings("unchecked") final T... items) {
...
+        if (iterator == null || clazz == null) {
+            return null;
+        }
+
+        final List<T> list = new ArrayList<>();
+        iterator.forEachRemaining(list::add);
+
+        final T[] array = (T[]) Array.newInstance(clazz, list.size());
+        return list.toArray(array);
+    }
```
ppkarwasz, line 8923: Some remarks:

* While returning `null` when the `iterator` is `null` is consistent with other `ArrayUtils` methods, I would not accept a `null` `Class` argument.
* The method can be safely generalized to accept an `Iterator<? extends T>` to support subtypes.
* It is not necessary to copy elements into a temporary list first; the array can be constructed directly via `Stream`.

```suggestion
    public static <T> T[] iteratorToArray(Iterator<? extends T> iterator, Class<T> clazz) {
        Objects.requireNonNull(clazz, "clazz must not be null");
        return iterator == null
                ? null
                 : Streams.of(iterator).toArray(size -> (T[]) Array.newInstance(clazz, size));
    }
```

ppkarwasz (changes requested):
Hi @Abyss-lord,

Thank you for your PR!

At first glance, I'm not convinced that the proposed methods offer a clear improvement over existing options. For example, consider the following comparisons:

```java
Iterator<String> iterator = ...
Object[] objectArray = ArrayUtils.iteratorToArray(iterator);
String[] stringArray = ArrayUtils.iteratorToArray(iterator, String.class);
```

versus the equivalent using existing `Streams` support:

```java
Iterator<String> iterator = ...
Object[] objectArray = Streams.of(iterator).toArray();
String[] stringArray = Streams.of(iterator).toArray(String[]::new);
```

Functionally, these are very similar. The main difference is that the `ArrayUtils` versions return `null` when the input is `null`, which—depending on perspective—can be a pro or a con. Personally, I view this as a downside: I prefer to eliminate `null`s early and treat empty arrays and `null`s as equivalent.

Could you share more context on your use case or motivation for introducing these methods? That might help clarify their intended value.

Also, setting aside the question of inclusion, I believe the proposed implementations need some adjustments. I've left comments on the relevant lines for review.

Thanks again for the contribution—looking forward to your thoughts!

ppkarwasz:
> This looks like a duplicate for https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/IteratorUtils.html

Right, nice catch! :wink:

src/main/java/org/apache/commons/lang3/ArrayUtils.java
Abyss-lord: > This method causes heap pollution: the return type `T[]` can't be created safely due to type erasure. The returned type will always be `Object[]`, which will cause a `ClassCastException` at runtime, when used:
> 
> ```java
> List<String> list = List.of("one", "two", "three");
> // Fails at runtime
> String[] array = iteratorToArray(list.iterator());
> ```

Thank you for the support. While this PR can be closed, I do have a follow-up question:
In the absence of the `Class` argument, would it be more appropriate for the method to return an `Object[]`?

Abyss-lord:
> Hi @Abyss-lord,
> 
> Thank you for your PR!
> 
> At first glance, I'm not convinced that the proposed methods offer a clear improvement over existing options. For example, consider the following comparisons:
> 
> ```java
> Iterator<String> iterator = ...
> Object[] objectArray = ArrayUtils.iteratorToArray(iterator);
> String[] stringArray = ArrayUtils.iteratorToArray(iterator, String.class);
> ```
> 
> versus the equivalent using existing `Streams` support:
> 
> ```java
> Iterator<String> iterator = ...
> Object[] objectArray = Streams.of(iterator).toArray();
> String[] stringArray = Streams.of(iterator).toArray(String[]::new);
> ```
> 
> Functionally, these are very similar. The main difference is that the `ArrayUtils` versions return `null` when the input is `null`, which—depending on perspective—can be a pro or a con. Personally, I view this as a downside: I prefer to eliminate `null`s early and treat empty arrays and `null`s as equivalent.
> 
> Could you share more context on your use case or motivation for introducing these methods? That might help clarify their intended value.
> 
> Also, setting aside the question of inclusion, I believe the proposed implementations need some adjustments. I've left comments on the relevant lines for review.
> 
> Thanks again for the contribution—looking forward to your thoughts!

@ppkarwasz , thank you for the detailed review and thoughtful feedback!

The reason I added this method is because I came across a related request in a JIRA issue, where someone expressed the need for this functionality. As a newcomer to the Commons Lang community, I wanted to start contributing by picking up something small but potentially useful.

Also, could you please advise where I can find guidelines or existing examples to help identify the kinds of utility methods that are considered suitable for inclusion in Commons Lang? That would really help me better align future contributions.

Thanks again!

src/main/java/org/apache/commons/lang3/ArrayUtils.java
Abyss-lord: Thank you for the suggestions—really helpful!

Also, if I may ask a quick question: what formatting tool does Commons Lang use for code style enforcement? I’ve been using `Spotless` in other projects, but I’d like to make sure my contributions follow the correct convention here.

Thanks again!

ppkarwasz: Commons has some checkstyle rules, but doesn't use a deterministic formatter.

Log4j and Maven uses the [`palantir-java-formatter`](https://github.com/palantir/palantir-java-format), so maybe it would be time to switch to it too. Feel free to start a thread on `dev@commons.apache.org`.

src/main/java/org/apache/commons/lang3/ArrayUtils.java
ppkarwasz: Yes, `Object[]` would be more appropriate. It is also the only choice here.

ppkarwasz:
> Also, could you please advise where I can find guidelines or existing examples to help identify the kinds of utility methods that are considered suitable for inclusion in Commons Lang? That would really help me better align future contributions.
Here's a refined and more structured version of your answer, keeping your key points but improving clarity, tone, and flow:

That’s a great question — and one that @garydgregory is best positioned to answer.

That said, here are my *two cents* based on experience:

### What Commons Lang Is For

Commons Lang was originally created to fill gaps in early JDKs — to provide utility methods that were missing or cumbersome to implement. Classic examples include:

* Better handling of `Date` and `Calendar`
* Utility methods for `String` (like `isBlank()`, `join()`, etc.)
* Helper functions for working with arrays
* Safer handling of `null` values

### How the JDK Has Evolved

Many of the original pain points Commons Lang addressed have since been resolved in newer JDK versions:

* `java.time` (introduced in JDK 8) largely replaces the need for `DateUtils`.
* Methods like `String.isBlank()` were added in JDK 11.
* Arrays are still performant, but have been largely replaced in day-to-day code by `List` and `Stream` APIs.
* Static null checks and null-safe wrappers are less necessary now that tools like IDEs and static analyzers offer better nullability support.

  For example, this:

  ```java
  StringUtils.isEmpty(string)
  ```

  is now often replaced by simply ensuring `string` is non-null and writing:

  ```java
  string.isEmpty()
  ```

  Also, null-propagating utility methods can create maintenance burdens: if one method returns `null` when given `null`, others in a call chain (like `baz(bar(foo(x)))`) must do the same to avoid `NullPointerException`s — making them harder to reason about (no `@NonNull` guarantee) and more verbose.

### In Summary

It’s getting harder to find genuinely useful additions to Commons Lang — and that's a good thing. The JDK has improved a lot. Rather than expanding Commons Lang with new utilities, it might be more helpful to offer guidance or utilities to help users *migrate away from* it in favor of modern JDK APIs where possible.

But again, for definitive inclusion guidelines or strategic direction, I’d defer to @garydgregory or `dev@commons.apache.org`.

garydgregory:
While I mostly agree with @ppkarwasz, I disagree with migrating away from the library. WRT using APIs in the fluent style, there has been a recent discussion (either in Jira, GitHub, or the mailing list, sorry I can't recall) about normalizing null versus `""` versus `[]` in classes like `StringUtils` but that was a presentation of the issue, not a proposal but compatibility would have to be considered. 

I do not have specific guidance to give @Abyss-lord on what to add out of thin air, as I do not have a to-do list for new methods and Jira doesn't have tickets for new code to add, IIRC. What Jira does have, on the other hand, are bug reports to process, research, and propose changes.

In general, I would say that Lang additions reflect real-world use cases from other libraries and applications that I've dealt with in the past in both an open-source and corporate context.

HTH

garydgregory:
Closing: This PR duplicates code in Commons Collection.

PR #1418, opened 2025-07-20
Fix null handling in ArrayUtils.toStringArray(Object[]) to avoid NPE

Naveenkumarsuk:
This PR fixes a potential NullPointerException in `ArrayUtils.toStringArray(Object[])` when the input array contains null elements. The method now delegates to `toStringArray(Object[], String)` with a default value of `"null"`, ensuring null-safe behavior.

Includes test coverage in `ArrayUtilsTest.java`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -9525,40 +9525,43 @@ public class ArrayUtils {
         return new ToStringBuilder(array, ToStringStyle.SIMPLE_STYLE).append(array).toString();
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


garydgregory:
-1: This is the same kind of change you tried in your other 2 pull requests. This is noise at this point. Some of these changes are the same as in another open PR. 

I'll say it again: You need a unit test to prove that the changes in `main` fix anything. The test must fail when the changes to `main` are not applied. This doesn't appear to fix anything; it just copies and duplicates code.

Follow https://github.com/apache/commons-lang/blob/master/.github/pull_request_template.md, which is normally in the description of a new PR. Don't delete templated text, add to it your PR.

garydgregory:
Closing: no reply.

PR #1447, opened 2025-09-14
[LANG-1769] Support lenient mode in FastDateFormat and FastDateParser like SimpleDateFormat

madhurlathi:
This PR extended `FastDateFormat` to support a strict (lenient) mode, following the same pattern as `SimpleDateFormat`: 
Thanks @ppkarwasz  for the direction. 

- Added methods: `isLenient`, `setLenient` and `reset`. 
- Throws an exception for invalid input data and lenient is false. 
- Added tests for new scenarios and confirmed all existing and new test cases pass with `mvn`. 

`reset()` method resets the object's state because the instance is cached and reused. 

There is support for serialization and deserialization of the `FastDateParser`. 
However, retaining the lenient value after serialization is not added. It currently resets to the default value. 
This helps avoid unintentional exceptions when lenient is set to false. Users can explicitly set the value if needed.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/FastDateFormat.java
+++ b/src/main/java/org/apache/commons/lang3/time/FastDateFormat.java
@@ -103,6 +103,13 @@ public class FastDateFormat extends Format implements DateParser, DatePrinter {
 
     private static final AbstractFormatCache<FastDateFormat> CACHE = new AbstractFormatCache<FastDateFormat>() {
         @Override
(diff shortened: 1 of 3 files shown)


garydgregory:
This class is being considered for depreciation in [LANG-1763](https://issues.apache.org/jira/browse/LANG-1763), so we shouldn't add more features to it until 1763 is resolved one way or another.

madhurlathi:
Ok. Thanks a lot.

Closing this PR.

PR #1449, opened 2025-09-23
Optimize ObjectToStringComparator.compare() method

mymde:
Modified the compare() method implementation to use if (o1 == o2) instead of  the original if (o1 == null && o2 == null) check. This allows for earlier  identification of identical objects (including null objects), avoiding unnecessary toString() calls and string comparison operations, improving performance.


garydgregory:
Hello @mymde 

Thank you for your PR. It looks to me like this is only 1/2 the change. I completed it in git master. You are credited in `changes.xml`.

Please verify your use and close this PR if appropriate.

TY!

mymde:
Hi @garydgregory ,
Thanks for finishing the change and crediting me! I’ll verify and close the PR soon.
Thanks again!

PR #1454, opened 2025-09-29
Fix internal inverted logic in private isEnum() method and correct its usage in getFirstEnum()

mymde:
Fix isEnum() Method Logic:
Correct the method to return true only when enumClass is not null and is actually an enum class
Change implementation to: return enumClass != null && enumClass.isEnum();

Update getFirstEnum() Method:
Invert the condition check to properly handle valid enum classes
Should proceed with stream operations only when isEnum(enumClass) returns true


garydgregory:
Merged, thank you @mymde !

PR #1455, opened 2025-10-02
Add ObjectUtils.ifNotNull helper method

stratozero:

This PR introduces two new utility methods in `ObjectUtils` to reduce boilerplate 
when executing an action only if an object is not null.

### Rationale
- Currently, developers either write explicit null checks or use 
  `Optional.ofNullable(obj).ifPresent(...)`.
- `Optional` adds readability but creates an extra object which is unnecessary 
  for simple null-guarded operations.
- Usual null check is prone to the error of calling a getter method twice, (like `if (obj.getX() != null) consume(obj.getX())`), a pattern seen many times in a lot of projects, which is an error in many ways:
   - causes twice the access to the property
   - potentially doubles the side effects
   - can be a partial guard as it actually does not prevent the getter to return null in the second call
- A static utility method in `ObjectUtils` makes the intent clearer and avoids 
  extra allocations.

### Proposed API
```java
public static <T> void ifNotNull(T obj, Consumer<? super T> action) {
    if (obj != null) {
        action.accept(obj);
    }
}
```

The proposed API could then be used like `ifNotNull(obj.getX(), x -> consume(x))` preventing the creation of an unnecessary Optional and avoiding the possibility of accessing a property twice.

Also, to accomplish the task of not allocating extra objects, we need to take care of injecting some context as a parameter (to avoid capturing lambdas):

``` java
public static <T, C> void ifNotNull(T obj, C ctx, BiConsumer<? super T, ? super C> action) {
    if (obj != null) {
        action.accept(obj, ctx);
    }
}
```

So that it becomes possible to write `ifNotNull(obj.getX(), this, (x, ctx) -> XType::setContext)`

-----------------------------------------

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best-practice.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ObjectUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ObjectUtils.java
@@ -30,6 +30,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeSet;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


stratozero:
Thank you for pointing out `Consumers.apply` - I wasn’t aware of it.

I agree that it provides the same functionality, but I would still argue for having this method also exposed under `ObjectUtils`:

- **Discoverability**: `ObjectUtils` is the natural place where developers look for
  null-handling helpers (`defaultIfNull`, `firstNonNull`, etc.).  
  Personally, I would not think of looking under `Consumers` for such a method.

- **Usability**: `ObjectUtils` is frequently imported statically 
  (`import static ObjectUtils.*`), which makes methods like `ifNotNull` extremely concise
  and encourages adoption.

- **API design**: the parameter order in `Consumers.apply` is reversed
  (lambda first, then the argument).  
  Having the object first (`ifNotNull(obj, consumer)`) reads more naturally and is more
  consistent with other `ObjectUtils` methods.

Therefore, even if the underlying logic is a duplicate, placing this utility in
`ObjectUtils` would simplify adoption and improve the intuitiveness of the API surface.

garydgregory:
> but I would still argue for having this method also exposed under ObjectUtils

Hello @stratozero 

Definitely not; the point of Commons Lang is to provide reusable APIs. This, in turn, allows other libraries and applications to avoid duplicating the same code and follow the DRY principle. Certainly Commons Lang is not going to create copy-pasta in its own codebase.

PR #1470, opened 2025-10-22
[LANG-1685] reflectionToString falls back to toString if reflective access is not permitted.

iwauo:
### Description
The `ToStringBuilder.reflectionToString()` implementation previously threw an exception when access to a field was prohibited by JPMS.  This patch adds defensive handling so that, when reflective access is not permitted, the method falls back to invoking the target object's native `toString()`.
Additionally, the change is implemented to work on JVMs that do not support JPMS (JDK 8).

----

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionToStringBuilder.java
@@ -849,6 +849,12 @@ public class ReflectionToStringBuilder extends ToStringBuilder {
         validate();
 
         Class<?> clazz = getObject().getClass();
+
+        if (!reflectiveAccess.isAllowed(clazz)) {
+          appendToString(getStyle().getContentStart() + getObject().toString() + getStyle().getContentEnd());
+          return super.toString();
+        }
+
         appendFieldsIn(clazz);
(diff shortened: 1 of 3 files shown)


iwauo:
I made a minor code change to refactor. Verified on JDK 8, 11, and 21.

iwauo:
Hi @garydgregory , could you please take a look at this PR and let me know your thoughts? I'd appreciate any feedback you have.
Please let me explain our context. I am a programmer working for a consulting company, and we have some clients who are planning to migrate their legacy systems to the latest JDK due to the EOL of older JDKs or middlewares they depend on. However, most of those systems depend on this library, and the ReflectionToString class is used for logging or other features. As a result, this issue is becoming a critical issue for those projects.
Of course, --add-opens is a solution to this problem. However, some clients refuse to set those flags for security reasons, although I understand that the main goal of the JPMS is improving modularity, not security. I believe that there must be many other projects that have the same issue."

LI123456mo:
Hi,@garydgregory ,@iwauo,, is there any contribution that you can assighn to me as a beginner , who only knows java DSA in this project , I would like to contribute, or can you guide me?

garydgregory:
Jira is https://issues.apache.org/jira/browse/LANG-1685

garydgregory:
Closing in favor of https://github.com/apache/commons-lang/pull/1647

PR #1521, opened 2025-12-09
[LANG-1801] Fix RandomStringUtils.random() does not strictly validate start/end when chars != null, causing potential IndexOutOfBoundsException

IcoreE:
```
# source code 
public static String random(int count, int start, int end, final boolean letters, final boolean numbers,
final char[] chars, final Random random) {
```
When a custom character array (chars != null) is supplied to RandomStringUtils.random(), the method does not strictly check that the start and end parameters fall within the valid bounds of the chars array.

As a result, if start or end exceeds chars.length, the method may generate a random index outside the array range, leading to an unexpected ArrayIndexOutOfBoundsException.

This fails the method contract and causes unpredictable runtime errors.

```
@Test
void testStartEndOutOfRangeWithChars() {
        char[] chars = {'a', 'b', 'c'};
        assertThrows(ArrayIndexOutOfBoundsException.class, () -> {
            RandomStringUtils.random(
                    5,
                    5,         // invalid: start > chars.length
                    10,             // invalid: end > chars.length
                    false,
                    false,
                    chars,
                    new Random()
            );
        });
}

```

Actual:
Throws ArrayIndexOutOfBoundsException

Expected:
Throw IllegalArgumentException indicating invalid start/end range when chars != null
[my issue](https://issues.apache.org/jira/browse/LANG-1801) @garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -278,6 +278,19 @@ public class RandomStringUtils {
                     "Parameter end (" + end + ") must be greater than start (" + start + ")");
         } else if (start < 0 || end < 0) {
             throw new IllegalArgumentException("Character positions MUST be >= 0");
+        }else if (chars != null && (start >= chars.length || end > chars.length)){
+            StringBuilder errorMsg = new StringBuilder();
+            int charsLength = chars.length;
+            if (start >= charsLength) {
+                errorMsg.append("Parameter start (").append(start).append(") must be less than chars array length ").append(charsLength);
+            }
+            if (end > charsLength) {
+                if (errorMsg.length() > 0) {
+                    errorMsg.append("; ");
+                }
+                errorMsg.append("Parameter end (").append(end).append(") must be less than or equal to chars array length ").append(charsLength);
+            }
+            throw new IllegalArgumentException(errorMsg.toString());
         }
 
         if (end > Character.MAX_CODE_POINT) {
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -278,9 +278,9 @@ public class RandomStringUtils {
                     "Parameter end (" + end + ") must be greater than start (" + start + ")");
         } else if (start < 0 || end < 0) {
             throw new IllegalArgumentException("Character positions MUST be >= 0");
-        }else if (chars != null && (start >= chars.length || end > chars.length)){
-            StringBuilder errorMsg = new StringBuilder();
-            int charsLength = chars.length;
+        } else if (chars != null && (start >= chars.length || end > chars.length)) {
+            final StringBuilder errorMsg = new StringBuilder();
+            final int charsLength = chars.length;
             if (start >= charsLength) {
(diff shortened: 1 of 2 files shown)


theodoral22:
Hi maintainers,

I have prepared a fix to the problem of the method RandomStringUtils.random(), that:

1) Adds proper validation to ensure start < chars.length and end <= chars.length.

2) Throws an IllegalArgumentException when the bounds are invalid.

3) Includes unit tests to cover these cases.

Would you be open to accepting a pull request with this fix?

Thank you for your time and for maintaining this excellent library!

Best regards,
Theodora Anastasia Lazaridou

IcoreE:
Hello @theodoral22 
Thank you for your attention to this issue. I have submitted the code and unit tests to fix it.[ PR#1521](https://github.com/apache/commons-lang/pull/1521)

IcoreE:
Hi @garydgregory 
I hope everything is going well on your end! It’s been a little while since I submitted PR (#1521), and I just wanted to follow up in case you’ve had a chance to take a look. If you have any feedback or if there’s anything I can improve, I’d be happy to make the necessary changes. Thanks again for all the work you put into the project!

garydgregory:
Hello @IcoreE 
There are higher priority issue I want to deal with.

src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
```diff
@@ -104,6 +104,7 @@ void testExceptionsRandom() {
         assertIllegalArgumentException(() -> RandomStringUtils.random(8, 32, 48, false, true));
         assertIllegalArgumentException(() -> RandomStringUtils.random(8, 32, 65, true, false));
         assertIllegalArgumentException(() -> RandomStringUtils.random(1, Integer.MIN_VALUE, -10, false, false, null));
+        assertIllegalArgumentException(() -> RandomStringUtils.random(2, 5, 6, false, false, new char[] { 'a', 'b', 'c', 'd' }, new Random()));
```
sebbASF: The test should cover checks for end and start both separately and together.
Also should check that the correct messages are generated.

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
garydgregory, line 251: Hello @IcoreE 
So is this Javadoc still relevant?

src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -278,6 +278,19 @@ public static String random(int count, int start, int end, final boolean letters
                     "Parameter end (" + end + ") must be greater than start (" + start + ")");
         } else if (start < 0 || end < 0) {
             throw new IllegalArgumentException("Character positions MUST be >= 0");
+        } else if (chars != null && (start >= chars.length || end > chars.length)) {
+            final StringBuilder errorMsg = new StringBuilder();
```
garydgregory, line 282: This is all too complicated IMO, we don't gather all possible invalid inputs elsewhere in the method. Please rebase on git master, fail-fast and use String.format(), so that would end up being 2 throws.

garydgregory (changes requested):
Hello @IcoreE 
Please see my comments.

garydgregory:
No reply, closing in favor of https://github.com/apache/commons-lang/pull/1682

PR #1523, opened 2025-12-14
Fix RandomStringUtils.random bounds validation

theodoral22:
This fixes an issue where start/end were not strictly validated when chars is not null, potentially causing IndexOutOfBoundsException.

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/RandomStringUtils.java
@@ -264,6 +264,19 @@ public class RandomStringUtils {
             throw new IllegalArgumentException("The chars array must not be empty");
         }
 
+        if (chars!=null) {
+            if (start>=chars.length) {
+                throw new IllegalArgumentException(
+                        "Parameter start (" +start+ ") must be less than chars.length ("+chars.length+")"
+                );
+            }
+            if (end>chars.length) {
+                throw new IllegalArgumentException(
+                        "Parameter end ("+end+") must be less than or equal to chars.length ("+chars.length+")"
+                );
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/RandomStringUtils.java
```diff
@@ -264,6 +264,19 @@ public static String random(int count, int start, int end, final boolean letters
             throw new IllegalArgumentException("The chars array must not be empty");
         }
 
+        if (chars!=null) {
```
ThrawnCA, line 267: Lots of missing spaces in this code block. There should be spaces before and after operators, as per the style used in the rest of this file, eg this line should be `if (chars != null) {`

src/test/java/org/apache/commons/lang3/RandomStringUtilsTest.java
```diff
@@ -815,4 +811,22 @@ void testRandomWithChars(final RandomStringUtils rsu) {
...
         assertNotEquals(r2, r3);
     }
+
+    @Test
+    void testStartEndOutOfRangeWithChars_shouldThrow() {
+        final char[] chars = { 'a', 'b', 'c' };
+        final Random random = new Random();
+        IllegalArgumentException ex;
+
+        ex=assertThrows(IllegalArgumentException.class, () ->
```
ThrawnCA, line 821: Missing spaces before and after operators.

garydgregory:
Closing, no reply, favor https://github.com/apache/commons-lang/pull/1682

PR #1557, opened 2026-01-09
LANG-1711: Skip inaccessible fields in ReflectionDiffBuilder

JeremySun0731:
This change makes ReflectionDiffBuilder skip fields that cannot be accessed via reflection
instead of failing fast with an exception.

Previously, if reflective access to a field was denied (for example due to Java 9+ module
restrictions or security constraints), the diff process would abort entirely. This prevented
producing a partial diff even when most fields were accessible.

The implementation is limited to appendFields(...), where reflection access failures are
caught and ignored. This covers IllegalAccessException on Java 8 as well as runtime reflection
failures on newer Java versions. No public APIs were changed.

This keeps full Java 8 compatibility while improving robustness on restricted environments.

Related issue: LANG-1711

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/builder/ReflectionDiffBuilder.java
@@ -204,11 +204,18 @@ public class ReflectionDiffBuilder<T> implements Builder<DiffResult<T>> {
         for (final Field field : FieldUtils.getAllFields(clazz)) {
             if (accept(field)) {
                 try {
-                    diffBuilder.append(field.getName(), readField(field, getLeft()), readField(field, getRight()));
-                } catch (final IllegalAccessException e) {
+                    final Object leftValue = readField(field,getLeft());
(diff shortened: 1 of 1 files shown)


garydgregory:
Hello @JeremySun0731 

- -1: There are ZERO tests in this PR.
- I think we need a comprehensive solution/proposal for the whole package, not a one-off.
- Please see and test PR #1558 for your use case. This is a new _opt-in_ feature for compatibility.

JeremySun0731:
Hi @garydgregory,

Thanks for the feedback and for pointing me to PR #1558.

I understand that a localized change in ReflectionDiffBuilder is not sufficient and that a package-wide, opt-in solution is preferred.

I’ll take a look at #1558 and test it against the LANG-1711 scenario.

Thanks for the guidance.

garydgregory:
@JeremySun0731 

I updated and merged PR #1558. Would you please check your use case?

garydgregory:
Closing in favor of https://github.com/apache/commons-lang/pull/1558

PR #1558, opened 2026-01-09
Feature : Make object accessibility optional

garydgregory:
See also https://issues.apache.org/jira/browse/LANG-1711

- [No] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?

### DIFF
--- a/src/main/java/org/apache/commons/lang3/builder/AbstractReflection.java
+++ b/src/main/java/org/apache/commons/lang3/builder/AbstractReflection.java
@@ -1,0 +1,171 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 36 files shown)


JeremySun0731:
Hi Gary,

I tried running `mvn test` on PR #1558 using IntelliJ’s bundled Maven and hit a Maven Enforcer failure:

Detected Maven Version: 3.6.3  
Required: Maven >= 3.8.1

This seems to be due to the IDE default Maven version rather than the code itself.  
I haven’t upgraded Maven locally yet, but wanted to report this environment behavior in case it’s relevant for contributors using IntelliJ defaults.

Happy to rerun tests with a newer Maven version if needed.

garydgregory:
That's correct, you need a recent version of Maven, not an ancient one 😉

JeremySun0731:
While reviewing PR #1558 with LANG-1711 in mind, I noticed that not all reflective
access paths appear to be centralized in AbstractReflection yet.

In particular, ReflectionDiffBuilder still uses
FieldUtils.readField(field, target, true), which forces accessibility unconditionally
and bypasses the accessibleFlag mechanism introduced here.

This means the new opt-in accessibility strategy does not currently apply
to ReflectionDiffBuilder in the same way as ReflectionToStringBuilder and others.

PR #1587, opened 2026-01-26
[LANG-1815] AnnotationUtils.equals fails for package-private annotations

inponomarev:
# Problem

AnnotationUtils.equals(Annotation, Annotation) may incorrectly return false for equal annotations when the annotation type or its members are not reflectively accessible from the caller’s package.

The root cause is twofold:

* Annotation member methods are invoked reflectively without ensuring accessibility, which can trigger access-related ReflectiveOperationExceptions (e.g. for package-private annotations accessed from outside their package).
* `ReflectiveOperationExceptions` are swallowed and treated as “not equal” by returning `false`, which hides reflection failure.

This behavior violates the method’s documented intent to implement the equality semantics defined by `Annotation.equals(Object)` and makes the result depend on reflective accessibility rather than annotation content.

This issue is tracked as [LANG-1815](https://issues.apache.org/jira/browse/LANG-1815).

### DIFF
--- a/src/main/java/org/apache/commons/lang3/AnnotationUtils.java
+++ b/src/main/java/org/apache/commons/lang3/AnnotationUtils.java
@@ -212,7 +212,9 @@ public class AnnotationUtils {
             for (final Method m : type1.getDeclaredMethods()) {
                 if (m.getParameterTypes().length == 0
                         && isValidAnnotationMemberType(m.getReturnType())) {
-                    m.setAccessible(true);
+                    if (!m.isAccessible()) {
+                        m.setAccessible(true);
+                    }
                     final Object v1 = m.invoke(a1);
                     final Object v2 = m.invoke(a2);
                     if (!memberEquals(m.getReturnType(), v1, v2)) {
--- a/src/main/java/org/apache/commons/lang3/AnnotationUtils.java
+++ b/src/main/java/org/apache/commons/lang3/AnnotationUtils.java
@@ -212,6 +212,7 @@ public class AnnotationUtils {
             for (final Method m : type1.getDeclaredMethods()) {
                 if (m.getParameterTypes().length == 0
                         && isValidAnnotationMemberType(m.getReturnType())) {
+                    m.setAccessible(true);
                     final Object v1 = m.invoke(a1);
                     final Object v2 = m.invoke(a2);
                     if (!memberEquals(m.getReturnType(), v1, v2)) {
@@ -220,7 +221,7 @@ public class AnnotationUtils {
                 }
             }
         } catch (final ReflectiveOperationException ex) {
-            return false;
+            throw new IllegalStateException(ex);
         }
         return true;
     }
--- a/src/test/java/org/apache/commons/lang3/external/AnnotationEqualsTest.java
+++ b/src/test/java/org/apache/commons/lang3/external/AnnotationEqualsTest.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/lang3/external/AnnotationEqualsTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Tag("value")
+    private final Object a = new Object();
+    @Tag("value")
+    private final Object b = new Object();
+
+    @Test
+    void equalsWorksOnPackagePrivateAnnotations() throws Exception {
+        Tag tagA = getClass().getDeclaredField("a").getAnnotation(Tag.class);
+        Tag tagB = getClass().getDeclaredField("b").getAnnotation(Tag.class);
+        Assertions.assertTrue(AnnotationUtils.equals(tagA, tagB));
```
garydgregory: We use static imports for JUnit `Assertions` (and only other JUnit classes).

src/test/java/org/apache/commons/lang3/external/AnnotationEqualsTest.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.commons.lang3.external;
+
+import org.apache.commons.lang3.AnnotationUtils;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+// CAUTION: in order to reproduce https://issues.apache.org/jira/browse/LANG-1815,
```
garydgregory: This can be a Javadoc

src/main/java/org/apache/commons/lang3/AnnotationUtils.java
```diff
@@ -220,7 +221,7 @@ && isValidAnnotationMemberType(m.getReturnType())) {
                 }
             }
         } catch (final ReflectiveOperationException ex) {
-            return false;
+            throw new IllegalStateException(ex);
```
garydgregory, line 226: This behavior has changed, but there is no new test to avoid a future regression.

src/main/java/org/apache/commons/lang3/AnnotationUtils.java
```diff
@@ -212,6 +212,7 @@ public static boolean equals(final Annotation a1, final Annotation a2) {
             for (final Method m : type1.getDeclaredMethods()) {
                 if (m.getParameterTypes().length == 0
                         && isValidAnnotationMemberType(m.getReturnType())) {
+                    m.setAccessible(true);
```
garydgregory: You'll want to call `isAccessible()` and only call `setAccessible()` if required, which will avoid calling the permission checks in `setAccessible()`.

garydgregory (commented):
Hello @inponomarev 

Thank you for the PR.

I added a few comments scattered throughout.

src/test/java/org/apache/commons/lang3/external/AnnotationEqualsTest.java
inponomarev: fixed

src/main/java/org/apache/commons/lang3/AnnotationUtils.java
inponomarev: fixed, thanks for spotting this -- now this branch is covered by a test

src/main/java/org/apache/commons/lang3/AnnotationUtils.java
inponomarev: fixed

garydgregory:
Hello @inponomarev 

I need to think some more about https://github.com/apache/commons-lang/pull/1558 before adding more calls to `setAccessible`, which is problematic for some people. If https://github.com/apache/commons-lang/pull/1558 goes in, then this PR can reuse the optional mechanism... or not. I'm not sure yet. You're welcome to have a look at https://github.com/apache/commons-lang/pull/1558

garydgregory:
Hello @inponomarev 
Finally coming back around to this PR. Please rebase on git master and reuse the facilities introduced in https://github.com/apache/commons-lang/pull/1558.
Thank you!

garydgregory:
@inponomarev 🔔 Ping.

garydgregory:
@inponomarev 🔔 Ping.

kzhunmax:
@garydgregory  Since the original PR appears inactive after follow-up pings, I'd be happy to work on this if the original author is no longer available.

garydgregory:
@kzhunmax Sure, go for it 😄

garydgregory:
@inponomarev CC @kzhunmax 
The new proposal is https://github.com/apache/commons-lang/pull/1760

PR #1626, opened 2026-04-15
NumberUtils.isCreatable(String) should match NumberUtils.createNumber(String), exactly

garydgregory:
`NumberUtils.isCreatable(String)` should match `NumberUtils.createNumber(String)`, exactly

Add tests from PR #1623


chtompki:
Looks good to me

PR #1644, opened 2026-05-15
StringUtils.repeat(String, [String,] int) now throws IllegalArgumentException instead of NegativeArraySizeException.

garydgregory:
StringUtils.repeat(String, [String,] int) now throws IllegalArgumentException instead of NegativeArraySizeException.

- [ ] ~~[I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?~~

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -351,7 +351,7 @@ public class StringUtils {
      * @throws IllegalArgumentException if the width is too small.
      * @since 3.6
      */
-    public static String abbreviate(final String str, String abbrevMarker, int offset, final int maxWidth) {
+    public static String abbreviate(final String str, String abbrevMarker, final int offset, final int maxWidth) {
         if (isEmpty(str)) {
             return str;
         }
@@ -6117,7 +6117,12 @@ public class StringUtils {
         if (inputLength == 1 && count <= PAD_LIMIT) {
             return repeat(repeat.charAt(0), count);
         }
-        final int outputLength = inputLength * count;
+        final int outputLength;
+        try {
+            outputLength = Math.multiplyExact(inputLength, count);
+        } catch (final Exception e) {
+            throw new IllegalArgumentException("The requested result is too large for a String.");
+        }
         switch (inputLength) {
         case 1:
             return repeat(repeat.charAt(0), count);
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/StringUtils.java
```diff
@@ -6117,7 +6117,12 @@ public static String repeat(final String repeat, final int count) {
         if (inputLength == 1 && count <= PAD_LIMIT) {
             return repeat(repeat.charAt(0), count);
         }
-        final int outputLength = inputLength * count;
+        final int outputLength;
+        try {
+            outputLength = Math.multiplyExact(inputLength, count);
+        } catch (final Exception e) {
+            throw new IllegalArgumentException("Count too large for input");
```
ppkarwasz: The `String.repeat` method throws `OutOfMemoryError("Required length exceeds implementation limit")`, should we align?

garydgregory: Interesting suggestion, I think the JRE made a mistake here and didn't read its own spec (see below).

`OutOfMemoryError` is an bad choice IMO because: 

- We've not actually run out of memory, we are preempting execution, we detected bad input that could be recoverable by an app. For example: ask the user to pick a smaller file, a smaller this, a smaller that.
- The PR converts a runtime exception into another runtime exception.
- Your suggestion would convert a runtime exception into an error. Now an app should catch `IllegalArgumentException` and `OutOfMemoryError` to handle bad inputs. It is unusual for apps to catch errors.
- How does an app detect the difference b/w actually running out of memory and bad input?
- `OutOfMemoryError` says: "Thrown when the Java Virtual Machine cannot allocate an object because it is out of memory, and no more memory could be made available by the garbage collector."
That's definitely _not_ what happened here, or in `String.repeat(int).`
- `OutOfMemoryError` is a subclass of `VirtualMachineError` which says: "Thrown to indicate that the Java Virtual Machine is broken or has run out of resources necessary for it to continue operating."

Note the "for it to continue operating". If we throw OOME, we are saying the JVM cannot continue operating.

The message here could be better, it is now: "The requested result is too large for a String." Better?

ppkarwasz: I agree that `OutOfMemoryError` is the wrong choice, but the difference prevents a smooth migration between `StringUtils` and newly introduced `String` methods. Maybe this is something worth documenting to explain users why they should not replace this method with `String.repeat`?

garydgregory: If you agree it's wrong, then why did you propose it in the first place?! Why did I just waste an hour?! :(

If we start documenting everything that's different from the JRE, we'll never end, and confuse users. Just imagine, every time a new Java version comes out, we'd document all the little differences with our APIs, yikes...

ppkarwasz: I believe this is a detail worth discussing. `OutOfMemoryError` might be an unfortunate choice, so we need to make a compromise between a more fitting exception and coherence with the JDK. We chose the former, but we did that choice consciously.

PR #1649, opened 2026-05-17
WordUtils.wrap() zero-width wrapOn regex match causes infinite loop

garydgregory:
Calling the deprecated method `WordUtils.wrap(String, int, String, boolean, String)` with zero-width wrapOn regex match causes infinite loop

- [x] OP used AI to create the original report.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -652,21 +652,19 @@ public class WordUtils {
                 str.substring(offset, Math.min((int) Math.min(Integer.MAX_VALUE, offset + wrapLength + 1L), inputLineLength)));
             if (matcher.find()) {
                 if (matcher.start() == 0) {
-                    offset += matcher.end();
+                    // If the match is zero-width, advance by at least 1 to avoid infinite loop.
+                    offset += matcher.end() > 0 ? matcher.end() : 1;
                     continue;
                 }
                 spaceToWrapAt = matcher.start() + offset;
             }
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/text/WordUtils.java
```diff
@@ -652,21 +652,19 @@ public static String wrap(final String str, int wrapLength, String newLineStr, f
                 str.substring(offset, Math.min((int) Math.min(Integer.MAX_VALUE, offset + wrapLength + 1L), inputLineLength)));
             if (matcher.find()) {
                 if (matcher.start() == 0) {
-                    offset += matcher.end();
+                    // If the match is zero-width, advance by at least 1 to avoid infinite loop.
+                    offset += matcher.end() > 0 ? matcher.end() : 1;
```
ppkarwasz, line 656: This stops the infinite loop, but drops the first character of the line.

Fixing this would require fixing a couple of other lines that assume matches are always of length 1. I am not sure if it is worth fixing in a **deprecated** method, but we should add some tests for variable-length matches in Commons Text.

garydgregory: Hello @ppkarwasz 
Good find! Do you have a unit test we can add for that and fix?

ppkarwasz: I'll make a PR.

ppkarwasz: #1655

PR #1654, opened 2026-05-17
StrBuilder.deleteImpl(int, int, int) doesn't clear its unused bytes.

garydgregory:
`StrBuilder.deleteImpl(int, int, int)` doesn't clear its unused bytes.

- [x] OP used AI to create the original report.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/text/StrBuilderClearTest.java
+++ b/src/test/java/org/apache/commons/lang3/text/StrBuilderClearTest.java
@@ -19,9 +19,14 @@ package org.apache.commons.lang3.text;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.io.Reader;
+import java.lang.reflect.Field;
+import java.nio.charset.StandardCharsets;
 
+import org.apache.commons.lang3.SerializationUtils;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/text/StrBuilderClearTest.java
```diff
@@ -86,6 +91,40 @@ public int read(final char[] cbuf, final int off, final int len) {
...
+        try (ObjectInputStream ois = new ObjectInputStream(new ByteArrayInputStream(serialized))) {
+            sb2 = (StrBuilder) ois.readObject();
+        }
+        final Field bufField = StrBuilder.class.getDeclaredField("buffer");
+        bufField.setAccessible(true);
+        final Field sizeField = StrBuilder.class.getDeclaredField("size");
+        sizeField.setAccessible(true);
+        final char[] buf2 = (char[]) bufField.get(sb2);
+        final String bufContent = new String(buf2);
+        assertFalse(bufContent.contains("secret_password"), "Deserialized StrBuilder buffer must not contain stale chars: " + bufContent);
```
ppkarwasz: `buffer` and `size` are `protected`, so there is no need for reflection.

ppkarwasz (approved):

src/test/java/org/apache/commons/lang3/text/StrBuilderClearTest.java
garydgregory: Fixed in commit 498c26b5f40598b484310f16cd86d78c57d5bc30.

ppkarwasz: Thanks!

PR #1655, opened 2026-05-17
Fix `WordUtils.wrap` for `wrapOn` matches different from one character

ppkarwasz:
`WordUtils.wrap` advanced the offset by a single character past each `wrapOn` match, assuming the separator is always exactly one character long:

 - When the `wrapOn` regex matched several characters, the surplus separator characters were left in the wrapped output instead of being consumed,
 - When the `wrapOn` regex matched zero characters, one additional character was consumed.

Now it tracks the end offset of each match and resume from there, so that separators of any length, including zero-width matches, are skipped correctly.

- [x] I used AI to refactor tests and reword descriptions


garydgregory:
@ppkarwasz 
Please stop creating branches in this repo, use your fork. It's too easy to have orphan branches.

ppkarwasz:
@garydgregory, no problem. I saw your message on the mailing list ([thread](https://lists.apache.org/thread/qx8x85wc4h3pndfs26cwj7qvx0dd1v3r)) and will submit PRs from my fork going forward.

PR #1684, opened 2026-06-02
Fix CharSequenceUtils.lastIndexOf for supplementary code points

alhudz:
CharSequenceUtils.lastIndexOf treats a start index equal to the last char as no match for supplementary code points, so StringUtils.lastIndexOf(seq, codePoint) on a StringBuilder or other non-String CharSequence always returns -1 even when the code point is present, because the default overload passes startPos equal to the length. A supplementary code point spans two chars and its high surrogate can begin no later than sz-2, so the search origin should be clamped there rather than bailing out. With the clamp the results match java.lang.String, as the added assertions verify.


garydgregory:
TY @alhudz , merged.

PR #1687, opened 2026-06-03
Fix indexOfAny matching an unpaired trailing high surrogate

alhudz:
indexOfAny(CharSequence, int, char...) reports a hit when the last char of the input is a lone high surrogate that equals the high half of a supplementary code point in searchChars: the surrogate guard short-circuits on `i >= csLast` and returns that index without ever matching the low half. So `indexOfAny("\uD840", new char[] {'\uD840', '\uDC01'})` returns 0 while `containsAny`/`containsNone` return false on the same input, even though the StringUtilsContainsTest cases pin that lone-half behavior. Reused the existing containsAny condition so the three methods agree.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -2815,8 +2815,7 @@ public class StringUtils {
             final char ch = cs.charAt(i);
             for (int j = 0; j < searchLen; j++) {
                 if (searchChars[j] == ch) {
-                    // ch is a supplementary character
-                    if (i >= csLast || j >= searchLast || !Character.isHighSurrogate(ch) || searchChars[j + 1] == cs.charAt(i + 1)) {
+                    if (!Character.isHighSurrogate(ch) || j == searchLast || i < csLast && searchChars[j + 1] == cs.charAt(i + 1)) {
                         return i;
                     }
                 }
--- a/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsEqualsIndexOfTest.java
@@ -411,6 +411,10 @@ class StringUtilsEqualsIndexOfTest extends AbstractLangTest {
         assertEquals(2, StringUtils.indexOfAny(CharU20000 + CharU20001, CharU20001.toCharArray()));
         assertEquals(0, StringUtils.indexOfAny(CharU20000, CharU20000.toCharArray()));
         assertEquals(-1, StringUtils.indexOfAny(CharU20000, CharU20001.toCharArray()));
+        // An unpaired trailing high surrogate must not match a supplementary code point, matching containsAny.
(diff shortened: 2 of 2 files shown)


garydgregory:
Hello @alhudz 

Thank you for your PR.

We have a LOT of String code in this class and in Lang in general. Do you have a sense of how pervasive this type of issue is in the code base?

alhudz:
Not endemic, from what I can tell. Most of the `String`/`char` code in Lang operates on `char`s by design and is documented that way, so it's fine. The bug class I've been hitting is narrower: a method that scans by `char` index but then exposes a *code-point* count or contract at its boundary, so a supplementary character throws the count off by one.

I've found three of those seams so far, each with a reproducer:
- `CharSequenceUtils.lastIndexOf` (#1684, merged)
- `StringUtils.indexOfAny` (this one)
- `LookupTranslator.translate` (#1691) — returned the matched key length in `char`s where the translator loop advances by `Character.charCount`

The way I find them is to look for places that cross between char-indexed scanning and a code-point boundary, then build a supplementary-key case and check the count. I haven't done an exhaustive sweep of the whole class, so I can't promise these are the last three, but they're the ones I could actually reproduce rather than guess at.

garydgregory:
Hello @alhudz 

Thank you for your prompt reply.

I would prefer to get a better sense of what we are looking at for the whole class or ideally the whole library. What I would prefer to avoid is an inconsistent treatment of this issue in the library.

I am planning for a release candidate fairly soon, so I would like to get an idea of whether I should hold off and get the set of PRs in or wait to do all these in a further maintenance release.

Any further analysis you can provide here is greatly appreciated while I'm off fighting other fires ;)

alhudz:
Went back and swept the whole class (plus the `text.translate` package) for the same seam rather than eyeballing it.

The only family in `StringUtils` that does a char-by-char surrogate-pair guard is `containsAny`/`containsNone`/`indexOfAny`/`indexOfAnyBut`. Three of those already agree: `containsAny` and `containsNone` use the canonical guard, and `indexOfAnyBut` routes through a `codePointAt`/`charCount` loop. The lone outlier is `indexOfAny(CharSequence, int, char...)`, which this PR brings into line.

Everything else in the class that touches surrogates already advances by `Character.charCount` off a `codePointAt` (`swapCase`/`replaceChars` around line 8661, the `firstCodePoint` sites), so it's code-point correct by construction. Same story across `text.translate`: `CharSequenceTranslator`, `CodePointTranslator`, `JavaUnicodeEscaper` and `NumericEntityUnescaper` all advance by `charCount`. The one exception there is `LookupTranslator`, which is #1691.

So the count is bounded rather than a swarm: the three seams I listed (#1684 merged, this one, #1691), each with a reproducer. I can't promise some other char-array method elsewhere in Lang won't share the shape, but within the String and translate code where it'd actually bite, those are the three. Small enough to batch into the RC if you'd rather not carry them to a later maintenance release.

garydgregory:
@alhudz 
Thank you for taking another pass.👏

alhudz:
Thank you for the review and the merge. I appreciate the feedback and the opportunity to contribute.

PR #1691, opened 2026-06-06
Fix LookupTranslator code point consumption for supplementary keys

alhudz:
Repro: build a `LookupTranslator` whose key is a single supplementary code point, e.g. `U+1D54F` mapped to `X`, then `translate("𝕏Y")`.
Expected: `XY`. Actual: `X` — the trailing `Y` is silently dropped (likewise `a𝕏b` becomes `aX`).
Cause: `translate` returns the matched key length counted in `char`s (`i`), but `CharSequenceTranslator` documents the return value as a count of code points and advances the cursor by `Character.charCount` per unit. A supplementary key is two `char`s but one code point, so the loop steps twice and over-consumes the following character.
Fix: return `Character.codePointCount(input, index, index + i)` so the count matches the framework's code-point stepping. The other translators in the package already return code-point counts, so this is the only affected site.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/translate/LookupTranslator.java
+++ b/src/main/java/org/apache/commons/lang3/text/translate/LookupTranslator.java
@@ -86,7 +86,7 @@ public class LookupTranslator extends CharSequenceTranslator {
 
                 if (result != null) {
                     out.write(result);
-                    return i;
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz 
Please reply to https://github.com/apache/commons-lang/pull/1687#issuecomment-4636440513

PR #1693, opened 2026-06-08
Re-assert CharRange start <= end invariant in readObject

alhudz:
1. `CharRange` reverses reversed endpoints in its constructor, so a constructed instance always holds `start <= end`; with no `readObject`, default deserialisation restores the raw fields.
2. a crafted stream (reachable through a serialised `CharSet`, which holds `Set<CharRange>`) can carry `start > end`, after which `contains(char)` matches nothing for that range and the iterator walks it incorrectly.

Repro: reflectively set a `CharRange.isIn('a', 'e')` to `start='z'`/`end='a'`, serialise, then deserialise. Before: it round-trips into an inverted range. After: `InvalidObjectException`, in line with the recent `Range` (#1686) and `Fraction` (#1688) readObject hardening.

Re-asserted `start <= end` after `defaultReadObject()`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharRange.java
+++ b/src/main/java/org/apache/commons/lang3/CharRange.java
@@ -16,6 +16,9 @@
  */
 package org.apache.commons.lang3;
 
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz 
PR merged, TY.

alhudz:
Thanks! Appreciated

PR #1696, opened 2026-06-09
Correct JavaDoc in Strings.java concerning case-sensitivity

mfg92:
The JavaDocs of some methods mentioned case-sensitivity or case-insensitivity, even though they were part of the abstract Strings class. However, the behaviour of these methods can be case-sensitive **or** case-insensitive, depending on whether the Strings.CS or Strings.CI class is used.


garydgregory:
Thank you @mfg92 ! Merged 🚀

PR #1697, opened 2026-06-09
Fix case-insensitive Strings.replace dropping length-changing matches

alhudz:
Repro: `Strings.CI.replace("aİb", "İ", "X", -1)` returns `aİb` instead of `aXb` under any non-Turkish locale (`StringUtils.replaceIgnoreCase`/`replaceOnceIgnoreCase` delegate here, so they are affected too).
Cause: `replace` lower-cases the search argument up front, but the case-insensitive `indexOf` it calls already matches case-insensitively through `CharSequenceUtils.regionMatches`. `'İ'` (`U+0130`) lower-cases to the two-char `"i̇"`, so both the needle and the derived `replLength` grow to length 2 and stop lining up with the single source character, so the match is silently dropped.
Fix: remove the redundant lower-casing and let the case-insensitive `indexOf` do the matching. The case-sensitive path never lower-cased, so it is unchanged.

The regression test pins the locale with `@DefaultLocale("en")` so the lower-case expansion is deterministic; it fails on the current code and passes with the fix.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Strings.java
+++ b/src/main/java/org/apache/commons/lang3/Strings.java
@@ -1288,12 +1288,9 @@ public abstract class Strings {
      * @param max          maximum number of values to replace, or {@code -1} if no maximum
      * @return the text with any replacements processed, {@code null} if null String input
      */
-    public String replace(final String text, String searchString, final String replacement, int max) {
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , merged 🚀

PR #1700, opened 2026-06-11
[LANG-1827] Support escaped single quotes ('') in DurationFormatUtils format patterns

sarankumarbaskar:
DurationFormatUtils.lexx() does not handle escaped single quotes, making it
impossible to include a literal apostrophe in formatted duration output.

The root cause is in the lexx() parser: when it encounters a single quote
while inside a literal block, it unconditionally ends the literal. It never
checks whether the next character is also a single quote, which should mean
"emit a literal apostrophe and continue the literal."

This behavior contradicts the quoting convention established by
java.text.SimpleDateFormat, where '' consistently produces a literal
single quote. The class Javadoc acknowledged the limitation explicitly:

    "Note: It's not currently possible to include a single-quote in a format."

and a TODO in lexx() read:

    // TODO: Need to handle escaping of '

This PR modifies lexx() to recognise '' as an escaped single quote — both
inside and outside quoted literal sections — matching SimpleDateFormat
behavior. The Javadoc has been updated to document the '' convention and
the TODO has been removed.

Note: This is a minor behavioral change. Existing format strings that relied
on '' as "close one literal then open another" will now produce a literal
apostrophe instead. This aligns with what most users would expect.

This issue is tracked as[LANG-1827](https://issues.apache.org/jira/browse/LANG-1827).

- [x] Read the contribution guidelines for this project.
- [x] Read the ASF Generative Tooling Guidance.
- [x] Ran a successful build with `mvn`.
- [x] Unit tests cover all new behavior and verify no regressions.
- [x] Commit message follows the [LANG-XXXX] convention.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
@@ -461,7 +461,7 @@ class DurationFormatUtilsTest extends AbstractLangTest {
         cal.set(Calendar.MILLISECOND, 0);
         time = cal.getTime().getTime();
         assertEquals("40", DurationFormatUtils.formatPeriod(time1970, time, "yM"));
-        assertEquals("4 years 0 months", DurationFormatUtils.formatPeriod(time1970, time, "y' ''years' M 'months'"));
+        assertEquals("4 'years 0 months", DurationFormatUtils.formatPeriod(time1970, time, "y' ''years' M 'months'"));
         assertEquals("4 years 0 months", DurationFormatUtils.formatPeriod(time1970, time, "y' years 'M' months'"));
         assertEquals("4years 0months", DurationFormatUtils.formatPeriod(time1970, time, "y'years 'M'months'"));
         assertEquals("04/00", DurationFormatUtils.formatPeriod(time1970, time, "yy/MM"));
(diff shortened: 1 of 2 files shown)


garydgregory:
@ppkarwasz @chtompki 
Do you foresee any issues with this solution aside from the obvious change in behavior?

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
```diff
@@ -584,6 +584,24 @@ void testLANG815() {
     void testLANG981() { // unmatched quote char in lexx
         assertIllegalArgumentException(() -> DurationFormatUtils.lexx("'yMdHms''S"));
     }
+
+    @Test
+    void testLANG1827() {
```
garydgregory: Hello @Sarankumar18 
Please write this new test as a `@ParameterizedTest` so that we can run all input/output individually and see them all fail if the `main` side of the patch is not applied.

garydgregory (changes requested):
Hello @Sarankumar18 
Please see my comment.
Thank you!

src/test/java/org/apache/commons/lang3/time/DurationFormatUtilsTest.java
sarankumarbaskar: @garydgregory  I've made as parametrize test for testLANG1827

sarankumarbaskar: @garydgregory  Are we good to merge this PR ?

garydgregory:
Not yet, it's in my queue. Discussing and thinking about compatibility...

sarankumarbaskar:
Okay, Thanks Gary

garydgregory:
Thank you @sarankumarbaskar for your patience, merged 🚀

garydgregory:
Jira ticket https://issues.apache.org/jira/browse/LANG-1827

sarankumarbaskar:
> Thank you @sarankumarbaskar for your patience, merged 🚀

Thank you @garydgregory

PR #1702, opened 2026-06-13
Reject doubled leading sign in NumberUtils.createBigInteger

alhudz:
`NumberUtils.createBigInteger` peels off one optional leading `+`/`-` (into a `negate` flag) and an optional `0x`/`#`/`0` radix prefix, then passes the rest to `new BigInteger(String)`, which parses a sign of its own. A doubled sign slips through and is silently mis-valued: `createBigInteger("--1")` returns `1`, `"-+1"` and `"+-1"` return `-1`, `"+-10"` returns `-10`, `"--010"` returns `10`. Since `createNumber` falls back to `createBigInteger` for plain integer tokens, `createNumber("--1")` returns `1` and `isCreatable("--1")` reports `true`, even though `isParsable("--1")` is `false` and the sibling `createInteger`/`createLong` reject the input via `Integer.decode`/`Long.decode`.

`Integer.decode`/`Long.decode` already throw `Sign character in wrong position` when a sign is not at the start, so the same check goes into `createBigInteger` once the leading sign and radix prefix are consumed. Putting it in the parser lets `createNumber`, `isCreatable` and the deprecated `isNumber` agree without each adding its own guard; `createBigDecimal` already rejects these through `new BigDecimal`. The added assertions cover the `createBigInteger` failures and the `isCreatable`/`createNumber` chain, and they fail on the current code. The `math` unit tests pass with the patch applied.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
+++ b/src/main/java/org/apache/commons/lang3/math/NumberUtils.java
@@ -232,6 +232,11 @@ public class NumberUtils {
             radix = 8;
             pos++;
         } // default is to treat as decimal
+        if (str.startsWith("-", pos) || str.startsWith("+", pos)) {
+            // a second sign here (e.g. "--1") is not a number; new BigInteger(String) would otherwise
+            // consume it and silently flip the sign. Integer.decode/Long.decode reject this the same way.
+            throw new NumberFormatException("Sign character in wrong position");
+        }
         final BigInteger value = new BigInteger(str.substring(pos), radix);
         return negate ? value.negate() : value;
     }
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz 
Note that I updated the worker method to check for parity:
```java
    protected void testCreateBigIntegerFailure(final String str) {
        assertThrows(NumberFormatException.class, () -> NumberUtils.createBigInteger(str), "createBigInteger(\"" + str + "\") should have failed.");
        // Should match java.math.BigInteger.BigInteger(String)
        assertThrows(NumberFormatException.class, () -> new BigInteger(str));
    }
```

garydgregory:
@alhudz 
TY, merged 🚀

PR #1704, opened 2026-06-14
LANG-1032 - Added overloads for joining an array of booleans with a p…

CptBartender:
…rovided StringBuilder

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3838,7 +3838,38 @@ public class StringUtils {
         if (array == null) {
             return null;
         }
-        return join(array, delimiter, 0, array.length);
+        return join(array, delimiter, 0, array.length, null).toString();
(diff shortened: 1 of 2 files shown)


CptBartender:
I took a swing at one of the issues from Jira, and here's my proposed approach on how to tackle that.

My general idea is that the `stringBuilder` param is _null-safe_ and providing `null` there results in the same result as if the other overload was called instead. Also, contrary to the proposed approach in the ticket, I'm suggesting we return the `StringBuilder` instance so that we can fluently use it, if needed.

I've done the overloads only for `boolean[]`  joining, as I didn't want to commit too much time if this approach is not deemed acceptable - once confirmed, I'm happy to provide similar overloads for other cases.

garydgregory:
Jira https://issues.apache.org/jira/browse/LANG-1032

garydgregory:
This feels like a bit of a YAGNI to me, see also my comment in the Jira ticket.

PR #1705, opened 2026-06-14
Fix: replace String == comparison with .equals() in ObjectToStringCom…

Jaderccesar:
## Problem
SpotBugs reports ES_COMPARING_STRINGS_WITH_EQ in ObjectToStringComparator.compare().
The original code compared String objects using `==`, which checks reference equality
rather than content equality. This can cause incorrect comparisons when two distinct
String objects have the same content.

## Solution
- Removed the `if (string1 == string2)` reference comparison
- Added null checks before content comparison (to avoid NullPointerException)
- Used `string1.equals(string2)` for correct content-based comparison

## Testing
Existing tests in ObjectToStringComparatorTest cover this method.
No new behavior is introduced, only the comparison mechanism is corrected.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/compare/ObjectToStringComparator.java
+++ b/src/main/java/org/apache/commons/lang3/compare/ObjectToStringComparator.java
@@ -64,15 +64,16 @@ public final class ObjectToStringComparator implements Comparator<Object>, Seria
         }
         final String string1 = o1.toString();
         final String string2 = o2.toString();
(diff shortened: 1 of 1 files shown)


sebbASF:
Whilst it is true that using == to compare strings is generally incorrect, in this case it is essential to ensure that nulls are handled correctly. If both objects return null for toString(), the result should be 0, but it will be 1.

Also, the equals comparison at line 74 is redundant.

PR #1706, opened 2026-06-14
Fix: replace ThreadLocalRandom with SecureRandom in ArrayUtils shuffl…

lendrop:
Replace ThreadLocalRandom with SecureRandom in ArrayUtils internal random() method.

The internal random() method previously returned ThreadLocalRandom.current(), which 
produces predictable sequences and is not suitable for security-sensitive contexts.


garydgregory:
-1 and closing: Dangerous PR that breaks documented behavior and kills performance.

PR #1707, opened 2026-06-15
Fix substringAfter(Last) for supplementary code points

alhudz:
Repro: `StringUtils.substringAfter("a" + new String(Character.toChars(0x1F600)) + "b", 0x1F600)` returns `"\uDE00b"` (a dangling low surrogate) instead of `"b"`; `substringAfterLast(prefix + grin, 0x1F600)` returns `"\uDE00"` instead of `""`.
Cause: the code-point overloads locate `find` with `indexOf`/`lastIndexOf`, which return the index of the high surrogate for a supplementary code point, then step over the match by a fixed `+ 1`. `substringAfterLast` also tests its terminal case with `pos == length() - 1`, so when the supplementary `find` is the last code point the guard never fires.
Fix: advance by `Character.charCount(find)` and test the terminal guard against `length() - charCount`. BMP code points keep `charCount == 1` so existing behaviour is unchanged, and `substringBefore(String, int)` already slices at `pos` so it needs no change.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8260,7 +8260,7 @@ public class StringUtils {
         if (pos == INDEX_NOT_FOUND) {
             return EMPTY;
         }
-        return str.substring(pos + 1);
+        return str.substring(pos + Character.charCount(find));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
TY @alhudz , merged 🚀

PR #1708, opened 2026-06-16
Fix CT_CONSTRUCTOR_THROW: move validation from constructor to factory…

daniellcoellho:
Fix CT_CONSTRUCTOR_THROW (BAD_PRACTICE) in Range constructor

SpotBugs detected CT_CONSTRUCTOR_THROW in Range: when a constructor
throws an exception before completing initialization, the partially
constructed object may be exploited via finalizer attacks.

This fix moves the null validation of fromInclusive and toInclusive
from the Range constructor to the static factory method
of(T, T, Comparator), which is the single entry point for all other
factory methods (of, is, between). The constructor no longer throws
exceptions.

Unit tests were added to RangeTest to verify that NullPointerException
is thrown from the factory method for null arguments.

- [x] Read the contribution guidelines.
- [x] Read the ASF Generative Tooling Guidance.
- [x] I used AI (Claude by Anthropic) to assist in identifying and implementing this fix.
- [x] Run a successful build with `mvn`.
- [x] Unit tests added to RangeTest.
- [x] Pull request description written above.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Range.java
+++ b/src/main/java/org/apache/commons/lang3/Range.java
@@ -189,6 +189,8 @@ public class Range<T> implements Serializable {
      * @since 3.13.0
      */
     public static <T> Range<T> of(final T fromInclusive, final T toInclusive, final Comparator<T> comparator) {
+        Objects.requireNonNull(fromInclusive, "fromInclusive");
+        Objects.requireNonNull(toInclusive, "toInclusive");
         return new Range<>(fromInclusive, toInclusive, comparator);
     }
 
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @daniellcoellho 
-1 because:
- We explicitly ignore this SB issue in our SB configuration.
- The unit tests pass when the `main` side of the patch is not applied. IOW, the test patch doesn't prove anything's changed.
- If we decide to fix this issue, they should be fix all 15 of them, not a random one.

sarankumarbaskar:
Thanks for clarifying. I see this is already intentionally excluded in the SpotBugs filter, so fixing one instance would not be useful without a broader decision to handle all CT_CONSTRUCTOR_THROW cases consistently.

garydgregory:
Closing, see above.

PR #1709, opened 2026-06-16
Fix spurious overflow in Fraction.add/subtract for coprime denominators

alhudz:

---

`Fraction.add`/`subtract` go through `addSub`, whose `d1 == 1` fast path (coprime denominators) builds the result numerator with `mulAndCheck(numerator, fraction.denominator)` and `mulAndCheck(fraction.numerator, denominator)`. Those cross products overflow an `int` even when the reduced result is comfortably in range, so `Fraction.getFraction(Integer.MAX_VALUE, 2).add(Fraction.getFraction(-Integer.MAX_VALUE, 1))` throws `ArithmeticException: overflow: mul`. Expected `-2147483647/2`, actual an exception. The sibling `d1 != 1` branch already computes the numerator in wider precision and only rejects the final reduced value.

Compute the two cross products and their sum/difference in `long` (both denominators are positive and below `2^31`, so the intermediate values cannot overflow `long`) and reject only when the resulting numerator does not fit an `int`, matching the wider branch. Keeping it inside `addSub` covers `add`, `subtract` and `divideBy`. The two helpers `addAndCheck`/`subAndCheck` were only used by this branch and are removed. Regression cases added to `FractionTest` fail with `overflow: mul` before the change.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -528,15 +528,12 @@ public final class Fraction extends Number implements Comparable<Fraction> {
         final int d1 = greatestCommonDivisor(denominator, fraction.denominator);
         if (d1 == 1) {
             // result is ((u*v' +/- u'v) / u'v')
-            // both denominators are positive and below 2^31, so the cross products and their
-            // sum stay within long range; only the reduced numerator has to fit in an int.
-            final long uvp = (long) numerator * (long) fraction.denominator;
-            final long upv = (long) fraction.numerator * (long) denominator;
-            final long t = isAdd ? uvp + upv : uvp - upv;
-            if (t < Integer.MIN_VALUE || t > Integer.MAX_VALUE) {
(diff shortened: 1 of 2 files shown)


garydgregory (commented):
Hello @alhudz 

Isn't there a way to do this using Java's own `java.lang.Math.addExact(int, int)` and `java.lang.Math.subtractExact(int, int)`, and so on?

alhudz:
Makes sense, done. Used `Math.toIntExact` for the final numerator and `Math.addExact`/`Math.subtractExact` for the sum.

The one catch is the `(int, int)` variants can't sit on the cross products directly: `u*v'` and `u'*v` overflow an `int` even when the reduced result fits, which is the actual bug here. So each cross product is widened to `long` first, then the exact helpers do the add/subtract and `toIntExact` narrows back. `FractionTest` passes (35/35).

garydgregory:
FYI I tested that the new tests fail without the main changes, merged 🚀  TY @alhudz

PR #1710, opened 2026-06-16
Drop trailing empty token from splitByWholeSeparator

alhudz:
Repro: `StringUtils.splitByWholeSeparator("a:b:", ":")` returns `["a", "b", ""]`; expected `["a", "b"]`. The sibling `StringUtils.split("a:b:", ":")` already returns `["a", "b"]`.

Cause: in `splitByWholeSeparatorWorker` the no-more-separator branch adds `str.substring(beg)` unconditionally. When the input ends on a separator, `beg == len`, so an empty trailing token is appended. Leading and adjacent separators are already collapsed in non-preserve mode, so only the trailing case leaked, which disagrees with the documented "adjacent separators are treated as one separator".

Fix: in non-preserve mode only add the trailing token when it is non-empty (`beg < len`). `splitByWholeSeparatorPreserveAllTokens` keeps the trailing empty token as before.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsTest.java
@@ -2436,6 +2436,13 @@ class StringUtilsTest extends AbstractLangTest {
         for (int i = 0; i < splitWithMultipleSeparatorExpectedResults.length; i++) {
             assertEquals(splitWithMultipleSeparatorExpectedResults[i], splitWithMultipleSeparator[i]);
         }
+
+        // a trailing separator must not leak an empty token (it is dropped, like leading and adjacent ones)
+        assertArrayEquals(new String[] {"a", "b"}, StringUtils.splitByWholeSeparator("a:b:", ":"));
+        assertArrayEquals(new String[] {"a"}, StringUtils.splitByWholeSeparator("a:", ":"));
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -2436,6 +2436,13 @@ void testSplitByWholeString_StringStringBoolean() {
         for (int i = 0; i < splitWithMultipleSeparatorExpectedResults.length; i++) {
             assertEquals(splitWithMultipleSeparatorExpectedResults[i], splitWithMultipleSeparator[i]);
         }
+
+        // a trailing separator must not leak an empty token (it is dropped, like leading and adjacent ones)
```
garydgregory, line 2450: Please add a new `@ParameterizedTest` method for these new assertions.

garydgregory (changes requested):
@alhudz 
Please see my comment.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
alhudz: Done. Moved them into a new `@ParameterizedTest`, `testSplitByWholeStringDropsTrailingEmpty`, driven by `@CsvSource`. Each row also asserts the preserve-all-tokens variant still keeps the trailing empty token. Confirmed it fails on every row without the worker guard.

garydgregory:
FYI I tested that the new tests fail without the main changes, merged 🚀  TY @alhudz

PR #1711, opened 2026-06-16
Fix int mask sign extension in BitField long methods

alhudz:
Repro: `new BitField(0x80000000)`, then `getRawValue(-1L)` / `clear(-1L)`.
Expected: `getRawValue(-1L)` == `0x80000000` and `clear(-1L)` == `0xFFFFFFFF7FFFFFFF`, since only bit 31 belongs to the field.
Actual: `getRawValue(-1L)` == `0xFFFFFFFF80000000` and `clear(-1L)` == `0x7FFFFFFF`, so bits 32-63 leak into the field.
Cause: the `long mask` field and the long-typed methods arrived in 3.21.0, but `BitField(int)` still assigns the int mask straight into the `long` field, so any mask with bit 31 set is sign-extended across bits 32-63. The int-typed methods cast back to `int` and hide it, the long ones do not.
Fix: build the field with `Integer.toUnsignedLong(mask)` so the int mask only occupies the low 32 bits.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BitField.java
+++ b/src/main/java/org/apache/commons/lang3/BitField.java
@@ -82,7 +82,7 @@ public class BitField {
      * @param mask the mask specifying which bits apply to this BitField. Bits that are set in this mask are the bits that this BitField operates on.
      */
     public BitField(final int mask) {
-        this.mask = mask;
+        this.mask = Integer.toUnsignedLong(mask);
         this.shiftCount = mask == 0 ? 0 : Integer.numberOfTrailingZeros(mask);
     }
 
--- a/src/main/java/org/apache/commons/lang3/BitField.java
+++ b/src/main/java/org/apache/commons/lang3/BitField.java
@@ -83,7 +83,7 @@ public class BitField {
      */
     public BitField(final int mask) {
         this.mask = Integer.toUnsignedLong(mask);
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/BitField.java
```diff
@@ -82,7 +82,7 @@ public class BitField {
      * @param mask the mask specifying which bits apply to this BitField. Bits that are set in this mask are the bits that this BitField operates on.
      */
     public BitField(final int mask) {
-        this.mask = mask;
+        this.mask = Integer.toUnsignedLong(mask);
         this.shiftCount = mask == 0 ? 0 : Integer.numberOfTrailingZeros(mask);
```
garydgregory: Hello @alhudz 
Thank you for the PR.
Should the argument to `numberOfTrailingZeros` be `mask` or `this.mask`?

alhudz: Good point. They're equivalent here since `Integer.toUnsignedLong` only zero-fills the high bits, so the trailing-zero count is unchanged, but deriving `shiftCount` from `this.mask` keeps it tied to the actual field and lines up with the `long` constructor. Switched it to `Long.numberOfTrailingZeros(this.mask)`.

PR #1712, opened 2026-06-17
Strip L suffix for oversized hex literals in createNumber

alhudz:
Repro: `NumberUtils.createNumber("0xFFFFFFFFFFFFFFFFL")` throws `NumberFormatException: For input string: "FFFFFFL"`.
Expected: a `BigInteger` of `0xFFFFFFFFFFFFFFFF`, the same result as `createNumber("0xFFFFFFFFFFFFFFFF")` (no suffix) and the decimal path `createNumber("9223372036854775808L")`.
Cause: the hex branch that is too large for a `Long` calls `createBigInteger(str)` with the `L`/`l` type suffix still attached. `createBigInteger` strips only the radix prefix and then runs `new BigInteger(substring, 16)`, so the trailing `L` reaches the hex parser and is rejected. The decimal branch already drops the suffix before its `BigInteger` fallback, which is why only the hex form throws.
Fix: drop the trailing type char before the `createBigInteger` fallback when the literal carries one, so an over-long hex literal with a suffix parses the same as the unsuffixed form.

Covers the `0x`/`0X`/`#` prefixes, signed and unsigned. Regression assertions added to `NumberUtilsTest.testCreateNumber`; they fail with `NumberFormatException` before the change and pass after.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
+++ b/src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
@@ -661,6 +661,13 @@ class NumberUtilsTest extends AbstractLangTest {
         assertEquals(Integer.decode("+0xF"), NumberUtils.createNumber("+0xF"), "createNumber(String) LANG-1645a failed");
         assertEquals(Long.decode("+0xFFFFFFFF"), NumberUtils.createNumber("+0xFFFFFFFF"), "createNumber(String) LANG-1645b failed");
         assertEquals(new BigInteger("+FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("+0xFFFFFFFFFFFFFFFF"), "createNumber(String) LANG-1645c failed");
+        // A hex literal too large for a Long but carrying an explicit 'L'/'l' type suffix must drop the suffix
+        // before falling back to BigInteger, matching the decimal path (e.g. "12345678901234567890L").
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("0xFFFFFFFFFFFFFFFFL"));
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("0xFFFFFFFFFFFFFFFFl"));
+        assertEquals(new BigInteger("8000000000000000", 16), NumberUtils.createNumber("0x8000000000000000L"));
+        assertEquals(new BigInteger("-FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("-0xFFFFFFFFFFFFFFFFL"));
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("#FFFFFFFFFFFFFFFFL"));
         // Map to a BigDecimal, not a Float.
         assertEquals(new BigDecimal("0.100000001490116121"), NumberUtils.createNumber("0.100000001490116121"));
     }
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
```diff
@@ -661,6 +661,13 @@ void testCreateNumber() {
...
         assertEquals(Long.decode("+0xFFFFFFFF"), NumberUtils.createNumber("+0xFFFFFFFF"), "createNumber(String) LANG-1645b failed");
         assertEquals(new BigInteger("+FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("+0xFFFFFFFFFFFFFFFF"), "createNumber(String) LANG-1645c failed");
+        // A hex literal too large for a Long but carrying an explicit 'L'/'l' type suffix must drop the suffix
+        // before falling back to BigInteger, matching the decimal path (e.g. "12345678901234567890L").
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("0xFFFFFFFFFFFFFFFFL"));
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("0xFFFFFFFFFFFFFFFFl"));
+        assertEquals(new BigInteger("8000000000000000", 16), NumberUtils.createNumber("0x8000000000000000L"));
+        assertEquals(new BigInteger("-FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("-0xFFFFFFFFFFFFFFFFL"));
+        assertEquals(new BigInteger("FFFFFFFFFFFFFFFF", 16), NumberUtils.createNumber("#FFFFFFFFFFFFFFFFL"));
         // Map to a BigDecimal, not a Float.
```
sarankumarbaskar, line 671: This looks consistent with the decimal suffix path, and the added tests fail without the production change. One small suggestion: since the description mentions signed forms and nearby LANG-1645 tests cover +0x..., it would give wider coverage to add a +0xFFFFFFFFFFFFFFFFL assertion as well.

```java
assertEquals(new BigInteger("+FFFFFFFFFFFFFFFF", 16),
        NumberUtils.createNumber("+0xFFFFFFFFFFFFFFFFL"));

sarankumarbaskar (changes requested):

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
garydgregory: @sarankumarbaskar 
You wrote "and the added tests fail without the production change."
If I add the test only side of the patch, ALL of the new BigInteger assertions in this patch fail. So what are you talking about?

garydgregory:
FYI I tested that the each new BI assertion fails without the main changes, merged 🚀  TY @alhudz

src/test/java/org/apache/commons/lang3/math/NumberUtilsTest.java
sarankumarbaskar: You’re right, I worded that poorly. I only meant the new assertions exercise the changed path on current master. Sorry for the confusion.

alhudz: Fair point. The `+` form runs through the same suffix-strip: the leading sign stays on `str` and only the trailing `L` is dropped, so `+0xFFFFFFFFFFFFFFFFL` resolves to `new BigInteger("+FFFFFFFFFFFFFFFF", 16)` exactly like the `-0x...L` case that's already asserted. It merged before I could fold the `+` assertion in, but happy to add it as a small follow-up if the explicit case is wanted.

PR #1713, opened 2026-06-17
Fix indexOfDifference splitting a surrogate pair

alhudz:
`Repro:` `getCommonPrefix("𐐀", "𐐁")` returns the lone high surrogate `\uD801`, and `difference("𐐀", "𐐁")` returns the lone low surrogate `\uDC01`. Both are malformed UTF-16.
`Cause:` both `indexOfDifference` overloads walk the inputs one `char` at a time, so when two strings share the high surrogate of a supplementary code point but differ in the low half, the reported index lands inside the pair (index `1` rather than `0`). `getCommonPrefix` then slices the pair in half and `difference` starts its result on the orphaned low half.
`Fix:` when the difference falls on a low surrogate whose preceding (common) `char` is a high surrogate, report the start of the pair. `getCommonPrefix` and `difference` follow for free. BMP inputs never hit the new branch, so existing behaviour is unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3002,6 +3002,12 @@ public class StringUtils {
                 break;
             }
         }
+        if (firstDiff > 0 && Character.isLowSurrogate(css[0].charAt(firstDiff))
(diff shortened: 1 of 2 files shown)


garydgregory:
I tested that the tests fail without the main changes, merged 🚀  TY @alhudz

PR #1717, opened 2026-06-19
Fix silent int overflow in Fraction.getFraction(double)

alhudz:

---

Repro: `Fraction.getFraction(2147483646.5)` returns `-3/2` and `getFraction(1073741824.5)` returns `-2147483647/2`.

Cause: the result numerator is rebuilt as `numer0 + wholeNumber * denom0` in `int`. `wholeNumber` can be up to `Integer.MAX_VALUE` and `denom0` is the convergent denominator (`> 1` for any non-integer value), so the product overflows and wraps. The value is not representable as an `int`/`int` `Fraction`, so the documented `ArithmeticException` should be thrown instead of returning a wrong reduced fraction.

Fix: route the numerator reconstruction through the existing `mulAndCheck` and `Math.addExact`, so the overflow is rejected the same way the `value > Integer.MAX_VALUE` guard already rejects out-of-range input. Whole numbers (where `denom0 == 1`) and all in-range values are unaffected.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -174,7 +174,10 @@ public final class Fraction extends Number implements Comparable<Fraction> {
         if (i == 25) {
             throw new ArithmeticException("Unable to convert double to fraction");
         }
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz 

I checked the new assertions fail without the `main` changes and merged the PR 🚀 

TY!

PR #1718, opened 2026-06-19
Find NaN in ArrayUtils.lastIndexOf for double and float arrays

alhudz:
Repro: `ArrayUtils.lastIndexOf(new double[]{Double.NaN}, Double.NaN)` returns `-1`, while `ArrayUtils.indexOf` on the same array returns `0` (and `contains` returns `true`).
Cause: `indexOf(double[])`/`indexOf(float[])` were given explicit `NaN` matching (the `indexOfNaN` helper for `double`, a `searchNaN` flag for `float`), but the `lastIndexOf` overloads still only compare with `valueToFind == array[i]`, which is never `true` for `NaN`.
Fix: route the two `double` `lastIndexOf` overloads (with `startIndex`, and with `startIndex`+`tolerance`) through a new private `lastIndexOfNaN` helper that mirrors `indexOfNaN`, and add the same `searchNaN` check to `lastIndexOf(float[], float, int)`, so `lastIndexOf` agrees with `indexOf` on `NaN`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -4265,6 +4265,9 @@ public class ArrayUtils {
      * @return the last index of the value within the array, {@link #INDEX_NOT_FOUND} ({@code -1}) if not found or {@code null} array input.
      */
     public static int lastIndexOf(final double[] array, final double valueToFind, int startIndex) {
+        if (Double.isNaN(valueToFind)) {
+            return lastIndexOfNaN(array, startIndex);
+        }
         if (isEmpty(array) || startIndex < 0) {
             return INDEX_NOT_FOUND;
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
@alhudz 

Would you please add tests for:
- `java.lang.Double|Float.NEGATIVE_INFINITY`
- `java.lang.Double|Float.POSITIVE_INFINITY`

Thank you!

alhudz:
Done. Added `testLastIndexOfDoubleInfinity` and `testLastIndexOfFloatInfinity` covering `POSITIVE_INFINITY` and `NEGATIVE_INFINITY` across the overloads, with each infinity appearing twice so the tests pin the last-occurrence index. Both infinities already matched via `==` (and the tolerance path), so this is test coverage only, no runtime change there.

PR #1719, opened 2026-06-19
Keep abbreviate and truncate off surrogate pair boundaries

alhudz:
Repro: `StringUtils.abbreviate("😀abcdef", 4)` returns a lone high surrogate followed by `...`, and `StringUtils.truncate("a😀b", 0, 2)` returns `a` plus a lone high surrogate.
Cause: both choose the cut position in `char` units and slice with `String.substring`, so a boundary that lands between the two halves of a surrogate pair leaves an unpaired surrogate behind.
Fix: nudge the boundary to the nearest code-point edge before slicing, so a pair is never split. The result still never exceeds `maxWidth`.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/StringUtilsAbbreviateTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringUtilsAbbreviateTest.java
@@ -19,6 +19,7 @@ package org.apache.commons.lang3;
 
 import static org.apache.commons.lang3.LangAssertions.assertIllegalArgumentException;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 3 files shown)


garydgregory:
@alhudz Please address https://github.com/apache/commons-lang/pull/1718#pullrequestreview-4532690270

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
```diff
@@ -3054,6 +3054,18 @@ void testTruncate_StringIntInt() {
...
+    @Test
+    void testTruncate_StringIntInt_surrogatePair() {
+        // U+1F600 GRINNING FACE is a single supplementary code point stored as a surrogate pair
+        final String grin = "😀";
+        // a cut that would land between the two halves keeps the result well formed instead of emitting a lone surrogate
+        assertEquals("a", StringUtils.truncate("a" + grin + "b", 0, 2));
+        assertEquals(grin, StringUtils.truncate("a" + grin + "b", 1, 2));
+        assertEquals("ab", StringUtils.truncate("ab" + grin, 0, 3));
+        // an offset that lands inside a pair skips the orphaned low surrogate
+        assertEquals("a", StringUtils.truncate(grin + "ab", 1, 2));
```
garydgregory, line 3066: You're missing assertions for an input string that _only_ contains the `grin`, with permutations of the other parameters.

src/test/java/org/apache/commons/lang3/StringUtilsAbbreviateTest.java
```diff
@@ -170,6 +171,31 @@ void testAbbreviate_StringStringIntInt() {
...
+                if (Character.isHighSurrogate(ch)) {
+                    assertTrue(i + 1 < result.length() && Character.isLowSurrogate(result.charAt(i + 1)), "lone high surrogate in: " + result);
+                    i++; // skip the paired low surrogate
+                } else {
+                    assertFalse(Character.isLowSurrogate(ch), "lone low surrogate in: " + result);
+                }
+            }
+        }
+    }
+
```
garydgregory, line 206: You're missing assertions for an input string that _only_ contains the `grin`, with permutations of the other parameters.

garydgregory (changes requested):
@alhudz 
Thank you for your PR.
Please see my 2 comments.

src/test/java/org/apache/commons/lang3/StringUtilsTest.java
alhudz: Added. `truncate(grin, ...)` is now covered across offset/maxWidth permutations: the whole pair when it fits (`0,2`, `0,3`), and empty when the cut or the offset lands inside it (`0,0`, `0,1`, `1,1`, `1,2`, `2,2`). Never a lone surrogate.

src/test/java/org/apache/commons/lang3/StringUtilsAbbreviateTest.java
alhudz: Done. Added `abbreviate(grin, ...)` across the width, offset and marker permutations. The pair is kept whole in the normal cases, and the empty-marker case (`grin, "", 0, 1`) exercises the head cut backing off to empty instead of emitting a lone surrogate.

alhudz:
The 1718 review is already handled, I pushed the `POSITIVE_INFINITY`/`NEGATIVE_INFINITY` tests for double and float there. For this PR I added the grin-only assertions across the parameter permutations and replied inline on both threads.

garydgregory (changes requested):
Hello @alhudz 

Thank you for this PR.

With this PR applied locally, and with the assertions in `org.apache.commons.lang3.StringUtilsAbbreviateTest.testEmoji()` commented _in_, that test fails. 

Would you please review `testEmoji()` failures in light of this PR.

Thank you!

alhudz:
Had a look at `testEmoji()`. With the `expectedResultsFox`/`expectedResultsFamilyWithCodepoints` assertions commented in, it fails both before and after this PR, so they have never been green on `master`:

`abbreviate("🦊…", 4)`
- before: `<lone high surrogate>...` (the malformed output this PR targets)
- this PR: `...`
- `testEmoji` expects: `🦊...`

The gap is the contract. `testEmoji` counts `maxWidth` in code points: the marker is 3, then `maxWidth - 3` whole code points (width 4 → 1 fox, width 5 → 2 foxes, and the family case counts each skin-tone modifier and ZWJ as one). So `🦊...` is 5 `char`s but 4 code points. This PR keeps the documented `char`-based contract (`result.length() <= maxWidth`) and only nudges the cut to a code-point boundary, which is always shorter, so it drops the partial emoji rather than keeping it whole and over the numeric width.

So this change is strictly the lone-surrogate fix. Making `testEmoji` pass is the larger LANG-1770 job of re-basing `abbreviate` on code points, which redefines what `maxWidth` means and the `length() <= maxWidth` guarantee. Happy to take that on as the actual LANG-1770 fix if you want it in this PR, or keep this one scoped to removing the malformed surrogate. Which would you prefer?

garydgregory:
@alhudz 
We can take it one task at time. I'll merge this PR and you can continue in another.
TY!

PR #1720, opened 2026-06-20
Fix int overflow in DurationFormatUtils.formatPeriod

alhudz:

---

Repro: `DurationFormatUtils.formatPeriod(0, 2175984000000L, "s", true, gmt)` for a ~69 year span (`gmt` is `TimeZone.getTimeZone("GMT")`).
Expected: `2175984000`, the value the sibling `formatDuration(2175984000000L, "s")` returns for the same span.
Actual: `-2118983296`.

Cause: `formatPeriod` keeps its day/hour/minute/second estimates in `int`. When the format omits a higher field, that field is funnelled into the next lower one (`hours += 24 * days`, `minutes += 60 * hours`, `seconds += 60 * minutes`) in `int` arithmetic before the values widen to the `long` parameters of `format`. Past a few decades the second (or millisecond) count exceeds `Integer.MAX_VALUE` and the multiply wraps negative.

Fix: declare the six estimate locals as `long` so the reduction runs in `long`, the same way `formatDuration` already reduces the span. `milliseconds` was already `long`; normal-size durations produce identical output.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -540,12 +540,12 @@ public class DurationFormatUtils {
         end.setTime(new Date(endMillis));
         // initial estimates
         long milliseconds = end.get(Calendar.MILLISECOND) - start.get(Calendar.MILLISECOND);
-        int seconds = end.get(Calendar.SECOND) - start.get(Calendar.SECOND);
-        int minutes = end.get(Calendar.MINUTE) - start.get(Calendar.MINUTE);
-        int hours = end.get(Calendar.HOUR_OF_DAY) - start.get(Calendar.HOUR_OF_DAY);
-        int days = end.get(Calendar.DAY_OF_MONTH) - start.get(Calendar.DAY_OF_MONTH);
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @alhudz 

Thank you for the PR.

Would you please add a tests that:
- asserts that the right formatting happens for inputs around and at `Long.MAX_VALUE`, for example the duration starting at `Long.MAX_VALUE  -1` and ending at `Long.MAX_VALUE`.
- Same for `Long.MIN_VALUE`
- Same for `Integer.MIN_VALUE`

This should hopefully show that we can correctly handle the full range of long millisecond values for the input to this method.

TY!

alhudz:
Added `testFormatPeriodLongRangeBounds`, which runs a one-millisecond span at each end of the input range:

- `Long.MAX_VALUE - 1` .. `Long.MAX_VALUE`
- `Long.MIN_VALUE` .. `Long.MIN_VALUE + 1`
- `Integer.MIN_VALUE - 1` .. `Integer.MIN_VALUE`

Each one asserts `S` is `1`, the full `y/M/d/H/m/s.S` is `0/0/0/0/0/0.001`, and that `formatPeriod` matches `formatDuration` for the same span, so the reduction holds across the whole long range and not just spans near the epoch. `DurationFormatUtilsTest` is green (37 tests).

PR #1721, opened 2026-06-20
Reject sign characters in UnicodeUnescaper hex values

alhudz:
`UnicodeUnescaper.translate` reads the four-character value of a `\uXXXX` escape with `Integer.parseInt(value, 16)`, which tolerates a leading sign.

1. `\u-047` decodes to `U+FFB9` rather than being rejected, because `parseInt("-047", 16)` returns `-71`.
2. an embedded sign such as `\u02-3` already throws `IllegalArgumentException`, so accepting a leading sign is inconsistent.

Reject a leading `+`/`-` in the value field before parsing. The documented `u+` notation (`\u+0047`) is unaffected.

Repro: `new UnicodeUnescaper().translate("\\u-047")`
Expected: `IllegalArgumentException`
Actual: returns the string `\uFFB9`

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/translate/UnicodeUnescaper.java
+++ b/src/main/java/org/apache/commons/lang3/text/translate/UnicodeUnescaper.java
@@ -59,6 +59,11 @@ public class UnicodeUnescaper extends CharSequenceTranslator {
                 // Get 4 hex digits
                 final CharSequence unicode = input.subSequence(index + i, index + i + 4);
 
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz 
Would you please review [`UnicodeEscaper`](https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/translate/UnicodeEscaper.java) in Apache Commons Text?

PR #1722, opened 2026-06-20
Keep initials from splitting a supplementary code point

alhudz:
Repro: `WordUtils.initials("Ben 😀mile Lee")` where the second word begins with U+1F600.
Cause: the loop copies the first `char` after a delimiter and skips the rest of the word, so a word that starts with a supplementary code point keeps only the high surrogate and the low half is dropped, leaving a lone surrogate in the result (`B` + U+D83D + `L`).
Fix: copy the trailing low surrogate together with its high half, and size the buffer to the input length so a two-`char` initial cannot run past it. BMP input is unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -265,7 +265,7 @@ public class WordUtils {
             return StringUtils.EMPTY;
         }
         final int strLen = str.length();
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz Please see my comment https://github.com/apache/commons-lang/pull/1719#pullrequestreview-4537228565

garydgregory:
@alhudz 
Would please check the Commons Text version of this class for the same issue?
TY!

PR #1724, opened 2026-06-22
Classify supplementary code points in StringUtils is* predicates

alhudz:
Repro: `StringUtils.isAlpha(new String(Character.toChars(0x10400)))` (`DESERET CAPITAL LETTER LONG I`, a letter) returns `false`, and `StringUtils.isNumeric(new String(Character.toChars(0x1D7CE)))` (`MATHEMATICAL BOLD DIGIT ZERO`, a digit) returns `false`.

Cause: the `is*` predicates (`isAlpha`, `isAlphaSpace`, `isAlphanumeric`, `isAlphanumericSpace`, `isNumeric`, `isNumericSpace`, `isAllLowerCase`, `isAllUpperCase`, `isMixedCase`) walk the input one `char` at a time and hand each `char` to `Character.isLetter`/`isDigit`/`isLetterOrDigit`/`isUpperCase`/`isLowerCase`. A supplementary code point is two surrogate chars, neither of which is a letter, digit or cased char on its own, so a supplementary letter or digit is rejected even though the Javadoc tests for Unicode letters/digits.

Fix: iterate with `Character.codePointAt` and advance by `Character.charCount`, classifying the whole code point at each site. BMP input keeps the same result (a lone surrogate still classifies as before).

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -3237,10 +3237,12 @@ public class StringUtils {
             return false;
         }
         final int sz = cs.length();
-        for (int i = 0; i < sz; i++) {
-            if (!Character.isLowerCase(cs.charAt(i))) {
+        for (int i = 0; i < sz;) {
+            final int codePoint = Character.codePointAt(cs, i);
+            if (!Character.isLowerCase(codePoint)) {
                 return false;
             }
+            i += Character.charCount(codePoint);
         }
         return true;
     }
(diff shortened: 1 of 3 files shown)


garydgregory (changes requested):
Hello @alhudz 
Please breakout the new test assertions into new test methods.
I think it best to see what works and what doesn't RE supplementary characters in their own tests.
TY!

alhudz:
Done. Split the supplementary code point assertions out into their own test methods (`testIsAlphaSupplementary`, `testIsNumericSupplementary`, `testIsMixedCaseSupplementary` and so on), one per predicate, so it's clear which supplementary cases pass and which don't. The original per-predicate tests are back to what they were.

garydgregory:
TY @alhudz , merged 🚀

PR #1725, opened 2026-06-22
Fold supplementary code points in CharSequenceUtils.regionMatches

alhudz:
`CharSequenceUtils.regionMatches` is the green stand-in for `String.regionMatches` used whenever an argument is not a `String`, but its `ignoreCase` branch folds case one UTF-16 `char` at a time. The two halves of a supplementary code point are surrogates with no case mapping, so a case-insensitive compare of a supplementary letter never matches. Repro with `U+10400` (`𐐀`, Deseret capital long I) and its lowercase `U+10428` (`𐐨`): `StringUtils.equalsIgnoreCase("𐐀", "𐐨")` returns `true` through the `String` fast path, yet `equalsIgnoreCase(new StringBuilder("𐐀"), "𐐨")` returns `false`. `startsWithIgnoreCase`, `endsWithIgnoreCase`, `containsIgnoreCase` and `Strings.CI.indexOf` go the same way, so the answer depends on the argument's runtime type.

On a `char`-level mismatch the fix rebuilds the full code point from the adjacent surrogate and folds at the code point level, which is what `String.regionMatches` itself does, so the green path lines back up with the `String`/`String` path. Cross-checked against `String.regionMatches` over a few million randomised BMP, supplementary and lone-surrogate inputs with no remaining divergence; the added `CharSequenceUtilsTest` cases fail before the change and pass after.

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
+++ b/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
@@ -319,9 +319,14 @@ public class CharSequenceUtils {
             if (!ignoreCase) {
                 return false;
             }
-            // The same case-insensitive check as String#regionMatches(boolean, int, String, int, int),
-            // which also folds case of a supplementary code point split across a surrogate pair.
+            // The same case-insensitive check as String#regionMatches(boolean, int, String, int, int).
             if (!equalsIgnoreCase(c1, c2)) {
+                // String only folds a supplementary code point split across a surrogate pair from Java 9
+                // on; on Java 8 it stays char-by-char. Track the running JDK so every CharSequence type
+                // gives the same result that String does there.
+                if (SystemUtils.IS_JAVA_1_8) {
+                    return false;
+                }
                 int cp1 = c1;
                 if (Character.isHighSurrogate(c1)) {
                     if (index1 + 1 < end1 && Character.isLowSurrogate(cs.charAt(index1 + 1))) {
--- a/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
+++ b/src/main/java/org/apache/commons/lang3/CharSequenceUtils.java
@@ -27,6 +27,17 @@ public class CharSequenceUtils {
 
     private static final int NOT_FOUND = -1;
 
+    /**
+     * Whether the running JDK folds a supplementary code point split across a surrogate pair when comparing case
+     * insensitively in {@link String#regionMatches(boolean, int, String, int, int)}. JDKs up to and including Java 11
+     * compare surrogate by surrogate and never match such a pair; later JDKs fold the whole code point. Probing what
+     * {@link String} actually does (rather than gating on a version constant) keeps every {@link CharSequence} type in
+     * step with {@link String} on whatever JDK is running. DESERET CAPITAL LETTER LONG I (U+10400) folds to its small
+     * form (U+10428).
+     */
+    private static final boolean STRING_FOLDS_SUPPLEMENTARY_CASE =
+            new String(Character.toChars(0x10400)).regionMatches(true, 0, new String(Character.toChars(0x10428)), 0, 2);
+
     static final int TO_STRING_LIMIT = 16;
 
     private static boolean checkLaterThan1(final CharSequence cs, final CharSequence searchChar, final int len2, final int start1) {
@@ -321,10 +332,9 @@ public class CharSequenceUtils {
             }
             // The same case-insensitive check as String#regionMatches(boolean, int, String, int, int).
             if (!equalsIgnoreCase(c1, c2)) {
-                // String only folds a supplementary code point split across a surrogate pair from Java 9
-                // on; on Java 8 it stays char-by-char. Track the running JDK so every CharSequence type
-                // gives the same result that String does there.
-                if (SystemUtils.IS_JAVA_1_8) {
+                // Only fold a supplementary code point split across a surrogate pair where String itself does, so
+                // every CharSequence type gives the same result that String does on the running JDK (see field).
+                if (!STRING_FOLDS_SUPPLEMENTARY_CASE) {
                     return false;
                 }
                 int cp1 = c1;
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz
The test fails here.

sarankumarbaskar:
I can reproduce the failure on Java 8.
`[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.commons.lang3.CharSequenceUtilsTest
[ERROR] Tests run: 14, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.202 s <<< FAILURE! -- in org.apache.commons.lang3.CharSequenceUtilsTest
[ERROR] org.apache.commons.lang3.CharSequenceUtilsTest.testRegionMatches -- Time elapsed: 0.009 s <<< FAILURE!
org.opentest4j.AssertionFailedError: String Failed test 𐐀[0] caseblind 𐐨[0] 2 => true ==> expected: <true> but was: <false>`

The new Deseret test cases are added to the existing regionMatches parity test data, which also checks the String.regionMatches(true, ...) path. On Java 8, String.regionMatches(true, 0, "𐐨", 0, 2) for "𐐀" returns false, so the String check fails before this can purely validate the CharSequenceUtils path.

That seems to make the intended behavior a design question:
@garydgregory 

Should CharSequenceUtils.regionMatches continue to match String.regionMatches behavior on the running JDK, including Java 8, or should it intentionally provide supplementary-code-point case folding even where Java 8 String.regionMatches does not?

If the goal is JDK parity, the new cases may need to be version-gated or adjusted. If the goal is improved Unicode behavior beyond Java 8 String.regionMatches, then these cases may need to move out of the existing String parity test data and into a dedicated CharSequenceUtils/non-String test.

alhudz:
@sarankumarbaskar has it right. `String.regionMatches(true, ...)` only folds a supplementary code point at the code point level from Java 9 on; on Java 8 it stays char-by-char and returns `false`, so the Deseret rows in the shared `TEST_DATA` failed on the `String`/`CSString` runners there.

This PR only touches the green non-String path, which now folds the whole code point on every JDK (matching Java 9+ `String`). On Java 8 that leaves the green path a touch ahead of the JDK's own `String`, which seems like the right call for a `CharSequence` helper rather than reproducing the old surrogate-by-surrogate behaviour. I've moved those cases out of the String-parity data into a dedicated `testRegionMatchesSupplementaryCaseFold` that drives only the non-String path, and kept `TEST_DATA` on BMP inputs that agree with `String` everywhere. Build is green on 8 through 25 now.

sarankumarbaskar:
Thanks for the update. Moving these out of the shared String parity data makes sense to me, since Java 8 differs from Java 9+ for String.regionMatches here.

garydgregory:
Very interesting discussion @alhudz, @sarankumarbaskar. Thank you for that. 

I think we might have more to consider here. Going back to commit 2541a62def1c6e106ab4f4e0c72c45872dffca00 , we have this code path in `CSU.regionMatches()`:
```java
        if (cs instanceof String && substring instanceof String) {
            return ((String) cs).regionMatches(ignoreCase, thisStart, (String) substring, start, length);
        }
```
I would argue that the question is whether that commit and the code until now delegates to String as an optimization or a feature (it's both IMO). Java has evolved since that commit as noted, and now presents different behavior as of Java 9. (As a point of reference, our base requirement is Java 8 with the goal of supporting current LTS releases, so 8, 11, 17, 21, 25.)

Passing in a `String`, a `StringBuilder`, or any CS implementation, to CSU should have the same result (so the original test was correct I think). I claim that if I call this CSU method, it is either because I have a variable typed as a CS, or I have a concrete CS implementation and I want the CSU behavior to future-proof a later change in my code to CS. 

Obviously, calling the CSU method with a String or calling the String method directly should have the same results.

All of this to say:
The non-String path in CSU should be a CS version of the `String` implementation, such that tests verify that `String`, `StringBuilder`, `StringBuffer`, `CharBuffer`, all have the same behavior. If that means that the guts of the CSU method has a switch on `SystemUtils.IS_JAVA_1_8`, then so bit it.

WDYT?

Tangent: Once this is resolved, `org.apache.commons.codec.binary.CharSequenceUtils` should also be reviewed.

alhudz:
Makes sense, that's the right framing. I've gated the supplementary fold on `SystemUtils.IS_JAVA_1_8`, so the non-String path now mirrors `String#regionMatches` on whatever JDK is running: Java 9+ folds the whole code point, Java 8 stays char-by-char and returns `false` the same way its own `String` does. Passing a `String`, `StringBuilder`, `StringBuffer` or `CharBuffer` to `CSU` now comes out the same on a given JDK, and matches calling `String` directly.

For the tests I drive the supplementary rows off the running JDK's `String` result rather than a fixed boolean, and assert `String`, `StringBuilder`, `StringBuffer` and `CharBuffer` all agree with it, so they're green on 8 through 25 instead of only on 9+. `TEST_DATA` keeps the BMP rows that read the same everywhere.

Happy to look at `org.apache.commons.codec.binary.CharSequenceUtils` for the same thing once this one lands.

garydgregory:
Hello @alhudz
Thank you for your update. 
The build fails. What versions did you test before pushing?

alhudz:
My fault on that one. I only had a single JDK in front of me last time and `String` happened to fold there, so the bad gate slipped through.

The behaviour isn't actually a Java 8 vs 9 split. `String.regionMatches(true, ...)` only started folding a supplementary code point at the code point level somewhere after 11: it returns `false` on 8 and 11, and `true` on 21/26. So gating on `IS_JAVA_1_8` was wrong, it folded on 11 where `String` itself still doesn't.

Rather than pin the exact cutover version, I now probe once at class load what `String` actually does and only fold when it does, so the non-String path tracks `String` on whatever JDK is running.

Checked it properly this time: `CharSequenceUtilsTest` is green on 11, 21 and 26, and a parity run of `StringBuilder`/`StringBuffer`/`CharBuffer` against `String` over ~6M random BMP/supplementary/lone-surrogate inputs gives 0 divergence on each. Java 8 takes the same no-fold path as 11.

garydgregory:
TY @alhudz & @sarankumarbaskar , I merged the PR 🚀

PR #1726, opened 2026-06-23
Emit surrogate pair for supplementary code points in UnicodeEscaper

alhudz:
Repro: escape any supplementary code point with the base `UnicodeEscaper`, e.g. `UnicodeEscaper.above(0x7f).translate(new String(Character.toChars(0x10437)))`.
Expected: `\uD801\uDC37`.
Actual: `\u10437`, a single `\u` carrying five hex digits, which no `\u` reader decodes back to U+10437 (`UnicodeUnescaper` reads `\u1043` then a literal `7`).
Cause: `toUtf16Escape` returns `"\\u" + hex(codePoint)`, but a code point above `0xffff` does not fit the four hex digits a `\u` escape holds, so the surrogate split is never applied.
Fix: emit the surrogate-pair form `\uXXXX\uXXXX` for supplementary code points, matching the method name and the behaviour `JavaUnicodeEscaper` already has. BMP code points are untouched.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/translate/UnicodeEscaper.java
+++ b/src/main/java/org/apache/commons/lang3/text/translate/UnicodeEscaper.java
@@ -102,7 +102,8 @@ public class UnicodeEscaper extends CodePointTranslator {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Good one, @alhudz, merged 🚀

PR #1727, opened 2026-06-23
Fix MethodUtils.getMatchingMethod false ambiguity on boxed arguments

alhudz:
`MethodUtils.getMatchingMethod` ranks overloads with the private `distance` helper, which charges a flat `+1` for any autoboxing conversion. For a primitive argument that makes `int` to `Integer` and `int` to `Number` both score `1`, so when a class declares both `f(Integer)` and `f(Number)` the call `getMatchingMethod(cls, "f", int.class)` puts two candidates in the same distance bucket and throws `IllegalStateException: Found multiple candidates`, even though `javac` resolves `f(someInt)` unambiguously to `f(Integer)`.

The fix keeps the flat penalty when a primitive boxes to its exact wrapper but adds the supertype cost when it boxes to a supertype, so `int` to `Integer` ranks strictly below `int` to `Number`. `distance` is private to `getMatchingMethod`, so the change stays inside that resolver; the auto-unboxing direction and every existing resolution are unaffected.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/MethodUtils.java
@@ -90,7 +90,12 @@ public class MethodUtils {
                 continue;
             }
             if (ClassUtils.isAssignable(aClass, toClass, true) && !ClassUtils.isAssignable(aClass, toClass, false)) {
+                // Autoboxing/unboxing conversion. When a primitive is boxed, rank the exact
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , now merged 🚀 and deployed to snapshot builds in https://repository.apache.org/content/repositories/snapshots/org/apache/commons/commons-lang3/3.21.0-SNAPSHOT/

PR #1728, opened 2026-06-24
Handle supplementary code points in WordUtils case methods

alhudz:
Repro: capitalise any word that begins with a supplementary cased letter, e.g. `WordUtils.capitalize("𐐨bc")` where `𐐨` is Deseret small long-i `U+10428`.
Cause: `capitalize`, `capitalizeFully`, `uncapitalize` and `swapCase` iterate one `char` at a time and call `Character.toTitleCase`/`toUpperCase`/`toLowerCase` on each unit. The two surrogate halves of a supplementary code point carry no case mapping, so the letter is left as-is and the result diverges from `StringUtils.capitalize`/`swapCase`, which already work on code points (`capitalize` yields `U+10428` instead of `U+10400`, `swapCase` leaves `U+10400` unchanged).
Fix: iterate with `Character.codePointAt` and write the mapped code point back with `Character.toChars`. A case mapping never changes the UTF-16 length, so the existing buffer is reused; BMP input and lone surrogates are untouched.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -301,8 +301,11 @@ public class WordUtils {
     /**
      * Tests if the code point is a delimiter.
      *
+     * <p>A {@code null} {@code delimiters} array treats any whitespace code point, as defined by
+     * {@link Character#isWhitespace(int)}, as a delimiter.</p>
+     *
      * @param codePoint  the code point to check.
-     * @param delimiters  the delimiters.
+     * @param delimiters  the delimiters, {@code null} matches whitespace.
      * @return true if it is a delimiter.
      */
     private static boolean isDelimiter(final int codePoint, final char[] delimiters) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/lang3/text/WordUtils.java
```diff
@@ -297,6 +298,25 @@ private static boolean isDelimiter(final char ch, final char[] delimiters) {
         return delimiters == null ? Character.isWhitespace(ch) : ArrayUtils.contains(delimiters, ch);
     }
 
+    /**
+     * Tests if the code point is a delimiter.
```
garydgregory, line 302: @alhudz 
This Javadoc needs to say something about the special case where a null `delimiters` does something special.

garydgregory (changes requested):
Hello @alhudz 
Please see my comment and this also likely need to be addressed in Apache Commons Text's `org.apache.commons.text.WordUtils`.

src/main/java/org/apache/commons/lang3/text/WordUtils.java
alhudz: Good catch. Updated the Javadoc to note that a `null` `delimiters` treats any whitespace code point (per `Character.isWhitespace(int)`) as a delimiter, same as the `char` overload.

alhudz:
Pushed the Javadoc fix. You're right about Commons Text, its `org.apache.commons.text.WordUtils` carries the same per-`char` iteration in these methods, so the supplementary code points break there too. I'll raise a matching PR against commons-text.

garydgregory:
@alhudz 
Merged, please port to Apache Commons Text.

PR #1730, opened 2026-06-24
Keep StrBuilder.reverse from splitting surrogate pairs

alhudz:
`Repro:` `new StrBuilder("a" + "😀" + "b").reverse()` returns `b\uDE00\uD83Da` (a low surrogate ahead of its high surrogate), where `new StringBuilder(...).reverse()` returns `b😀a`.
`Cause:` `reverse()` swaps the buffer one `char` at a time, so every surrogate pair is left in low-high order and the supplementary code point becomes malformed UTF-16.
`Fix:` after the char swap, walk the buffer once and swap each adjacent low-high surrogate pair back to high-low, matching `StringBuilder`/`StringBuffer` which this class is documented to mimic. BMP text, lone surrogates and odd-length buffers are untouched.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
@@ -2758,10 +2758,25 @@ public class StrBuilder implements CharSequence, Appendable, Serializable, Build
 
         final int half = size / 2;
         final char[] buf = buffer;
+        boolean hasSurrogates = false;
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @alhudz , the CI is running... please port to Apache Commons Text as this class is deprecated here. TY!

alhudz:
Done — ported to Commons Text in apache/commons-text#756, same surrogate-pair restore on `TextStringBuilder.reverse`.

PR #1731, opened 2026-06-25
Keep WordUtils.wrap from splitting a surrogate pair

alhudz:
Repro: `WordUtils.wrap("a😀😀😀😀", 4, "\n", true)`, four `U+1F600` after a leading `a`.
Cause: with `wrapLongWords` set, a word longer than the column is hard-broken at the fixed char offset `wrapLength + offset` and the new-line is inserted there. When that offset lands between the high and low surrogate of a supplementary code point the pair is split, so a lossless wrap emits a lone high surrogate at the end of one line and a lone low surrogate at the start of the next.
Fix: nudge the break one char forward when it would land inside a pair, keeping the whole code point on the current line. BMP input and the delimiter-based wrap paths are unaffected.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -679,10 +679,14 @@ public class WordUtils {
                 offset = endOfWrapAt;
             } else // really long word or URL
             if (wrapLongWords) {
-                // wrap really long word one line at a time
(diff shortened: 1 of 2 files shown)


garydgregory:
Merged, TY @alhudz, please port to Apache Commons Text.

alhudz:
Ported in apache/commons-text#755 - same fix, the wrap test fails on the current Text tree without it.

PR #1732, opened 2026-06-25
Fix WordUtils.containsAllWords missing words across line breaks

alhudz:
1. `containsAllWords` builds the per-word regex `.*\b<word>\b.*` and runs it with `matches()`, but compiles it without `Pattern.DOTALL`, so `.` never matches a line terminator.
2. `matches()` has to consume the whole input, so a word sitting on a different line than the surrounding `.*` can reach is never found.

Repro: `WordUtils.containsAllWords("foo\nbar", "bar")`
Expected: `true` (`bar` is present as a whole word)
Actual: `false`
Fix: compile the pattern with `Pattern.DOTALL` so the leading and trailing `.*` span line breaks. The `\b` anchors are unchanged, so single-line results and the documented examples behave the same and no false positive is introduced.

The same regex is in `commons-text` `WordUtils.containsAllWords`.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/WordUtils.java
+++ b/src/main/java/org/apache/commons/lang3/text/WordUtils.java
@@ -198,7 +198,7 @@ public class WordUtils {
             if (StringUtils.isBlank(w)) {
                 return false;
             }
-            final Pattern p = Pattern.compile(".*\\b" + Pattern.quote(w.toString()) + "\\b.*");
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @alhudz 
Please port to Apache Commons Text.
Thank you!

alhudz:
Ported in apache/commons-text#754 — same `Pattern.DOTALL` fix and a matching regression test.

PR #1735, opened 2026-06-29
Avoid building a Set on every StringUtils.indexOfAnyBut call

nishantmehta:
## Summary

`StringUtils.indexOfAnyBut(CharSequence, CharSequence)` — which backs `containsOnly` and is also called directly — collected the search characters into a `Set<Integer>` on **every call**:

```java
final Set<Integer> searchSetCodePoints =
    searchChars.codePoints().boxed().collect(Collectors.toSet());
```

allocating an `IntStream`, a boxed `Stream<Integer>`, the boxed `Integer` elements and a `HashSet` just to test membership while scanning the input.

This scans the search characters directly for each code point instead, matching the existing nested-scan approach already used by `containsAny`/`containsNone` in this same class. Iterating by code point preserves supplementary-character (surrogate-pair) correctness, so behavior is unchanged.

## Measurement

ThreadMXBean allocation driver, 300k warmed ops:

| call | before | after |
|------|--------|-------|
| `containsOnly("Hello, World!", ...)` | 744 B/op | 55 B/op (~4× faster) |
| `containsOnly(numeric25, digits)` | 744 B/op | 40 B/op (~3× faster) |

The residual ~40 B/op is the `char[]`/`CharBuffer` in the `String` overload.

## Testing

`StringUtilsTest`, `StringUtilsContainsTest` and `StringUtilsIndexOfTest` pass unchanged (266 tests).

## Notes

Like the sibling `containsAny`/`containsNone`, the direct scan is O(n·m); for the typical small `searchChars` it is both faster and allocation-free, and it is consistent with how those neighbours are already implemented.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -2926,17 +2926,34 @@ public class StringUtils {
         if (isEmpty(seq) || isEmpty(searchChars)) {
             return INDEX_NOT_FOUND;
         }
-        final Set<Integer> searchSetCodePoints = searchChars.codePoints()
-                .boxed().collect(Collectors.toSet());
         // advance character index from one interpreted codepoint to the next
         for (int curSeqCharIdx = 0; curSeqCharIdx < seq.length();) {
             final int curSeqCodePoint = Character.codePointAt(seq, curSeqCharIdx);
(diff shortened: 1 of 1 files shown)


garydgregory:
Closing. No one's asking for this. Fix bugs in Jira if want to help improve this things.

PR #1737, opened 2026-06-30
Fix Strings.CI.indexOf() returning out-of-range index for empty search

alhudz:
1. `CiStrings.indexOf` (the case-insensitive `Strings.CI.indexOf`) bails out only when `startPos > endLimit`, where `endLimit = str.length() - searchStr.length() + 1`.
2. For an empty search `endLimit` is `str.length() + 1`, so a `startPos` of exactly `str.length() + 1` slips past the guard and the empty-search branch returns `startPos` unchanged.

Repro: `Strings.CI.indexOf("abc", "", 4)`
Expected: `-1` (the class Javadoc documents out-of-range start positions as `-1`, e.g. `Strings.CI.indexOf("abc", "", 9) = -1`)
Actual: `4`, an index one past the end of a length-`3` string

The case-sensitive `CS.indexOf` delegates to `String.indexOf` and clamps, so only the hand-rolled `CI` implementation is affected. Changed the guard to `startPos >= endLimit`; `-1` is returned once `startPos` passes the last valid empty-match position (`str.length()`), and every documented example is unchanged.

---

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Strings.java
+++ b/src/main/java/org/apache/commons/lang3/Strings.java
@@ -152,7 +152,7 @@ public abstract class Strings {
                 startPos = 0;
             }
             final int endLimit = str.length() - searchStr.length() + 1;
-            if (startPos > endLimit) {
+            if (startPos >= endLimit) {
                 return INDEX_NOT_FOUND;
             }
             if (searchStr.length() == 0) {
--- a/src/test/java/org/apache/commons/lang3/StringsTest.java
+++ b/src/test/java/org/apache/commons/lang3/StringsTest.java
@@ -67,13 +67,16 @@ class StringsTest extends AbstractLangTest {
     @Test
     void testCaseInsensitiveIndexOfEmptyOutOfRange() {
         // repro: returned 4 (past the end of a length-3 string) before the fix
-        assertEquals(-1, Strings.CI.indexOf("abc", "", 4));
+        final String emptySearch = StringUtils.EMPTY;
+        assertEquals(-1, Strings.CI.indexOf("abc", emptySearch, 4));
         // documented out-of-range example, also -1
-        assertEquals(-1, Strings.CI.indexOf("abc", "", 9));
+        assertEquals(-1, Strings.CI.indexOf("abc", emptySearch, 9));
         // the end position is still a valid empty match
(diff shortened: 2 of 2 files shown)


garydgregory:
@alhudz 
FYI: I cleaned up the test a bit to make it clear we are testing for the specific edge case of an "empty search". Rebuilding...

garydgregory:
Thank you @alhudz , merged 🚀

garydgregory:
@alhudz Would you please check if there are any other issues in this class regarding empty strings? TY!

alhudz:
Had a look through the rest of the class. The empty-search paths in `lastIndexOf`, `contains`, `startsWith`/`endsWith`, `equals` and the `*Any`/`remove`/`replace` helpers all behave the same for `CI` and `CS`.

The one spot left is `indexOf` with a start position past the end:

`CI.indexOf("abc", "", 4)` -> `-1`
`CS.indexOf("abc", "", 4)` -> `3`

`CS` delegates to `String.indexOf`, which clamps the start to the length and so keeps matching the empty search; the hand-rolled `CI` returns `-1` once `startPos > length`. The shared prose says 'a start position greater than the string length only matches an empty search CharSequence', which lines up with the `CS` side. Want me to make `CI` clamp to match `CS` (and drop the `CI.indexOf("abc", "", 9) = -1` example), or keep the documented `-1`?

garydgregory:
> Had a look through the rest of the class. The empty-search paths in `lastIndexOf`, `contains`, `startsWith`/`endsWith`, `equals` and the `*Any`/`remove`/`replace` helpers all behave the same for `CI` and `CS`.
> 
> The one spot left is `indexOf` with a start position past the end:
> 
> `CI.indexOf("abc", "", 4)` -> `-1` `CS.indexOf("abc", "", 4)` -> `3`
> 
> `CS` delegates to `String.indexOf`, which clamps the start to the length and so keeps matching the empty search; the hand-rolled `CI` returns `-1` once `startPos > length`. The shared prose says 'a start position greater than the string length only matches an empty search CharSequence', which lines up with the `CS` side. Want me to make `CI` clamp to match `CS` (and drop the `CI.indexOf("abc", "", 9) = -1` example), or keep the documented `-1`?

@alhudz 
Yeah, the reply should be consistent, regardless of case.

PR #1738, opened 2026-07-01
Keep DurationFormatUtils.lexx() from merging repeated fields split by a literal

alhudz:
`Repro:` `formatDuration(Duration.ofHours(5).toMillis(), "HxH", false)` returns `5x`, and `formatDuration(Duration.ofDays(2).toMillis(), "d-d", false)` returns `2-`.
`Expected:` `5x5` and `2-2` — the field letter on each side of the literal is a separate field.
`Actual:` the trailing field is dropped and folded into the first token, whose pad width silently grows.
`Cause:` `lexx` keeps a `previous` pointer so adjacent identical field letters (`HH`) collapse into one padded token, but it never clears that pointer when a non-field character is seen. The run therefore leaks across a plain literal, a quoted literal, and `[`/`]` blocks, so `HxH` merges the two `H`s while `HmH` (separated by another field) does not.
`Fix:` reset `previous` in the `[`, `]`, quote and default branches so a run only collapses when the letters are truly adjacent. `HH` still collapses; `H'x'H`, `H[x]H` and `d-d` no longer do.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
+++ b/src/main/java/org/apache/commons/lang3/time/DurationFormatUtils.java
@@ -676,12 +676,14 @@ public class DurationFormatUtils {
                 }
                 optionalIndex++;
                 inOptional = true;
+                previous = null;
                 break;
             case ']':
                 if (!inOptional) {
                     throw new IllegalArgumentException("Attempting to close unopened optional block at index: " + i);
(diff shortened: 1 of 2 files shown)


sarankumarbaskar (approved):

garydgregory:
I'll review today or tomorrow...

PR #1740, opened 2026-07-03
Stop ExtendedMessageFormat seekNonWs reading past the pattern end

dxbjavid:
Port of apache/commons-text#759 to the equivalent org.apache.commons.lang3.text.ExtendedMessageFormat.

The registry-based ExtendedMessageFormat constructor parses its pattern with seekNonWs, which asks the split matcher about the character at the current parse index without first checking it is still inside the buffer. A truncated format element such as {, {0, or trailing text like {0}extra{ leaves the parse position at the end of the pattern, so the matcher reads one past the char array and the constructor throws ArrayIndexOutOfBoundsException rather than the documented IllegalArgumentException that the plain constructor already gives for the same input.

This bounds the seek to the pattern length so those patterns report the usual unterminated-format-element error. A regression test covers the three shapes above, alongside the JRE sanity checks you added post-merge on the commons-text side.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -46,6 +46,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
   <release version="3.21.0" date="YYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
     <!-- FIX -->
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @dxbjavid , merged 🚀 TY!

dxbjavid:
Thank you @garydgregory for the review and merge. I appreciate your guidance in porting the fix from Commons Text to Commons Lang. Glad to contribute, and thanks again!

PR #1744, opened 2026-07-07
Keep ExtendedMessageFormat from rejecting a quoted format style

alhudz:
`ExtendedMessageFormat.parseFormatDescription` walks a format element's style with a `for` loop whose update clause always calls `next(pos)`, but the `QUOTE` branch hands off to `getQuotedString`, which already leaves `pos` one past the closing quote. The loop then advances a second time and skips the character right after that quote, so when a style ends with a quoted literal the terminating `}` is never seen: `depth` never returns to `0`, the loop runs off the end, and the constructor throws `IllegalArgumentException: Unterminated format element`. This only happens on the registry path, so `new ExtendedMessageFormat("{0,number,0'%'}", registry)` throws even though `java.text.MessageFormat` and `ExtendedMessageFormat` without a registry both parse that pattern to `5%`.

The fix switches the loop to the `while` form the two sibling parsers in this class already use (`applyPattern`, `insertFormats`): advance one position per non-quote branch and let the `QUOTE` branch's `getQuotedString` do its own advance. Leaving the position arithmetic to the single loop that consumes the quoted run is what stops the double count. The added test fails before the change with the `Unterminated format element` exception and passes after.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/ExtendedMessageFormat.java
+++ b/src/main/java/org/apache/commons/lang3/text/ExtendedMessageFormat.java
@@ -382,21 +382,24 @@ public class ExtendedMessageFormat extends MessageFormat {
         seekNonWs(pattern, pos);
         final int text = pos.getIndex();
         int depth = 1;
-        for (; pos.getIndex() < pattern.length(); next(pos)) {
+        while (pos.getIndex() < pattern.length()) {
             switch (pattern.charAt(pos.getIndex())) {
             case START_FE:
                 depth++;
+                next(pos);
                 break;
             case END_FE:
                 depth--;
                 if (depth == 0) {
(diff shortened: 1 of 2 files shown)


garydgregory:
@alhudz Thank you, please port to Apache Commons Text if needed.

alhudz:
Checked the Commons Text version: its `parseFormatDescription` already uses the `while` form, with `getQuotedString` doing the advance in the `QUOTE` branch, so the double advance isn't there. Verified against commons-text 1.14.0: `new ExtendedMessageFormat("{0,number,0'%'}", registry).format(new Object[]{5})` -> `5%`. Nothing to port. Thanks for merging.

garydgregory:
@alhudz 
You'll want to check against git master, not a release tag.

PR #1745, opened 2026-07-08
Match String case folding in StrBuilder.equalsIgnoreCase

alhudz:
`Repro`: `new StrBuilder("K").equalsIgnoreCase(new StrBuilder("K"))` (KELVIN SIGN vs `K`) returns `false`, and so does the å/ANGSTROM pair `new StrBuilder("å").equalsIgnoreCase(new StrBuilder("Å"))`. `String.equalsIgnoreCase` returns `true` for both.

`Cause`: the per-character loop only rules a differing pair equal when `Character.toUpperCase(c1) == Character.toUpperCase(c2)`. Characters that fold together only through their lower-case mapping (`U+212A` → `k`, `U+212B` → `å`) fail that test, so the whole compare reports not-equal.

`Fix`: also accept the pair when `Character.toLowerCase` of the upper-cased chars agrees, the same fold `String#regionMatches` and this library's own `CharSequenceUtils.equalsIgnoreCase` already use. The case-sensitive `equals` sibling is untouched.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
@@ -1912,8 +1912,12 @@ public class StrBuilder implements CharSequence, Appendable, Serializable, Build
         for (int i = size - 1; i >= 0; i--) {
             final char c1 = thisBuf[i];
             final char c2 = otherBuf[i];
-            if (c1 != c2 && Character.toUpperCase(c1) != Character.toUpperCase(c2)) {
-                return false;
+            if (c1 != c2) {
+                final char u1 = Character.toUpperCase(c1);
+                final char u2 = Character.toUpperCase(c2);
+                if (u1 != u2 && Character.toLowerCase(u1) != Character.toLowerCase(u2)) {
+                    return false;
+                }
             }
         }
         return true;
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , merged 🚀 
Since this class is deprecated in Commons Lang (but bugs should still be fixed) in favor of Commons Text's `TextStringBuilder`, would you please verify parity there?

alhudz:
Checked, parity is not there. `TextStringBuilder.equalsIgnoreCase` has the same upper-case-only fold (line 1911 on current master).

`Repro` against commons-text 1.15.0: `new TextStringBuilder("K").equalsIgnoreCase(new TextStringBuilder("K"))` (KELVIN SIGN vs `K`) -> `false` while `String` -> `true`; same for `å` vs ANGSTROM SIGN `Å` -> `false` vs `true`. The same one-line `toLowerCase` fold from this PR fixes it there too.

Thanks for merging.

garydgregory:
Would you please create a PR for Commons Text?

alhudz:
Done: apache/commons-text#762, same `toLowerCase` fold plus the Kelvin/Angstrom regression assertions in `TextStringBuilderTest`.

PR #1749, opened 2026-07-13
Fix sign extension in BitField.getValue for top-bit fields

alhudz:
`BitField.getValue(int)` and `getValue(long)` read the field with `getRawValue` and then shift it down with an arithmetic `>>`. When the field sits on the top bit of the holder (bit 31 for the `int` overload, bit 63 for the `long` overload added in 3.21.0) the masked value is negative, so `>>` sign-extends and the field reads back negative. `new BitField(0x8000000000000000L).getValue(0x8000000000000000L)` gives `-1` for a one-bit field whose value is `1`, and `new BitField(0xFF000000).getValue(0xFF000000)` gives `-1` from the `int` overload while `getValue(long)` returns `255` for the same field.

Use the unsigned `>>>` in both getters so the selected bits come down as magnitude. The value is reconstructed at the shift, so fixing it there also covers `getShortValue`/`getByteValue`, which delegate through `getValue`, and it leaves the mask handling from #1711 untouched. Fields that do not reach the top bit are unchanged because `>>` and `>>>` agree on a non-negative value.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/BitField.java
+++ b/src/main/java/org/apache/commons/lang3/BitField.java
@@ -196,7 +196,7 @@ public class BitField {
      * @see #setValue(int,int)
      */
     public int getValue(final int holder) {
-        return getRawValue(holder) >> shiftCount;
+        return getRawValue(holder) >>> shiftCount;
     }
 
     /**
@@ -212,7 +212,7 @@ public class BitField {
      * @since 3.21.0
      */
     public long getValue(final long holder) {
-        return getRawValue(holder) >> shiftCount;
+        return getRawValue(holder) >>> shiftCount;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Good find @alhudz , will merge, do you see other issues in the code base wit bit shifts?

alhudz:
Thanks for merging this. Nothing new at the moment: the other shift issues I'd found are already in (#1711 for the int mask sign extension in the `long` methods, #1751 for the `FluentBitSet.setInclusive` overflow). I had another look at `Conversion` and `CachedRandomBits`; both mask after the arithmetic shift (`0xff & src >> shift` style), so sign extension can't leak out there. If anything else turns up at the top-bit boundaries I'll open a PR.

PR #1750, opened 2026-07-16
Fix ArrayUtils.reverse range underflow on Integer.MIN_VALUE end

alhudz:
Repro: `ArrayUtils.reverse(new int[]{1, 2, 3}, 0, Integer.MIN_VALUE)`.
Expected: no change, since the `endIndexExclusive` Javadoc documents an undervalue `< start index` as no change.
Actual: `ArrayIndexOutOfBoundsException: Index 2147483647 out of bounds for length 3`.
Cause: `Math.min(array.length, endIndexExclusive) - 1` keeps `Integer.MIN_VALUE` and the `- 1` underflows to `Integer.MAX_VALUE`, so the `while (j > i)` loop indexes past the array. All nine primitive and `Object` range overloads share the line.
Fix: clamp the end to zero before the decrement with the existing `max0` helper, as the sibling `subarray` and `shift` methods already do. Every other negative end was already a no-op, so valid ranges are unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ArrayUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ArrayUtils.java
@@ -6721,7 +6721,7 @@ public class ArrayUtils {
             return;
         }
         int i = Math.max(startIndexInclusive, 0);
-        int j = Math.min(array.length, endIndexExclusive) - 1;
+        int j = max0(Math.min(array.length, endIndexExclusive)) - 1;
         boolean tmp;
         while (j > i) {
             tmp = array[j];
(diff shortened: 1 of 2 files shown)


garydgregory:
Good find @alhudz, merged 🚀  Do you see any other similar index issue in this class?

alhudz:
Thanks for the merge. I went back over the class with the same `Integer.MIN_VALUE`/`MAX_VALUE` probes and didn't find another escape: `subarray` already clamps with `max0`, `shift` returns early on `endIndexExclusive <= 0` before any subtraction, `swap` runs `max0` on both offsets and bounds `len` by the remaining lengths, and `indexOf`/`lastIndexOf` (including the NaN and tolerance variants) either clamp or bail on a negative start. `add(index)`, `insert` and `remove` throw `IndexOutOfBoundsException` by documented contract, so `reverse` looks to have been the only spot where an extreme value got past the clamps.

garydgregory:
@alhudz thank you for checking.

PR #1754, opened 2026-07-26
[LANG-1830] Avoid repeated regex compilation in ClassUtils.toCleanName()

kzhunmax:
## Summary

Replace the repeated use of String.matches() in ClassUtils.toCleanName() with a precompiled static Pattern.

### Motivation

String.matches() recompiles the regular expression on every invocation.
Reusing a compiled Pattern avoids this overhead while preserving the existing behavior.

A local JMH benchmark showed approximately 4x-9x higher throughput for the validation step after this change.

Related: LANG-1830

### DIFF
--- a/src/main/java/org/apache/commons/lang3/ClassUtils.java
+++ b/src/main/java/org/apache/commons/lang3/ClassUtils.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


kzhunmax:
The Jira ticket is [LANG-1830](https://issues.apache.org/jira/browse/LANG-1830)

garydgregory:
Thank you @kzhunmax , merged 🚀 Since this is a simple and straightforward conversion, a JMH benchmark is not required for inclusion in this PR. For other non-trivial claims, we would want to verify and weight any such claim.

PR #1755, opened 2026-07-27
[LANG-1831] Add missing test for ObjectUtils.isEmpty(Object)

kzhunmax:
Add a missing unit test for ObjectUtils.isEmpty(Object) to verify that
unsupported object types (for example, Integer) are reported as non-empty,
matching the documented behavior.


kzhunmax:
Jira ticket is [LANG-1831](https://issues.apache.org/jira/browse/LANG-1831)

garydgregory:
@kzhunmax 
You don't need a Jira ticket for this type of tiny test change.

PR #1758, opened 2026-07-29
Keep abbreviateMiddle off surrogate pair boundaries

kdelay:
Repro: with U+1F600 GRINNING FACE (a surrogate pair) as `EMOJI`, `StringUtils.abbreviateMiddle("a" + EMOJI + EMOJI + "b", ".", 4)` returns `a` plus a lone high surrogate before the middle, and `StringUtils.abbreviateMiddle(EMOJI + "abc" + EMOJI, "..", 5)` returns a lone low surrogate after the middle.
Cause: the head and tail cut offsets are computed in `char` units and sliced with `String.substring`, so a cut that lands between the two halves of a surrogate pair leaves an unpaired surrogate in the result. #1719 fixed the same defect in `abbreviate` and `truncate`, but the sibling `abbreviateMiddle` was left out.
Fix: reuse the `splitsSurrogatePair` helper introduced by #1719 to nudge each cut to the nearest code-point edge before slicing, so a pair is never split. The result still never exceeds `length`, and cuts that already sit between whole code points are unchanged. The new test fails on the current code and passes with the fix.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -429,8 +429,15 @@ public class StringUtils {
             return str;
         }
         final int targetString = length - middle.length();
-        final int startOffset = targetString / 2 + targetString % 2;
-        final int endOffset = str.length() - targetString / 2;
+        int startOffset = targetString / 2 + targetString % 2;
+        int endOffset = str.length() - targetString / 2;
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @kdelay , merged 🚀

kdelay:
Thanks for the quick merge, @garydgregory.

PR #1759, opened 2026-07-29
Fix StrBuilder.lastIndexOf(String) start index for an empty search

alhudz:
1. Cause: `lastIndexOf(String)` delegates to `lastIndexOf(String, int)` with `size - 1`, but the last valid start position for an empty search string is `size`.
2. Change: pass `size`. Non-empty searches are unaffected, because their last possible match start is `size - str.length()`, which is at most `size - 1`.

Repro:

```java
new StrBuilder("abc").lastIndexOf("")     // expected 3, actual 2
new StrBuilder("").lastIndexOf("")        // expected 0, actual -1
new StrBuilder("abc").lastIndexOf("", 3)  // 3, the two-arg overload is already correct
new StrBuilder("abc").indexOf("")         // 0
new StrBuilder("abc").contains("")        // true
```

`String.lastIndexOf`, `StringBuilder.lastIndexOf` and `StringUtils.lastIndexOf` all return the length for an empty search, so `StrBuilder` disagreed with both the JDK and the rest of Lang, and with its own `indexOf` and `contains`. Found by differential fuzzing of `StrBuilder` against `StringBuilder`; over 400k random cases this was the only method that diverged. The new assertions in `StrBuilderTest.testLastIndexOf_String` fail without the runtime change (`expected: <4> but was: <3>`).

`TextStringBuilder` in Commons Text (checked against `master`) has the same `size - 1` in `lastIndexOf(String)`, and its hand-rolled two-arg overload clamps `startIndex` to `size - 1` before the empty-search early return, so both overloads diverge there. Glad to raise a matching PR if that is useful.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
+++ b/src/main/java/org/apache/commons/lang3/text/StrBuilder.java
@@ -2362,7 +2362,7 @@ public class StrBuilder implements CharSequence, Appendable, Serializable, Build
      * @return The last index of the string, or -1 if not found.
      */
     public int lastIndexOf(final String str) {
-        return lastIndexOf(str, size - 1);
+        return lastIndexOf(str, size);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , merged 🚀 Please check [Apache Commons Text](https://github.com/apache/commons-text):
- `org.apache.commons.text.StrBuilder`
- `org.apache.commons.text.TextStringBuilder`

PR #1762, opened 2026-07-30
Fix negative week year formatting in FastDatePrinter

alhudz:

`Repro`: `FastDateFormat.getInstance("YYYY").format(cal)` where `cal` is 15 July 42 BC, that is a `GregorianCalendar` with `ERA` set to `GregorianCalendar.BC`.
`Expected`: `-0041`, what `SimpleDateFormat` prints for the same date and pattern.
`Actual`: `000` followed by `U+0007`. `YY` gives `,/` against `SimpleDateFormat`'s `-41`.

1. `GregorianCalendar.getWeekYear()` is proleptic, so a BC date gives a negative week year (`-41` here) and `WeekYear` hands it to the wrapped `NumberRule` unchanged.
2. The digit rules build each character arithmetically (`(char) (value + '0')`, `value / 10 + '0'`), so a negative value walks below `'0'` and lands on control characters or unrelated code points rather than digits.

`Fix`: emit the sign in `WeekYear` and pass the magnitude down, the same handling `TimeZoneNumberRule` and `Iso8601_Rule` already apply to negative UTC offsets. Every other rule takes its value from `Calendar.get`, which never returns a negative, so `getWeekYear` is the only source of one; `YYY` and `Y` keep the four-digit padding the class Javadoc describes.

Checked `YYYY` against `SimpleDateFormat` over 46,667 dates spanning 249 BC to 4188 AD: 2,611 mismatches before this change, 337 of them carrying a control character, and 0 after.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/time/FastDatePrinter.java
+++ b/src/main/java/org/apache/commons/lang3/time/FastDatePrinter.java
@@ -864,11 +864,16 @@ public class FastDatePrinter implements DatePrinter, Serializable {
         public void appendTo(final Appendable buffer, final Calendar calendar) throws IOException {
             // Some Calendar implementations (JapaneseImperialCalendar) do not support week-dates.
             // Fall back to Calendar.YEAR in that case.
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , merged 🚀

PR #1771, opened 2026-08-11
Fix low surrogate false match in containsAny/containsNone/indexOfAny

alhudz:
`containsAny`, `containsNone` and `indexOfAny` share one surrogate-aware guard that only special-cases a `high` surrogate.

1. when the matched char is a `low` surrogate the `!isHighSurrogate(ch)` term returns a match without checking the preceding high surrogate, so a low surrogate in the search set matches the low half of a `different` pair in the input.
2. `StringUtils.containsAny("😀", "🨀".toCharArray())` is `true` and `indexOfAny` returns `1` although the inputs are U+1F600 and U+1FA00, distinct code points that only share the low surrogate `\uDE00`; index `1` also lands inside the pair, so slicing there yields malformed UTF-16.

Extended the guard so a low surrogate that is the trailing half of a pair in the search set only matches when the input carries the same high surrogate immediately before. Lone surrogates, BMP chars and matching pairs are unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -1091,7 +1091,9 @@ public class StringUtils {
             final char ch = cs.charAt(i);
             for (int j = 0; j < searchLength; j++) {
                 if (searchChars[j] == ch) {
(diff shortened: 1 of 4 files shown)


garydgregory:
Thank you @alhudz , merged 🚀

alhudz:
thanks for landing this one

PR #1773, opened 2026-08-14
Keep StringUtils stripStart and stripEnd off surrogate pair boundaries

alhudz:
Repro: `StringUtils.stripStart("𝔞abc", "𝔟")` — input starts with U+1D51E, the strip set holds only U+1D51F, and the two share the high surrogate `\uD835`.
Expected: `"𝔞abc"` unchanged, since U+1D51E is not in `{U+1D51F}`.
Actual: `"\uDD1Eabc"` — only the high surrogate is removed, so a well-formed input becomes ill-formed UTF-16. `stripEnd`/`strip`/`stripAll` corrupt the same way against a supplementary strip character that shares either surrogate half (for example `stripEnd("abc𝄞", "𝔞")` returns `"abc\uD834"`).
Cause: both membership loops test one `char` at a time with `stripChars.indexOf(str.charAt(...))`, and `String.indexOf(char)` matches a single surrogate half of a supplementary set character.
Fix: walk the strip membership by code point (`codePointAt`/`codePointBefore` + `Character.charCount`) so a supplementary strip character only matches a whole code point and the cut can never land inside a pair. The already surrogate-aware `containsAny`/`containsNone`/`indexOfAny`/`indexOfAnyBut` agree the character is not a member on these inputs. BMP input and the `null` (whitespace) path are unchanged.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/StringUtils.java
+++ b/src/main/java/org/apache/commons/lang3/StringUtils.java
@@ -8091,8 +8091,12 @@ public class StringUtils {
         } else if (stripChars.isEmpty()) {
             return str;
         } else {
-            while (end != 0 && stripChars.indexOf(str.charAt(end - 1)) != INDEX_NOT_FOUND) {
-                end--;
+            while (end != 0) {
+                final int codePoint = str.codePointBefore(end);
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you @alhudz , merged 🚀

PR #1775, opened 2026-08-16
Fix CharRange.contains(CharRange) for negated argument ranges

alhudz:
Repro: `CharRange.isIn('a', Character.MAX_VALUE).contains(CharRange.isNotIn((char) 0, (char) ('a' - 1)))` returns `false`, but the argument denotes `['a', MAX_VALUE]`, the same set as the receiver, so it should be `true`. `CharRange.is('x').contains(CharRange.isNotIn((char) 0, Character.MAX_VALUE))` also returns `false` although the argument is the empty set.
Cause: in the non-negated-receiver / negated-argument branch, `contains(CharRange)` only accepts a negated argument when the receiver is the full `[0, MAX_VALUE]` range. A negated range denotes `[0, start - 1]` union `[end + 1, MAX_VALUE]`, which is empty when it spans the whole domain and a single contiguous interval when its excluded range touches `0` or `MAX_VALUE`; a smaller receiver can hold those.
Fix: handle the empty-set and single-interval cases in that branch, keeping the full-range requirement only when the excluded range is strictly interior. The other three negation combinations are left as they were.

Checked against the per-character oracle `contains(char)` over a boundary sweep: 422 mismatches before, 0 after. Regression assertions added in `CharRangeTest`.

### DIFF
--- a/src/test/java/org/apache/commons/lang3/CharRangeTest.java
+++ b/src/test/java/org/apache/commons/lang3/CharRangeTest.java
@@ -261,6 +261,33 @@ class CharRangeTest extends AbstractLangTest {
     }
 
     @Test
+    void testContains_Charrange_negatedArgumentTouchingBounds() {
+        // A negated argument denotes [0, start - 1] union [end + 1, MAX_VALUE]. When its excluded range touches 0 or
+        // MAX_VALUE that set collapses to the empty set or a single contiguous interval, so a non-full range can contain it.
+        final CharRange emptySet = CharRange.isNotIn((char) 0, Character.MAX_VALUE);
+        assertTrue(CharRange.is('x').contains(emptySet));
+        assertTrue(CharRange.isIn((char) 0, Character.MAX_VALUE).contains(emptySet));
+
+        // isNotIn(0, 'a' - 1) denotes ['a', MAX_VALUE]
+        final CharRange fromA = CharRange.isNotIn((char) 0, (char) ('a' - 1));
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/lang3/CharRangeTest.java
```diff
@@ -260,6 +260,33 @@ void testContains_Charrange() {
         assertTrue(notbd.contains(notae));
     }
 
+    @Test
+    void testContains_Charrange_negatedArgumentTouchingBounds() {
+        // A negated argument denotes [0, start - 1] union [end + 1, MAX_VALUE]. When its excluded range touches 0 or
+        // MAX_VALUE that set collapses to the empty set or a single contiguous interval, so a non-full range can contain it.
+        final CharRange emptySet = CharRange.isNotIn((char) 0, Character.MAX_VALUE);
```
garydgregory, line 267: You can use `Character.MIN_VALUE `instead of `0`.

garydgregory:
Thank you @alhudz  Merged 🚀

src/test/java/org/apache/commons/lang3/CharRangeTest.java
alhudz: Good point, `Character.MIN_VALUE` reads better than the cast. I'll use it in future tests.

alhudz:
Thanks for merging this one.

PR #1779, opened 2026-08-24
Fix int overflow in Conversion count and position bounds guards

alhudz:
Repro: call any `Conversion` array/hex/binary converter with a large count or position, e.g. `Conversion.binaryToInt(new boolean[]{true}, 0, 0, Integer.MAX_VALUE, 2)`, `Conversion.byteArrayToInt(new byte[]{1}, 0, 0, 0, Integer.MAX_VALUE)`, or `Conversion.intToHex(0, 0, "", 0, Integer.MAX_VALUE)`.

Cause: each method documents `@throws IllegalArgumentException` when the requested count and position exceed the destination width (`nBools - 1 + dstPos >= 32`, `(nBytes - 1) * 8 + dstPos >= 32`, and so on) and evaluates that guard first, but in `int`. For a large count (`nBools`/`nBytes`/`nHex`/`nInts`/`nShorts`) or a large `srcPos`/`dstPos` the expression overflows and wraps negative, so the guard is skipped.

Fix: widen the leading operand of each guard to `long` (`(long) nBools - 1 + dstPos`, `((long) nBytes - 1) * 8 + dstPos`, ...) so the documented condition holds. Small valid inputs are unchanged.

Without the fix the reading methods throw an undocumented `ArrayIndexOutOfBoundsException`/`StringIndexOutOfBoundsException` from the loop, and the hex writers (`intToHex`/`longToHex`/`shortToHex`/`byteToHex`) skip the guard entirely and build a several-hundred-million-char `String`. Applied to all 28 guard sites in the class; `ConversionTest.testLargeCountThrowsIllegalArgumentException` exercises each and fails before the change.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/Conversion.java
+++ b/src/main/java/org/apache/commons/lang3/Conversion.java
@@ -160,7 +160,7 @@ public class Conversion {
         if (src.length == 0 && srcPos == 0 || 0 == nBools) {
             return dstInit;
         }
-        if (nBools - 1 + dstPos >= Byte.SIZE) {
+        if ((long) nBools - 1 + dstPos >= Byte.SIZE) {
             throw new IllegalArgumentException("nBools - 1 + dstPos >= 8");
         }
         byte out = dstInit;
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you for the PR @alhudz , merged 🚀

alhudz:
thanks for the quick merge, glad the guards are sorted

PR #1782, opened 2026-09-03
Fix TypeUtils.isAssignable() for wildcards with multiple upper bounds

Alwaysgaurav1:
## Description
Fixes an issue in `TypeUtils.isAssignable(Type, WildcardType, Map)` where checking assignability of a `WildcardType` with multiple upper bounds (intersection types such as `? extends Serializable & Cloneable`) to another `WildcardType` (such as `? extends Serializable`) incorrectly returned `false`.

### Root Cause
When the subject `type` is a `WildcardType`, the upper bounds loop previously enforced that **every** upper bound in the subject wildcard had to be assignable to each target upper bound `toBound` (`forall bound in upperBounds: isAssignable(bound, toBound)`):

```java
for (Type toBound : toUpperBounds) {
    toBound = substituteTypeVariables(toBound, typeVarAssigns);
    for (final Type bound : upperBounds) {
        if (!isAssignable(bound, toBound, typeVarAssigns)) {
            return false;
        }
    }
}

### DIFF
--- a/src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java
+++ b/src/main/java/org/apache/commons/lang3/reflect/TypeUtils.java
@@ -1229,13 +1229,17 @@ public class TypeUtils {
                 // if there are assignments for unresolved type variables,
                 // now's the time to substitute them.
                 toBound = substituteTypeVariables(toBound, typeVarAssigns);
-                // each upper bound of the subject type has to be assignable to
-                // each
-                // upper bound of the target type
+                // at least one upper bound of the subject type has to be assignable to
+                // each upper bound of the target type
+                boolean satisfied = false;
                 for (final Type bound : upperBounds) {
-                    if (!isAssignable(bound, toBound, typeVarAssigns)) {
-                        return false;
+                    if (isAssignable(bound, toBound, typeVarAssigns)) {
+                        satisfied = true;
+                        break;
                     }
+                }
+                if (!satisfied) {
+                    return false;
                 }
             }
             for (Type toBound : toLowerBounds) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Alwaysgaurav1** [comment]
@garydgregory, please review it quickly.

**Comment by garydgregory** [comment]
@Alwaysgaurav1 
Excuse me, quickly? Why? You're not the only PR author on the planet you know.

**Comment by Alwaysgaurav1** [comment]
> @garydgregory,

Okay, whenever you want . I am okay.

**Comment by garydgregory** [comment]
@Alwaysgaurav1 
Until copilot comes back with its review, here are a couple of findings to address:

- Lower-bound behavior changes without a regression test.

At updated `TypeUtils` lines 1251–1259, the patch changes “every source lower bound” to “at least one source lower bound.” This is outside the upper-bound problem described by the PR.

See:
```
TypeUtils.isAssignable(
    TypeUtils.wildcardType()
        .withLowerBounds(Number.class, CharSequence.class).build(),
    TypeUtils.wildcardType()
        .withLowerBounds(Integer.class).build())
```

The result changes from false to true. This is not itself evidence of incorrect behavior, but it needs an explicit rationale and positive/negative tests. Either include those or leave the lower-bound change for a separate PR.

- The new test does not protect the “every target bound” requirement.

All four assertions use a target with one upper bound. Add a target with multiple bounds where only one is satisfied, and another where all are satisfied. Also test the reverse direction: `? extends Serializable` must not become assignable to the synthetic `Serializable & Cloneable` wildcard.
These cases protect the distinction between “for every target bound, some source bound matches” and the incorrect “any matching pair is enough.”

Alwaysgaurav1:
@garydgregory 

Thank you for the feedback! I have updated the PR to address both points:

1. **Reverted Lower-Bound Modifications**: Reverted the lower-bound checking logic back to the original strict matching to keep this PR strictly focused on multiple upper bounds.
2. **Enhanced Test Coverage in `TypeUtilsTest`**:
   - Added multi-bound targets where all bounds are satisfied (`? extends Serializable & Cloneable` assignable to `? extends Serializable & Cloneable` and `? extends Object & Serializable`).
   - Added multi-bound targets where only one bound is satisfied (`? extends Serializable & Cloneable` is NOT assignable to `? extends Serializable & CharSequence`).
   - Added reverse-direction tests (`? extends Serializable` is NOT assignable to `? extends Serializable & Cloneable`).

PR #1784, opened 2026-09-07
Fix Fraction.add and subtract for operands not in lowest terms

lenamonj:
`Fraction.addSub` runs Knuth 4.5.1 on the operands as given, but the algorithm assumes lowest terms and `getFraction(50, 100)` is stored as 50/100. A shared factor inside an operand survives into the common denominator, so `add` and `subtract` throw `ArithmeticException` while the exact result fits an int (1073741823/2147483646 plus 3/5 is 11/10) and return unreduced results against the reduced form the javadoc promises. #1769 fixed the same shape in `multiplyBy`.

Both operands are now reduced before the Knuth step, as `multiplyBy` does. Operands already in lowest terms take the identical path, so the existing overflow and identity assertions hold.

Test: `testAddSubtractUnreducedOperands`, red on master and green with the change. Default Maven goal green.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude (Anthropic) found the defect and drafted the fix and test; I reviewed them, reproduced the failure on a fresh clone of master, and ran the default Maven goal.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -556,23 +556,31 @@ public final class Fraction extends Number implements Comparable<Fraction> {
         if (fraction.numerator == 0) {
             return this;
         }
+        // Knuth 4.5.1 assumes operands in lowest terms and this class does not reduce on
+        // construction, so reduce both first, as multiplyBy does.
+        final int thisGcd = greatestCommonDivisor(numerator, denominator);
+        final int thatGcd = greatestCommonDivisor(fraction.numerator, fraction.denominator);
+        final int thisNumerator = numerator / thisGcd;
+        final int thisDenominator = denominator / thisGcd;
+        final int thatNumerator = fraction.numerator / thatGcd;
+        final int thatDenominator = fraction.denominator / thatGcd;
         // if denominators are randomly distributed, d1 will be 1 about 61%
         // of the time.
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @lenamonj 
Thank you for the PR, the main changes look good but I think you need more tests, specifically:

- Extend the regression tests for _subtraction_. The new test covers both denominator branches for addition, but its only subtraction example has coprime denominators after reduction. Add an unreduced subtraction case with shared reduced denominators, for example `2/4 - 2/12 = 1/3`, and a cancellation case such as `2/4 - 3/6 = 0/1`.
- Exercise both operands being unreduced and minimum-integer normalization. Examples such as `2/4 + 2/6 = 5/6` and `Integer.MIN_VALUE/2 + 2/4 = -2147483647/2` directly protect the two new normalization steps and their signed boundary behavior.
- Bonus: Clarify the existing reduced-result promise around zero. The new normalization runs after the zero shortcuts in [`addSub` (line 550)](/Users/garygregory/git/commons/commons-lang/src/main/java/org/apache/commons/lang3/math/Fraction.java:550). Consequently, `ZERO.add(2/4)` still returns `2/4`. Also, `ZERO.subtract(Integer.MIN_VALUE/2)` still throws even though its reduced result, `1073741824/1`, fits. These are pre-existing limitations, not regressions.

Thank you!

garydgregory:
Thank you @lenamonj , looks good, merged 🚀

PR #1787, opened 2026-09-09
Fraction.add and subtract return the reduced form when an operand is zero

lenamonj:
Follow-up to #1784. The zero shortcuts in `addSub` returned the other operand as constructed, so `ZERO.add(2/4)` was `2/4` against the Javadoc's reduced-form promise, and `ZERO.subtract(Integer.MIN_VALUE/2)` threw although the reduced result `1073741824/1` fits, both noted in the #1784 review. The shortcuts now reduce before returning or negating. The new test covers both operations with zero on either side and the `Integer.MIN_VALUE` boundary; it fails on master and passes with the change. Default `mvn` goal green with `-Ddoclint=all`.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude (Anthropic) wrote the fix and the test from the #1784 review and ran the reproduction and the default Maven goal on a fresh clone; I directed and reviewed the work.

### DIFF
--- a/src/main/java/org/apache/commons/lang3/math/Fraction.java
+++ b/src/main/java/org/apache/commons/lang3/math/Fraction.java
@@ -551,10 +551,10 @@ public final class Fraction extends Number implements Comparable<Fraction> {
         Objects.requireNonNull(fraction, "fraction");
         // zero is identity for addition.
         if (numerator == 0) {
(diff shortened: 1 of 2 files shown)


garydgregory:
@lenamonj Looks good, thank you, merged 🚀

PR #1793, opened 2026-09-18
Fix ConstructorUtils.getMatchingAccessibleConstructor on non-public classes

Alwaysgaurav1:
## Summary

Fixes `ConstructorUtils.getMatchingAccessibleConstructor` and `ConstructorUtils.invokeConstructor` to properly enforce class accessibility checks on non-public classes (e.g. package-private classes or public inner classes declared inside non-public outer classes), achieving consistency across exact matching, inexact matching, and `MethodUtils` (following PR #1783).

## Problem

In `ConstructorUtils.getMatchingAccessibleConstructor(Class<T> cls, Class<?>... parameterTypes)`:
1. The exact-match fast-path attempted to resolve `cls.getConstructor(parameterTypes)` directly.
2. In Java reflection, `Class.getConstructor(...)` returns any constructor with the `public` modifier, even if declaring class `cls` is package-private or enclosed in a non-public outer class.
3. The fast-path called `MemberUtils.setAccessibleWorkaround(...)` directly and returned the constructor without verifying whether `cls` is accessible.
4. Conversely, the fallback loop for inexact assignment-compatible matching filtered each candidate through `getAccessibleConstructor(ctor)`, which checks `MemberUtils.isAccessible(ctor) && isAccessible(ctor.getDeclaringClass())`.

This created inconsistent behavior:
- `ConstructorUtils.getAccessibleConstructor(PackagePrivateBean.class, exactTypes)` returned `null`.
- `ConstructorUtils.getMatchingAccessibleConstructor(PackagePrivateBean.class, inexactTypes)` returned `null`.
- But `ConstructorUtils.getMatchingAccessibleConstructor(PackagePrivateBean.class, exactTypes)` bypassed the check and returned the constructor.
- Consequently, `ConstructorUtils.invokeConstructor(PackagePrivateBean.class, exactArgs)` attempted reflective instantiation on non-public classes (violating access control and throwing `IllegalAccessException` or `InaccessibleObjectException` under JPMS), whereas inexact arguments threw `NoSuchMethodException("No such accessible constructor on object: ...")`.

## Changes

1. **`ConstructorUtils.java`**:
   - In `getMatchingAccessibleConstructor`:
     - Added an early return `if (!isAccessible(cls)) return null;`. Since constructors are never inherited and always belong to their declaring class, an inaccessible class can never have an accessible constructor.
     - In the exact-match `try` block, verified the candidate using `getAccessibleConstructor(cls.getConstructor(parameterTypes))` before returning.
   - In `getAccessibleConstructor(Class<T> cls, Class<?>... parameterTypes)`:
     - Added early return `if (!isAccessible(cls)) return null;` for consistent fast-path checking.

2. **`ConstructorUtilsTest.java`**:
   - Added `testGetMatchingAccessibleConstructorOnNonPublicClass` asserting that `getMatchingAccessibleConstructor` returns `null` for non-public classes and public inner classes of non-public outer classes (for both exact and inexact signatures).
   - Added `testInvokeConstructorOnNonPublicClass` and `testInvokeExactConstructorOnNonPublicClass` asserting that `NoSuchMethodException` is consistently thrown.


Alwaysgaurav1:
Hi @garydgregory,

This is a companion fix to #1783 (`MethodUtils.invokeMethod on instances of non-public classes`). 

While #1783 addressed method lookup on non-public classes, `ConstructorUtils.getMatchingAccessibleConstructor` had the identical issue in its exact-match fast path, bypassing `getAccessibleConstructor` checks on non-public classes. This PR brings `ConstructorUtils` into full alignment.

garydgregory:
@Alwaysgaurav1 
These change now cause the documentation to no longer match the behavior. Specifically, you should update the class-level Javadoc to match the new behavior. The new accessibility guard rejects package-private classes immediately, but the "Known Limitations" section still promises a `setAccessible(true)` workaround for their public constructors. Remove or revise that explanation and document that these calls now return `null` or throw `NoSuchMethodException`.

garydgregory (changes requested):
See my comment.

Alwaysgaurav1:
Thanks for the review, @garydgregory! I have updated the class-level Javadoc in ConstructorUtils to remove the obsolete setAccessible(true) workaround description and explicitly document that constructor lookup methods return null and invocation methods throw NoSuchMethodException for non-public classes.

PR #1794, opened 2026-09-18
LANG-1834 Fix Fraction reduction for Integer.MIN_VALUE

makarandhinge:

- [ ] I used AI to create any part of, or all, of this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?

Fix `Fraction.getReducedFraction` handling for `Integer.MIN_VALUE` with a negative denominator.

When the numerator is `Integer.MIN_VALUE` and the denominator is negative, the current implementation can attempt to negate `Integer.MIN_VALUE`, which overflows and throws an `ArithmeticException` even when the fraction can be reduced to a valid integer fraction.

This change reduces common powers of two from the numerator and denominator before sign normalization, avoiding the overflow for representable fractions while preserving overflow handling for results that cannot be represented as `int`.

## Tests

Added regression coverage for:

- `Integer.MIN_VALUE / -2`
- `Integer.MIN_VALUE / -6`
- `Integer.MIN_VALUE / Integer.MIN_VALUE`
- `Integer.MIN_VALUE / -1` (overflow)
- `Integer.MIN_VALUE / -3` (overflow)

All existing `FractionTest` tests continue to pass.


garydgregory:
@makarandhinge Thank you, merged 🚀

PR #1795, opened 2026-09-19
Defensively clone the TimeZone in FastDatePrinter and FastDateParser

alhudz:
`FastDateFormat` is documented as immutable and thread-safe and hands out instances from a process-wide cache, but `FastDatePrinter` and `FastDateParser` keep the caller's `TimeZone` by reference, and `getTimeZone()` returns that same object. `TimeZone` is mutable (`setRawOffset()`, `setID()`), so one caller can change the zone of a formatter that unrelated code already holds. Same hazard as `TimeZones.GMT` in #1666, on the cached formatters (the `DateFormatUtils` constants included).

Repro:
```java
FastDateFormat format = FastDateFormat.getInstance("yyyy-MM-dd HH:mm Z", TimeZone.getTimeZone("UTC"));
// anywhere else in the JVM: same cached instance
FastDateFormat.getInstance("yyyy-MM-dd HH:mm Z", TimeZone.getTimeZone("UTC")).getTimeZone().setRawOffset(5 * 3_600_000);
format.format(new Date(0));
```
Expected: `1970-01-01 00:00 +0000`.
Actual: `1970-01-01 05:00 +0500`. `parse()` shifts by the same five hours, and mutating the `TimeZone` passed to `getInstance()` after the call has the same effect.
Cause: both constructors store the `TimeZone` argument as is, and both `getTimeZone()` getters return the field.
Fix: clone the zone in both constructors and return a clone from both getters, as `TimeZone.getDefault()` does. `FastDateFormat.getTimeZone()` delegates to the printer. `GmtTimeZone` and `ImmutableTimeZone` are already immutable, and `TimeZoneStrategy` only copies offsets out of its zones, so these are the only exits. `getTimeZone()` still `equals()` the zone passed in; only its identity changes.

The added `FastDateFormatTest` and `FastDateParserTest` cases fail on the current source and pass after the change. These classes have no Commons Text counterpart.

---

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute? Claude Code (Anthropic) was used to find the defect and to write the patch, the tests and this description. The default `mvn` build was run locally and is green.


garydgregory:
@alhudz merged 🚀 Thank you. Note that the PR was incomplete, see commit 55dcbc4bb1449aa9f7c6ee422cc22b465a35fa2f

alhudz:
Thanks for merging, and for the follow-up in 55dcbc4. I'd missed that the `ArrayKey` in `AbstractFormatCache` kept the caller's `TimeZone` as well, so my "only exits" claim in the description was wrong.

PR #1796, opened 2026-09-20
Fix NumberUtils.createNumber() and isCreatable() for positive Long literals with explicit '+' sign

Alwaysgaurav1:
### Summary
Fixes an issue where `NumberUtils.createNumber()` threw `NumberFormatException` when given valid positive `Long` or `BigInteger` literals with an explicit `+` sign and `l`/`L` type qualifier (e.g. `"+100L"`, `"+0L"`, `"+12345678901234567890L"`), which also caused `NumberUtils.isCreatable()` to unexpectedly return `false`.

### Problem
In `NumberUtils.createNumber(str)`:
```java
case 'l':
case 'L':
    if (dec == null && exp == null && (!numeric.isEmpty() && numeric.charAt(0) == '-' && isDigits(numeric.substring(1)) || isDigits(numeric))) {
        try {
            return createLong(numeric);
        } catch (final NumberFormatException ignored) {
            // Too big for a long
        }
        return createBigInteger(numeric);
    }
    throw new NumberFormatException(str + " is not a valid number.");


Alwaysgaurav1:
@garydgregory, could you pls review this!

garydgregory:
@Alwaysgaurav1 

A negative test for malformed signed input such as `++1L`, plus a Long boundary case, would strengthen the regression coverage.

Alwaysgaurav1:
Thanks @garydgregory! I've added regression tests covering malformed signed inputs (++1L, ++1l, --1L, +-1L, -+1L, +L, -L) for both createNumber() and isCreatable(), as well as Long boundary cases (Long.MAX_VALUE, +Long.MAX_VALUE, Long.MIN_VALUE, and Long.MAX_VALUE + 1 overflowing to BigInteger).

garydgregory:
@Alwaysgaurav1 
FYI, I had to fix a bogus comment.

garydgregory:
@Alwaysgaurav1 merged 🚀 Thank you.

PR #1798, opened 2026-09-22
[LANG-1249] Propagate excludeFieldNames to recursive toString rendering

jadhavgaurav:
https://issues.apache.org/jira/browse/LANG-1249

## What

`RecursiveToStringStyle.appendDetail()` rendered nested objects by calling
the static `ReflectionToStringBuilder.toString(value, style)` convenience
method. That method builds a brand-new `ReflectionToStringBuilder` with no
exclusions, so a field name excluded on the outer builder via
`setExcludeFieldNames()` only applied to the top-level object's own
fields; the exclusion silently stopped applying as soon as traversal
entered a nested object.

## How

`ReflectionToStringBuilder.toString()` now pushes its own
`excludeFieldNames` onto a per-thread stack kept on the style, when the
style is a `RecursiveToStringStyle`, before rendering, and pops it
afterward in a `finally` block. `RecursiveToStringStyle.appendDetail()`
reads the current top of that stack and applies it to each nested
`ReflectionToStringBuilder` it constructs, so the same exclusions now
propagate through every level of recursion.

A stack, rather than a single field, correctly restores the enclosing
value if the style is reentered on the same thread with a different
exclusion list. It is backed by a `LinkedList` rather than an
`ArrayDeque` because the common case, no exclusions configured, pushes a
`null` element, which `ArrayDeque` rejects (this bit me during testing:
switching from `ArrayDeque` to `LinkedList` was needed to avoid an NPE on
every call that doesn't set exclusions).

## Why

Reported behavior in the ticket: excluding a field that only exists on a
nested object (e.g. `Bar.b` when the top-level object is `Foo` containing
a `Bar bar` field) has no effect, even though excluding a top-level field
(`Foo.bar`) works correctly. This fixes that asymmetry.

## Testing

Added `RecursiveToStringStyleTest#testExcludeFieldNamesAppliesToNestedObjects`,
which fails against the prior behavior (the nested field stays in the
output) and passes with this change. Verified the regression test fails
with only the source fix reverted (test kept) and passes with the fix
reapplied.

Ran the full test suite (`mvn test`): 98254 tests, 0 failures, 0 errors,
69 skipped (pre-existing, environment-related, unrelated to this change).
Ran `mvn checkstyle:check`: clean.


garydgregory:
Set to draft status. I'm not sure we actually need this, the Jira is over ten years old without further requests for implementation.

